PR #1, opened 2026-05-05
feat: pin signature requestor to Harvest webapp + RequestAnyAccess flow

sanity:
## Problem

Harvest's "My Store" empty state shows "No ghostkey identities found" because the ghostkey delegate's `ListGhostKeys` filters per-requestor and Harvest has no grant — chicken-and-egg, since discovery is what `ListGhostKeys` is supposed to provide. Resolved upstream by `freenet/ghostkeys#7`, which adds a `RequestAnyAccess` flow: the user picks one of their ghostkeys to share with the requesting app and the delegate auto-grants `{ReadPublic, Sign}` for it.

But there's a second concern: once a user has shared a key with multiple apps, what stops App B from minting signatures that look like App A's? The ghostkey delegate already binds the runtime-attested calling contract id into every signature it produces (`ScopedPayload.requestor`), but Harvest's verifier wasn't checking it. So a `Sign` grant for the same key shared with two different apps could let either one impersonate the other.

## Solution

Two coordinated changes against `ghostkey-common 0.2.3`:

### 1. Verifier-side requestor pin

`HARVEST_WEBAPP_CONTRACT_ID = "6FzSeAUKcqJrveKyU8RJgGKc5jRB1Z2juvxXtwTA4Em9"` is exported from `harvest-common`. `verify_scoped_signature` extracts the requestor embedded in the `ScopedPayload` and hard-fails if it isn't the Harvest webapp. The webapp container contract id is stable across Harvest releases (releases update state, not container code) so the pin doesn't impose a migration burden.

Both feature branches updated:
- `cfg(feature = "ghostkey")` deserialises directly via `ghostkey_common::ScopedPayload` and compares to `expected_harvest_requestor()`.
- `cfg(not(feature = "ghostkey"))` walks raw CBOR (`extract_webapp_requestor_bytes_from_cbor`) and compares against the base58-decoded `HARVEST_WEBAPP_CONTRACT_ID` bytes.

### 2. `RequestAnyAccess` UI flow

The auto-`ListGhostKeys` on connect is removed — wasted work now that the delegate gates list results behind a grant. The empty state shows a "Connect a ghostkey" button that fires `RequestAnyAccess`; the gateway shell-page renders the delegate's prompt; on approval the response handler folds the returned one-element `GhostKeyList` into `APP_STATE.ghostkeys` and `IdentityList` renders.

`AccessDenied` and `NoIdentityAvailable` get explicit response-handler arms with user-facing notifications so the empty state isn't silently stuck after a deny or a vault that lacks any keys.

## Testing

- **`test_authorized_listing_wrong_requestor_fails`** (new regression): a signature whose requestor is some other webapp must NOT verify, even though the signature is valid and the payload matches. Pins the "Sign grant on a shared key shouldn't impersonate Harvest" invariant.
- Existing tests (`test_authorized_listing_verify`, `test_authorized_listing_wrong_key_fails`, `test_authorized_listing_tampered_payload_fails`) updated to use the Harvest webapp id as the requestor.
- `cargo test --workspace --exclude harvest-ui` clean (13/13 pass).
- `cargo clippy --target wasm32-unknown-unknown -- -D warnings` clean for `harvest-common`, `store-contract`, `reputation-contract`, `mailbox-contract`, `harvest-delegate`.

## Impact on existing data

The store contract WASM hash changes (ghostkey-common 0.2.2 → 0.2.3 plus the verifier code change). Any test stores published before this PR are still addressable under the old contract id, but stores published after this PR have a new id. Harvest is in early development (no production stores), so this is acceptable scope.

## Out of scope

- `expected_requestor` is hardcoded rather than carried in `StoreParameters` because every store today must come from the same Harvest webapp. If we ever fork or want per-store flexibility, promoting it to a wire-additive `#[serde(default)] Option<SignatureRequestor>` field is straightforward.
- Vault "Manage app permissions" UI: ghostkeys PR #7 deferred this to a follow-up. Until it ships, users approve a `RequestAnyAccess` grant once and it persists; revocation requires invoking `RevokePermission` from a CLI. Filed as a follow-up.

## Fixes

Closes the cross-app ghostkey access gap that was blocking Harvest from being usable.

[AI-assisted - Claude]

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ harvest-common = { path = "common" }
 freenet-stdlib = { version = "0.6", features = ["contract"] }
 freenet-scaffold = "0.2.2"
 freenet-scaffold-macro = "0.2.2"
-ghostkey-common = "0.2.1"
+ghostkey-common = "0.2.3"
 
 # Serialization
 serde = { version = "1.0", features = ["derive"] }
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -24,29 +24,88 @@ pub use mailbox::{ConversationId, EncryptedMessage, MailboxParameters, MailboxSt
 pub use reputation::{FeedbackEntry, ReputationParameters, ReputationStateV1};
 pub use store::{StoreParameters, StoreStateV1};
 
-/// Stable contract id of the published Harvest webapp container. Used as
-/// the `expected_requestor` when verifying ghostkey-signed listings and
-/// store info: every signature must have been produced by a delegate
-/// call originating from this webapp, ensuring an app the user has
-/// granted ghostkey access to (Harvest itself or another) cannot mint
-/// signatures that pass Harvest's contract verifiers.
+/// Contract id of the Harvest webapp container that this build targets.
 ///
-/// The id is stable across Harvest releases because the underlying web
-/// container contract WASM rarely changes; releases ship by updating
-/// the container's *state* (the bundled UI bytes), not the container
-/// code. If the container code is ever rebuilt with a different hash,
-/// this constant will need to be updated and old stores re-published.
-pub const HARVEST_WEBAPP_CONTRACT_ID: &str = "6FzSeAUKcqJrveKyU8RJgGKc5jRB1Z2juvxXtwTA4Em9";
+/// Read at compile time from `published-contract/contract-id.txt` so that:
+///
+/// * Production, staging, dev, and forked Harvest builds each pin to
+///   their own deployed id — a fork that publishes under a different
+///   container id verifies its own signatures rather than always
+///   trying to match the upstream Harvest deployment.
+/// * The file `published-contract/contract-id.txt`, written by
+///   `cargo make publish-harvest`, is the single source of truth. The
+///   constant tracks whatever's checked in.
+///
+/// Used as the `expected_requestor` when verifying ghostkey-signed
+/// listings and store info: every signature must have been produced
+/// by a delegate call originating from this webapp, ensuring an app
+/// the user has separately granted ghostkey access to via
+/// `RequestAnyAccess` cannot mint signatures that pass Harvest's
+/// contract verifiers.
+///
+/// **Migration note:** if the webapp container WASM is ever rebuilt
+/// with a different hash (e.g. a stdlib bump), the deployed id changes
+/// and listings/stores published under the old id are orphaned. The
+/// migration path is to add the previous id to
+/// `LEGACY_HARVEST_WEBAPP_CONTRACT_IDS` below so historical signatures
+/// continue to verify, similar to how `legacy_delegates.toml` works
+/// for the delegate WASM.
+pub const HARVEST_WEBAPP_CONTRACT_ID: &str =
+    include_str!("../../published-contract/contract-id.txt").trim_ascii();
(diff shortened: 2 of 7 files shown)


sanity:
## Review pass 2: addressed

Pushed 55c36cc with fixes for the parallel review pass.

**CRITICAL (Codex P1 / Skeptical):**
- **Hardcoded webapp id** → `HARVEST_WEBAPP_CONTRACT_ID = include_str!("../../published-contract/contract-id.txt").trim_ascii()`. Each build pins to its own deployment. `LEGACY_HARVEST_WEBAPP_CONTRACT_IDS` slice added for future container-WASM bumps.
- **No way to connect 2nd ghostkey** → added `ConnectAnother` button in `MyStore`. `GhostKeyList` responses now MERGE into `APP_STATE.ghostkeys` (deduped by fingerprint) instead of replacing.
- **`pending_listing` not cleared on auth denial** → `AccessDenied`/`NoIdentityAvailable`/`PermissionDenied`/`Error` arms now uniformly clear pending state + in-flight flag.
- **`connect_ghostkey()` no in-flight guard** → new `request_any_access_in_flight: bool` in `AppState`; Connect buttons disable while set.

**HIGH:**
- Dep pin `ghostkey-common = "0.2.3"` (was `"0.2.1"`).
- **`cfg(not(feature = "ghostkey"))` branch tested + dormant bug fixed** — `extract_payload_from_cbor` was reading `as_bytes()` but `Vec<u8>` actually serializes as a CBOR array of integers. Pre-existing bug exposed by the new no-feature tests; helper now accepts either shape. Verified with `cargo test -p harvest-common --no-default-features` (17/17 pass) AND default features (18/18 pass).
- New tests: `test_authorized_listing_delegate_requestor_fails`, `test_authorized_store_info_verifies`, `test_authorized_store_info_wrong_requestor_fails`, `harvest_webapp_contract_id_parses`, `expected_harvest_requestor_constructs`.

**MEDIUM:**
- `expected_harvest_requestor` cached via `LazyLock` (no per-verify base58 reparse).
- Doc comment on `verify_scoped_signature` now explicitly says the pin defends against delegate-mediated cross-app misuse, NOT PEM theft / `Export`-grant abuse (per Skeptical's H3/L1).
- `connect_ghostkey()` shows user-visible notification when clicked before delegate registration completes (was silent log).

**Considered but not applied:**
- **CI workflow** (testing CRITICAL): repo has no `.github/workflows/` today; adding one is a separate scope. Filed as follow-up. Local `cargo test --workspace --exclude harvest-ui` and `cargo test -p harvest-common --no-default-features` both clean.
- **Vault manage-permissions UI** (big-picture): symmetric defer with ghostkeys#7. Open issue tracking exists in ghostkeys repo; will mirror to harvest.
- **`SetLabel` scope** (Skeptical L4): Harvest doesn't call SetLabel today; comment notes the constraint for future maintainers.

**Verified safe (skeptical questions):**
- `ciborium` preserves struct field insertion order on serialise + `Value::Map` deserialise — `requestor_map.first()` is stable.
- Cross-version `[u8; 32]` shape matches: ghostkey-common 0.2.3 + stdlib 0.6 produce CBOR array-of-integers, which the helper now handles. Both feature configurations exercise this path under test.

[AI-assisted - Claude]

Issue #18, opened 2026-09-03
Committed contract and delegate WASMs are stale and not reproducible


sanity:
## Problem

The four artifacts committed under `ui/public/contracts/` — which the UI embeds with
`include_bytes!`, and whose hashes are therefore the on-network addresses of every
store, reputation and mailbox contract and of the harvest delegate — do not match
what the current source compiles to, and cannot be reproduced by anyone.

Two separate defects.

**They are stale.** Rebuilt from `main`'s source with today's dependency resolution
(pinned toolchain 1.94.1, the one that produced them), every hash differs:

| artifact | committed | rebuilt from source |
|---|---|---|
| `reputation_contract` | `b8577fa15470f472…` | `35dcb87eb8dcf4d2…` |
| `store_contract` | `4d7ad3c31238a4b7…` | `594723cbae3ad700…` |
| `mailbox_contract` | `db0b1c286442209e…` | `c40311d789778c52…` |
| `harvest_delegate` | `1fa5776b332464a2…` | `8cda05db5673bc3f…` |

The cause is the one #4356-shaped hazard this repo had no defence against until
now: they were built without a lockfile, so the dependency versions baked into them
(`chrono 0.4.44`, `blake3 1.8.4`, …) are simply whatever resolved on the day
someone last ran `cargo make sync-wasm`.

**They are not reproducible.** They were built without `--remap-path-prefix`, so
they carry absolute paths from the machine that built them:

```
/home/<user>/.cargo/registry/src/index.crates.io-…/chrono-0.4.44/src/format/mod.rs
```

Anyone else rebuilding gets different bytes, hence a different address, purely
because their home directory is spelled differently.

## Why this was not fixed in the CI PR

Refreshing them is a re-key of every Harvest contract and delegate. That is a
deliberate act with a migration question attached, not something to fold into a
change to the build system — so the drift guard added in #19 is a *relative* check
(does this PR move the address) rather than an absolute one.

Harvest appears to be dormant with no known users (#5: "nobody holds a link to its
container contract"), so the cost of the re-key is currently near zero. This is the
cheap moment to do it.

## Fix

1. Fix #7 first if possible — the delegate currently imports `env.__getrandom_custom`
   and chrono's wasmbind JS `Date`, neither of which the runtime provides, so a
   refresh done now bakes a delegate that still cannot instantiate. Otherwise the
   refresh will just have to happen twice.
2. `./scripts/build-contract-wasm.sh --sync`, commit the result with the
[...]

Issue #7, opened 2026-08-26
deployed delegate cannot instantiate: chrono wasmbind emits JS Date imports the runtime cannot resolve


sanity:
## Problem

The deployed harvest delegate **cannot instantiate**. It was built with chrono's `wasmbind` feature, so `Utc::now()` compiled to JavaScript `Date` calls rather than to the Freenet host clock.

The module imports:

- `__wbindgen_placeholder__.__wbg_new_0_…`
- `__wbindgen_placeholder__.__wbg_getTime_…`
- `__wbindgen_externref_xform__.*`
- `env.__getrandom_custom`

The Freenet runtime registers only `freenet_contract_io`, `freenet_delegate_ctx`, `freenet_delegate_secrets`, `freenet_log`, `freenet_rand` and `freenet_time`. None of the above resolves, so instantiation fails before any code runs.

Affected artifact: `7fhAoEwvGt89gnyGSYVob1eu2qxJMkoGu6QHzGWDMcsD.wasm`, present in two delegate stores on the machine where this was found (`freenet-conformance-capture/data/delegates` and `freenet-hosted/data/delegates`).

## How it was found

A sweep of **every** WASM in every contract and delegate store on one machine — 517 modules — looking for imports the runtime cannot satisfy. **This is the only one.** Every other module imports exclusively from the registered set.

That makes it a clean single instance rather than a class, but the cause is a footgun that will recur: chrono's default features pull in `wasmbind` on `wasm32-unknown-unknown`, and the resulting binary looks fine until it is loaded.

## Fix

Build with chrono's default features off and use the host clock:

```toml
chrono = { version = "0.4", default-features = false }
```

and call `freenet_stdlib::time::now()` rather than `Utc::now()` in code compiled to WASM. (freenet-stdlib pins chrono this way itself, which is why `Utc::now()` is not normally reachable from a contract or delegate build.)

Note the delegate's use of the clock is **not** the problem and does not need removing: rate limiting "N per hour" genuinely requires knowing the hour, delegates hold private per-node state, are not replicated, and have no merge laws to satisfy. This is purely about linking against the wrong clock.

## Suggested guard

A pre-publish check that a delegate's or contract's import set is a subset of what the runtime registers would have caught this at build time rather than at load. Cheap to implement — `wasm-objdump -x` and a set comparison — and it generalises to every future instance of this footgun.

## Context

[...]

PR #19, opened 2026-09-03
ci: commit Cargo.lock and fail PRs that move a contract address

sanity:
## Problem

Harvest's cargo workspace contains three contract crates and a delegate crate. In
Freenet the compiled WASM **is** the address — a contract lives at
`BLAKE3(BLAKE3(wasm) || parameters)` — so anything that changes codegen moves the
address and strands whatever state lived at the old one.

Before this PR the repository had none of the machinery that makes that safe:

- **No committed `Cargo.lock`** (it was explicitly gitignored). Resolved dependency
  versions are an input to the address, so the contracts had no stable identity;
  it also meant GitHub could not build a dependency graph, so Dependabot **security**
  alerts could never fire for this repo.
- **No `rust-toolchain.toml`.** rustc codegen decides the address too.
- **No GitHub Actions workflows at all.** Nothing built the repo, nothing ran the
  tests, nothing looked at the addresses.

Dependabot was just enabled across the org. There are **eight open dependency PRs on
this repository right now**, all with zero checks.

This is not hypothetical. A single-patch dependency change re-keys everything:

```
chrono 0.4.45 -> 0.4.44   (one patch release, transitive to the contracts)

  reputation_contract  35dcb87eb8dcf4d2… -> dd846ebb812ca780…
  store_contract       594723cbae3ad700… -> 59e437370e21d77b…
  mailbox_contract     c40311d789778c52… -> 1dd96288c60ac072…
  harvest_delegate     8cda05db5673bc3f… -> ffe5c614efe9a38b…
```

Nothing in the diff of such a PR says "this is a re-key".

## Approach

**1. Commit `Cargo.lock`** and remove it from `.gitignore`. The workspace resolves
and builds with it: `cargo check --workspace --locked --all-targets` and
`cargo test --locked --workspace` (19 tests) both pass, and all four wasm artifacts
build from it.

**2. Pin `rust-toolchain.toml` to 1.94.1** — the toolchain that produced the
artifacts currently committed under `ui/public/contracts/`, so the pin records a
fact rather than a preference. See "Two things this PR found" below: raising it is
not a routine bump.

**3. One reproducible build path.** `scripts/build-contract-wasm.sh` builds all four
artifacts in a single cargo invocation with `--locked` and `--remap-path-prefix`,
and prints their BLAKE3 code hashes. `cargo make sync-wasm` and CI both call it, so
the addresses CI reports are the addresses a publish would deploy. The two `cargo
build` tasks it replaces in `Makefile.toml` are removed rather than kept as aliases —
a second build path is a second set of bytes, which is a second contract.

Two properties this buys, both verified rather than assumed:

- *Byte-reproducible across machines.* The committed artifacts today carry
  `/home/<user>/.cargo/registry/...` strings baked into panic locations, so nobody
  else can reproduce them. With the remap flags, building the same tree at two
  different absolute paths produced **byte-identical** output.
- *One invocation.* Cargo unifies features across the packages named in a single
  invocation, so the old `build-delegate` + `build-contracts` split could resolve
  different features than a combined build.

**4. `ci.yml`** — fmt, clippy (`-D warnings`), test, plus a wasm32 build of the
contracts and delegate. The wasm job is not redundant with the native one: these
crates link differently for wasm32, and a bump can compile clean on the host and
fail to link for the real target.

**5. `contract-drift.yml`** — the guard. It rebuilds the contracts and delegate at
the PR head **and** at the merge base, both with this PR's toolchain pin and build
script, and fails if any BLAKE3 code hash moved. The old and new hashes are printed
in the job summary.

Why base-vs-head rather than "rebuild and diff against the committed artifact"
(atlas's shape): the artifacts committed under `ui/public/contracts/` are already
stale relative to source, so that check would be red on arrival for a reason
unrelated to any PR. Refreshing them is a deliberate re-key that should be its own
decision, not smuggled into a CI change. Base-vs-head answers the question a
dependency bump actually raises — *does this change move the address* — with no
false positives and nothing to bootstrap. Filed as follow-up:
**#18** (refresh the artifacts, then the guard can be upgraded to the simpler form).

The escape hatch is the `contract-rekey-acknowledged` label. The job still runs and
still prints every moved address; the label records that a human read them. Two
re-key vectors the comparison cannot see by construction — a change to
`rust-toolchain.toml` or to the build script, which rebuild *both* sides under the
new rules — are checked directly by a second step and need the same label.

**This PR carries that label itself**, and that is not a workaround: it introduces
the toolchain pin and the build script, so by its own (correct) rule every address
moves. The artifacts under `ui/public/contracts/` are untouched here.

## Two things this PR found and did not fix

Both are pre-existing, both are reported rather than quietly patched, because
fixing either changes contract bytes.

1. **`harvest-delegate` no longer links for wasm32 on rustc ≥ 1.95.** Its graph
   enables `getrandom`'s `custom` feature and nothing registers an implementation,
   so `env.__getrandom_custom` is undefined. Up to 1.94.1 rust-lld turned that into
   a wasm import and linked anyway; newer rust-lld rejects it. The import is a live
   defect either way — the Freenet runtime registers no `env` module — and it is
   already listed in **#7**. The newer toolchain's strictness is the correct
   behaviour, not the obstacle; #7 has to be fixed before the pin can be raised, and
   `rust-toolchain.toml` says so at the point where someone would try.
2. **The committed artifacts are stale.** Rebuilt from current source and today's
   dependency resolution, all four hashes differ from what is in
   `ui/public/contracts/`. Filed as **#18**.

`main` is unprotected, so this check can be bypassed by a direct push. Worth
requiring `no contract or delegate address moved` and the `CI` jobs in branch
protection.

## Testing

**Guard fires, then passes — verified locally against real builds:**

| step | result |
|---|---|
| build at baseline lockfile | 4 hashes recorded |
| `cargo update -p chrono --precise 0.4.44`, rebuild | all four hashes **MOVED** (table above) |
| restore lockfile, rebuild | all four hashes `unchanged` |

**Verified in CI on this PR** by pushing a temporary commit that downgraded one
transitive dependency in `Cargo.lock`, watching `no contract or delegate address
moved` go red with the moved addresses in the summary, then reverting it and
watching the same job go green. Run links in the comments below.

**Determinism**, which the guard depends on: two clean builds of the same tree at
the same path are byte-identical, and — with the remap flags — so are builds at two
*different* absolute paths.

**Other checks:** `cargo fmt --all -- --check` clean; `cargo clippy --locked
--workspace --all-targets -- -D warnings` clean after two fixes in
`ui/src/document_title.rs` (an unused `title` parameter on the host build and a
`thread_local` that can be `const`). Both are UI-only and provably do not affect the
contract or delegate bytes — the four code hashes are identical before and after.

[AI-assisted - Claude]


sanity:
Applying `contract-rekey-acknowledged`. Recording why, since the label is the whole point of the gate and an unexplained one is worthless.

**No deployed address moves by merging this.** The PR touches no artifact — `ui/public/contracts/` is untouched, verified against the file list. What it changes is what a *future* rebuild produces, via the toolchain pin and the `--remap-path-prefix` flags in the new build script.

**That future divergence already exists and is not created here.** The committed artifacts are already stale against current source and current dependency resolution (#18), and they carry `/home/<user>/.cargo/registry/...` paths baked into panic locations, so nobody but their original author could reproduce them at all. The choice is not between "addresses stable" and "addresses move" — it is between a build nobody can reproduce and one anybody can. This PR takes the second.

**The pin records a fact rather than a preference:** 1.94.1 is the toolchain that produced the artifacts currently committed. Raising it is separately blocked by #7 (`harvest-delegate` stops linking for wasm32 on rustc >= 1.95 because `getrandom`'s `custom` feature is enabled with no implementation registered), and `rust-toolchain.toml` says so where someone would try.

I also agree with the base-vs-head design over atlas's rebuild-and-compare-to-artifact shape. With the artifacts stale, atlas's shape would be red on arrival for a reason unrelated to any PR, and a check that is red for unrelated reasons trains people to ignore it.

Reviewed the diff, the failure log, and the four moved hashes. Merging once the re-run is green.

[AI-assisted - Claude]

sanity:
## Guard verified in CI, not just locally

Fail-then-pass, on real runs of this workflow.

**It fires.** A temporary commit downgraded one transitive dependency by a single
patch release (`chrono 0.4.45 -> 0.4.44`) and changed nothing else. The
`compare addresses` step went red with every address it moved:

https://github.com/freenet/harvest/actions/runs/33806866348

```
reputation_contract RE-KEYS: 35dcb87eb8dcf4d2… -> dd846ebb812ca780…
store_contract      RE-KEYS: 594723cbae3ad700… -> 59e437370e21d77b…
mailbox_contract    RE-KEYS: c40311d789778c52… -> 1dd96288c60ac072…
harvest_delegate    RE-KEYS: 8cda05db5673bc3f… -> ffe5c614efe9a38b…
```

That commit has been reverted; the branch no longer contains it.

**It passes when nothing moved.** On the same branch without that downgrade,
`compare addresses` reports all four `unchanged`:

https://github.com/freenet/harvest/actions/runs/33806640195

**Reproducibility, checked three independent ways rather than assumed.** The
guard is only meaningful if the same source compiles to the same bytes, so:
two clean builds at the same path are byte-identical; builds at two *different*
absolute paths are byte-identical (this is what the `--remap-path-prefix` flags
buy, and the currently committed artifacts fail it — they carry
`/home/<user>/.cargo/registry/...` strings); and the GitHub runner's hashes are
byte-identical to the ones produced on a developer machine. That last one is the
useful property in practice: `./scripts/build-contract-wasm.sh` locally prints
the same addresses CI does.

**The other half of the guard.** The `compare addresses` step builds both sides
under this PR's toolchain and build script, so it is blind by construction to a
change in either — and both re-key everything. That is why there is a second
step checking them directly, and it is the step that is red on this PR:

```
##[error]this PR changes: rust-toolchain.toml
Both determine the compiled bytes, so every contract and delegate address moves.
```

Correct, and acknowledged with the label. The artifacts under
`ui/public/contracts/` are untouched here; refreshing them is #18.

One gap found while doing this and fixed in `abe71ff`: the acknowledgement label
is read from the event payload, so without `labeled`/`unlabeled` in the trigger
types, applying it had no effect until the next push — and `gh run rerun` does
not help, because a re-run replays the original payload. The escape hatch would
have been unusable.

[AI-assisted - Claude]

PR #20, opened 2026-09-03
ci: name the acknowledgement label in the drift guard's failure output

sanity:
## Problem

Follow-up to #19. Two gaps in the drift guard, both found in review.

**1. A red check did not state its own remedy.** The `::error::` output named the
addresses that moved but not what to do about it. The label is the only way past
the gate, so a reviewer reading a red check had to open the workflow file to find
that out — and the escape hatch is exactly the part nobody should have to go
looking for.

**2. The `labeled` path was never actually demonstrated.** #19 added
`types: [opened, synchronize, reopened, labeled, unlabeled]`, but the run that
went green on that PR was triggered by a *push* that happened to carry the label,
not by the label event itself. So the fix was plausible rather than proven. It is
proven now — see Testing.

## Approach

- Both failing steps now print the remedy as `::error::` output, naming
  `contract-rekey-acknowledged` explicitly and saying that applying it re-runs the
  check with no push needed.
- A comment records why the `paths:` filter is correct for label events: it applies
  to every activity type, so labelling a PR that touches none of the listed paths
  does not run the job — which is right, because such a PR cannot have moved an
  address. It is not a hole either: a PR that *does* touch those paths always runs
  the job on its own push, so a label event only ever re-evaluates a check that has
  already run.

No behavioural change to the comparison itself.

## Testing

The label path, demonstrated end to end on this PR with **no code push between
states** — a temporary commit downgraded `chrono` one patch release so the check
was genuinely red, and every transition below was caused by a label event alone:

| # | event | trigger | result |
|---|---|---|---|
| 1 | push (temp downgrade present) | `synchronize` | red — four addresses moved |
| 2 | label applied | `labeled` | **green**, no push |
| 3 | label removed | `unlabeled` | **red again**, no push |
| 4 | temp commit dropped, no label | `synchronize` | green |

Run links in the comment below. Step 3 is the one that matters as much as step 2:
removing an acknowledgement re-asserts the gate rather than leaving the last green
result standing.

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/contract-drift.yml
+++ b/.github/workflows/contract-drift.yml
@@ -44,6 +44,12 @@ on:
     # matters as much as `labeled`: removing the acknowledgement has to turn the
     # check red again rather than leaving a stale green.
     types: [opened, synchronize, reopened, labeled, unlabeled]
+    # The paths filter applies to every activity type, label events included.
+    # That is the behaviour we want: labelling a PR that touches none of these
+    # paths cannot have moved an address, so the job does not need to run and
+    # does not. The filter is not a hole in the gate either -- a PR that DOES
+    # touch them always runs the job on its own push, so the label event is
+    # only ever re-evaluating a check that already ran.
     paths:
       # Everything that can change the compiled bytes. `ui/**` is absent on
       # purpose: the UI is not a dependency of the contracts or the delegate.
@@ -191,7 +197,7 @@ jobs:
               echo "unchanged: $a  $h"
               echo "| \`$a\` | \`${b:0:16}…\` | \`${h:0:16}…\` | unchanged |" >> "$GITHUB_STEP_SUMMARY"
             else
-              echo "::error::$a RE-KEYS: $b -> $h"
+              echo "::error::$a RE-KEYS: $b -> $h  (see the job summary; label this PR contract-rekey-acknowledged if the re-key is intended)"
               echo "| \`$a\` | \`${b:0:16}…\` | \`${h:0:16}…\` | :warning: **MOVED** |" >> "$GITHUB_STEP_SUMMARY"
               moved=1
             fi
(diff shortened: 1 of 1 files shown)


sanity:
## The label path, demonstrated

Four states, **one head SHA (`11ab721`) across the first three** — so nothing
between red and green was caused by a code push. A temporary commit downgraded
`chrono` one patch release to make the check genuinely red; it has been removed.

| # | what changed | triggering event | run | result |
|---|---|---|---|---|
| 1 | pushed the temp downgrade | `synchronize` | [33807555703](https://github.com/freenet/harvest/actions/runs/33807555703) | **red** — four addresses moved |
| 2 | applied the label, nothing else | `labeled` | [33807732001](https://github.com/freenet/harvest/actions/runs/33807732001) | **green**, same SHA |
| 3 | removed the label, nothing else | `unlabeled` | [33807926555](https://github.com/freenet/harvest/actions/runs/33807926555) | **red again**, same SHA |
| 4 | dropped the temp commit, no label | `synchronize` | [33808068105](https://github.com/freenet/harvest/actions/runs/33808068105) | green |

Step 3 matters as much as step 2: withdrawing an acknowledgement re-asserts the
gate rather than leaving the previous green result standing.

The green run at step 2 did not pass by skipping anything. It printed all four
`RE-KEYS` errors and then took the acknowledged path explicitly:

```
##[error]reputation_contract RE-KEYS: 35dcb87eb8dcf4d2… -> dd846ebb812ca780…  (see the job
         summary; label this PR contract-rekey-acknowledged if the re-key is intended)
…
##[warning]addresses moved, but the PR carries contract-rekey-acknowledged
```

## The comparison step is not passing for the wrong reason

Asked directly, so answered directly. On the clean run at step 4 the `compare
addresses` step reports each artifact by name with its full hash, and the second
step separately confirms the inputs it guards:

```
unchanged: reputation_contract  35dcb87eb8dcf4d296dfbd992bbaca83f1ddd1736a2a82a59fbd91177d2fe058
unchanged: store_contract       594723cbae3ad7002b51b1bd201c50ff727569a75881a2afb08cca443334ee0a
unchanged: mailbox_contract     c40311d789778c522f816f69a4d406799ebecea6593c246277b705b4ead3758c
unchanged: harvest_delegate     8cda05db5673bc3f6eb1fad5e421a8b694a566b83d8032aa581afe5256a9ebaf
toolchain pin and build procedure unchanged
```

Three independent reasons that is a real pass rather than a vacuous one: the same
step went red on runs 33807555703 and 33807926555 under a one-patch dependency
change; a missing artifact on either side fails the step rather than being skipped;
and those four hashes are reproduced byte-identically on a developer machine and on
the runner.

[AI-assisted - Claude]

PR #27, opened 2026-09-06
fix(publish): point the publish task at a node, and stop it claiming success

sanity:
Two defects, both found by publishing Harvest for real for the first time.

**`fdev publish` needs `encodingProtocol=native` on the node URL, and the task set no URL at all.** Without the parameter the node answers in a different encoding and fdev misparses the reply as `invalid value: integer 12, expected Ok or Err` — which reads like a client/node version mismatch. It is not: an fdev built from the running node's own release commit fails identically. That misdiagnosis cost a `cargo install` before the real cause turned up.

**Worse, `fdev` exits 0 on a failed publish.** It printed `Error: Failed to receive response` and returned 0, so `set -e` did not catch it and the task would have printed "Published Harvest to Freenet" over a publish that never happened. The success line is now conditioned on fdev's own success output.

Also records the verification step, because **a 200 from the node proves nothing** — it serves the previous bundle until re-extraction. The only honest check is comparing the served asset hash against the built one, which is how the first failed publish was caught. Every other signal said it had worked: fdev said "Request submitted", the endpoint returned 200, and the contract id matched.

Commits the version counter for today's publish, as the task itself asks.

[AI-assisted - Claude]


sanity:
## Review record

**Tier: Full.** This is release tooling — the task that publishes the app — which the review rule puts at Full regardless of diff size.

### Lenses that ran

| lens | outcome |
|---|---|
| Independent blind review (separate session, read-only, given the diff and the background) | **DO NOT MERGE** — one blocking finding, fixed in 3533854 |
| Guard exercised against real fdev outputs | passes, see table below |
| **External non-Claude (`codex review`)** | **UNAVAILABLE** — usage limit, reset 2026-09-07 08:14; substituted per the review rule |

Four subagent reviewers spawned for this PR and #31 went idle without delivering reports; that work is **not** counted as review, which is why the lens above was obtained through a different mechanism.

### Blocking finding — the guard was wrong on the case that matters

The new success gate matched only `"Contract published successfully"`. But `fdev`'s `put()` has **two** success arms, both returning `Ok(())` (`fdev-0.3.295 src/commands.rs`):

```rust
PutResponse    => tracing::info!(%response_key, "Contract published successfully");  // :287
UpdateResponse => tracing::info!(%response_key, "Contract updated successfully");    // :295
```

The node answers `UpdateResponse` when it **already hosts the contract** — which is every publish after the first, i.e. the normal workflow for updating the deployed site. So the gate would have declared each successful update a failure and refused to commit `last-version`, starting with the very next publish.

Verified against the installed `fdev-0.3.295` source rather than taken on the reviewer's word, since a cited line number is not evidence by itself.

**Why I did not catch this.** I watched the guard work — on a genuine first publish, which took the `Put` path. That is the one case it would never meet again. A guard verified only on the path that happens to work has not been verified, and this repo's recurring defect is guards that report success while measuring nothing. This was the same shape, inverted: a guard that would report failure while measuring nothing.

Fixed to `grep -qE "Contract (published|updated) successfully"`.

### The guard, exercised

Not argued — run:

| input | result |
|---|---|
| `Contract published successfully` (first publish) | ACCEPT |
| `Contract updated successfully` (re-publish) | ACCEPT |
| `Error: Failed to receive response` (the observed exit-0 failure) | REJECT |
| `Error: invalid value: integer 12, expected Ok or Err` (missing `encodingProtocol`) | REJECT |
| empty output (crash) | REJECT |
| `Website updated successfully!` (fdev's `website` command) | REJECT |

The last one is included because it is close enough to matter: `website.rs:412` emits similar wording, and a looser pattern would have accepted it.

### Non-findings, checked and cleared

- **`: "${FREENET_NODE_URL:=...}"`** — correct, including the usual footgun: `:=` treats empty the same as unset, so `FREENET_NODE_URL=""` is defaulted rather than passed through empty.
- **`|| true`** — masks nothing. It exists only so `set -e` does not abort before the content check runs; every failure still reaches the gate and becomes `exit 1`.
- **Quoting / stray expansion** — no repeat of the `$SERVE` bug. The only live expansion is `$FREENET_NODE_URL`, which is intended; `2>&1` is correctly attached to the last line of the multi-line invocation, so it captures the whole command.
- **`last-version`** — `sign-webapp` writes it before `fdev publish`, and the committed value is consistent with the publish that actually succeeded.

### Shell verified

`Makefile.toml` parses, and the extracted script passes `bash -n`.

[AI-assisted - Claude]

PR #31, opened 2026-09-06
fix(ci): drop the drift job's paths filter so its required check always reports

sanity:
## Problem

The repository ruleset **"Ensure CI passes before merge"** (added 2026-09-05) requires four status checks, one of which is this workflow's job name, `no contract or delegate address moved`. The workflow also carried a `paths:` filter listing the build inputs.

Those two are incompatible. On a pull request touching none of those paths the job does not run, so the required check never reports — and a required check that never reports is treated as **missing**, not satisfied. GitHub then shows the pull request as `BLOCKED` while displaying no failing check, so there is nothing for a reader to act on.

This blocked every pull request that avoids the build inputs: docs, CI, UI, tooling.

Observed on #27, a Makefile-only change:

```
mergeStateStatus: BLOCKED
reviewDecision:   ''
fmt, clippy, test                          -> SUCCESS
contracts, delegate and UI build to wasm   -> SUCCESS
committed contract wasm matches a fresh build -> SUCCESS
license/cla                                -> SUCCESS
```

Four green, and the fifth required check absent because it had no reason to run.

## Approach

Remove the `paths:` filter; the job always runs.

The filter's original reasoning was sound and this change does not dispute it — a pull request that touches no build input genuinely cannot have moved an address, so the two contract builds really are pointless work there. But that is an argument about **cost**, and the ruleset is asking a different question: does the check **report**. Skipping settles the first by breaking the second.

**Why not skip-logic inside the job instead?** That was the alternative: always trigger, compute whether relevant paths changed, exit early if not. It keeps the cost saving, and I rejected it. This repository's characteristic defect is a guard that reports success while measuring nothing — the `paths:` filter comment itself records one, where `StoreParameters` shed two fields, took its CBOR from 109 to 56 bytes, re-keyed every published store, and both drift guards truthfully reported "unchanged". A hand-written path comparison inside the job is exactly that shape: one wrong pattern and a genuine re-key merges under a green check.

The change as made can only ever add checking, never remove it, which is the property worth having in a guard.

The cost is real and accepted. `Swatinem/rust-cache` absorbs most of it, and dependabot's pull requests touch `Cargo.lock`, so they run the job under either arrangement.

## Also fixed

Two comments this change falsifies, since stale comments asserting properties the code lacks are the other recurring defect here:

- the note claiming `main` is unprotected and that requiring this check would be an argument for future work — the ruleset now requires it;
- the note in the merge-base build step justifying itself by `build-contract-wasm.sh` being in the `paths:` filter, which no longer exists. A change to that script still runs this job, because everything now does.

## Testing

No test is possible for a workflow trigger short of running it. The verifiable claims:

- the YAML parses and `pull_request` now has only `types`, no `paths`;
- the job's `name:` is `no contract or delegate address moved`, matching the required-check string in the ruleset exactly — if that string ever drifts, the check goes missing again and the same deadlock returns;
- this PR touches `.github/workflows/contract-drift.yml`, which was **in** the old filter, so the job runs on this PR under the old rules too. That is what makes this fix able to merge while the bug it fixes is still in effect.

After this lands, #27 needs a rebase to pick up the workflow in its merge commit before its check will report.

[AI-assisted - Claude]

### DIFF
--- a/.github/workflows/contract-drift.yml
+++ b/.github/workflows/contract-drift.yml
@@ -64,32 +64,49 @@ on:
     # matters as much as `labeled`: removing the acknowledgement has to turn the
     # check red again rather than leaving a stale green.
     types: [opened, synchronize, reopened, labeled, unlabeled]
-    # The paths filter applies to every activity type, label events included.
-    # That is the behaviour we want: labelling a PR that touches none of these
-    # paths cannot have moved an address, so the job does not need to run and
-    # does not. The filter is not a hole in the gate either -- a PR that DOES
-    # touch them always runs the job on its own push, so the label event is
-    # only ever re-evaluating a check that already ran.
-    paths:
-      # Everything that can change the compiled bytes. `ui/**` is absent on
-      # purpose: the UI is not a dependency of the contracts or the delegate.
-      # `ui/public/contracts/**` IS here, because editing a committed artifact
-      # is a re-key by definition.
-      - "Cargo.toml"
-      - "Cargo.lock"
-      - "rust-toolchain.toml"
-      - "common/**"
-      - "contracts/**"
-      - "delegates/**"
-      - "ui/public/contracts/**"
-      - "scripts/build-contract-wasm.sh"
-      - ".github/workflows/contract-drift.yml"
+    # THERE IS DELIBERATELY NO `paths:` FILTER, and adding one back breaks every
+    # unrelated pull request.
+    #
+    # This job's name is a REQUIRED status check in the repository ruleset
+    # "Ensure CI passes before merge". A required check that never reports is
+    # not treated as satisfied -- it is treated as missing, and GitHub reports
+    # the pull request as BLOCKED with no failing check to point at. A `paths:`
+    # filter makes the job not run at all on a pull request that touches none
+    # of those paths, so such a pull request can never merge.
+    #
+    # That is not hypothetical: the filter and the ruleset coexisted for a day,
+    # and PR #27 (a Makefile-only change) sat MERGEABLE with four green checks
+    # and mergeStateStatus BLOCKED, because the fifth required check had no
+    # reason to run and so never reported.
+    #
+    # The earlier reasoning here was sound in isolation -- a pull request that
+    # touches no build input cannot have moved an address, so the work is
+    # pointless -- and it is still true. It is just not a reason to SKIP the
+    # job, because "pointless work" and "check that never reports" are settled
+    # by different mechanisms. The job now always runs and, on such a pull
+    # request, builds both sides to identical bytes and reports green, which is
+    # the correct answer rather than the absent one.
+    #
+    # The cost is real (two contract builds) and is paid on pull requests that
+    # cannot have moved an address. It is accepted here rather than reintroduced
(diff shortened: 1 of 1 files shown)


sanity:
## Review record

**Tier: Full.** CI configuration is a high-risk surface under our review rule regardless of diff size, and this change is to a required merge gate.

### Lenses that ran

| lens | outcome |
|---|---|
| Independent blind review (separate session, read-only, given the diff without my reasoning) | **VERDICT: MERGE**, with one finding — fixed, see below |
| Self-review against the workflow's own comparison logic | no further findings |
| **External non-Claude (`codex review --base origin/main`)** | **UNAVAILABLE** — see below |

### External pass: attempted, unavailable, substituted

`codex review` was run as part of the default Full-tier flow and failed:

```
ERROR: You've hit your usage limit. ... try again at Sep 7th, 2026 8:14 AM.
```

The limit names a reset ~16h out, so the single permitted retry could not succeed and was not spent. Per the review rule an extra independent Claude lens was substituted rather than the independent pass being skipped. Recording it here rather than downgrading silently.

Separately, four subagent reviewers spawned for this PR and #27 went idle without delivering reports, so their work is **not** counted as review. The independent lens above was obtained through a different mechanism precisely because of that.

### Finding (fixed in 8fe6eb7)

The first commit reworded the paragraph about direct pushes to `main` while keeping the clause *"a direct push to main bypasses this check entirely."* That clause predates the ruleset and was carried forward unverified.

Checked: the ruleset targets `~DEFAULT_BRANCH`, `enforcement: active`, `bypass_actors: []`. Required status checks are not solely a pull-request mechanism, so the claim is at best untested and probably wrong. Rewritten to state the ruleset's actual scope and to say plainly that the push path has not been tested here.

This is the exact defect class the file exists to prevent — a comment asserting a property the configuration does not have — so thanks to the lens that caught it.

### The load-bearing claim, verified empirically

The claim this whole change rests on is that the job **passes**, rather than errors, on a PR touching no build inputs. If it errored, this would convert "blocked with no failing check" into "blocked with a failing check", which is not a fix.

**This PR is itself that test case.** It changes no contract, `common/`, or delegate source, so both sides build byte-identical WASM and every address matches. Result on `9c3aa38`:

```
no contract or delegate address moved   pass   2m22s
mergeStateStatus: CLEAN
```

Before this change that check was absent and the PR would have read `BLOCKED`. Reading the comparison step confirms why it passes: identical addresses take the `moved=0` path and exit 0.

Measured cost: **~2m22s** per otherwise-unaffected PR. The commit message originally hand-waved this as "rust-cache absorbs most of it"; that is the actual number.

### Considered and rejected

**De-requiring the check instead.** Under a ruleset only checks on the required list gate a merge, so a red non-required drift check would not block. De-requiring would let a genuine contract re-key merge under a visibly red but ignored check — strictly worse than the bug being fixed.

**Skip-logic inside the job.** Keeps the cost saving; rejected because a hand-written path comparison is exactly the "guard that reports success while measuring nothing" shape this repository keeps producing. The change as made can only ever add checking.

### Known gap, not addressed here

There is no guard against someone re-adding a `paths:` filter to a required check later. Raised by the independent lens; left out of scope for a config fix this small, and worth a follow-up.

[AI-assisted - Claude]

Issue #30, opened 2026-09-06
bitcoin: build-time addressing constants are stale, so payment silently cannot complete


sanity:
## What

Harvest's build-time Bitcoin addressing constants no longer match the deployed bridge, so **payment cannot complete and the failure is silent**. Both wrong values are well-formed, so every existing well-formedness check passes.

| constant | Harvest ships | live deployment |
|---|---|---|
| `well_known_tip_contract_id(Signet)` | `B24HMUFasG3Yd1EJxfzb3qTPos1tLMiKo5gYiKwaihqT` | `FXFgLKfuMm3NPtzWg3Ghgt5otv4Yo7N4CWGDvHpVeZMm` |
| `ADDRESS_CONTRACT_CODE_HASH_HEX` | `3b5f1df28497b1cfb365798cb86fc87a7e45480d47c79e22f09b9f801e95463f` | `cd2ae7418e3b29c2770fab549763b8a268d54787a86a9bfad5b59ca3d2023555` |
| `TRUSTED_BRIDGE_ID_BS58` | `4MZnDAQWccEWXBUb1wt4iTEkDi6Z2MCcZ9WQN1umRsVL` | same — **correct** |

Both stale values are in `ui/src/gateway/bitcoin_config.rs`.

Note the split: the **trust** constant is right and both **addressing** constants are wrong. The app trusts the correct bridge and then looks in the wrong places, which is why nothing errors.

## Evidence

`3b5f1df28497…` is **A3** in `freenet-bitcoin/legacy/address_contract.toml`, dated 2026-09-03, in a file whose header states it lists *only superseded* generations. The current entry is A8 = `cd2ae741…`, which is exactly what the running bridge reports.

Live bridge (`GET http://127.0.0.1:8431/v1/status` on nova):

```json
{ "address_code_hash": "cd2ae7418e3b29c2770fab549763b8a268d54787a86a9bfad5b59ca3d2023555",
  "bridge_id": "4MZnDAQWccEWXBUb1wt4iTEkDi6Z2MCcZ9WQN1umRsVL",
  "networks": [ { "network": "signet", "tip_contract_id": "FXFgLKfuMm3NPtzWg3Ghgt5otv4Yo7N4CWGDvHpVeZMm", "tip_height": 320955 } ] }
```

[...]

PR #32, opened 2026-09-06
fix(bitcoin): correct stale addressing constants, and refuse to publish stale ones

sanity:
Fixes the immediate breakage in #30 and adds the gate that stops it recurring on the next contract rebuild.

## Problem

Harvest's build-time Bitcoin **addressing** constants no longer matched the deployed bridge. Both wrong values were well-formed, so every existing check passed.

| constant | was | now |
|---|---|---|
| `well_known_tip_contract_id(Signet)` | `B24HMUFa…` | `FXFgLKfu…` |
| `ADDRESS_CONTRACT_CODE_HASH_HEX` | `3b5f1df2…` (generation **A3**) | `cd2ae741…` (**A8**) |
| `TRUSTED_BRIDGE_ID_BS58` | *correct* | unchanged |

The split is why nothing errored: the app trusted the right signer and looked in the wrong places.

## The failure was not what the code predicted

The module docs anticipated this going wrong as *"this network has no data yet"*. Observed in the published app instead:

```
#320557   34 tx     3 days ago
#320556   903 tx    3 days ago
...
```

while the bridge was at height **320955**. The superseded tip contract still exists on the network and still holds the last state it was given, so Payments rendered a tip ~400 blocks stale **as though it were current**. Confirmation depth is measured against that tip.

Runtime evidence from the deployed app, not inferred from source:

```
INFO ui/src/gateway/delegate_api.rs:51  GET contract (subscribe=true):
     ContractInstanceId("B24HMUFasG3Yd1EJxfzb3qTPos1tLMiKo5gYiKwaihqT")
```

**A stale content-addressed constant does not fail quiet, it fails plausible.** The comments are updated to say so, because the old wording would send the next reader hunting for silence.

Separately, `state.rs:863` stamped the stale code hash onto every invoice, naming an address contract that was never published — unpayable, and nothing about it looked wrong.

## Verification of the values

Not taken from the bridge's self-report alone, since this file's own docs warn against trusting a bridge for addressing:

- the bridge's `/v1/status` reports both;
- `freenet-bitcoin/legacy/address_contract.toml` lists `3b5f1df2…` as **A3**, in a file whose header says it records *only superseded* generations, with A8 = `cd2ae741…`;
- `fdev diagnostics` lists both tip contracts as distinct live keys on the node — two real contracts, and the build watched the one nothing writes to.

## The gate

`scripts/check-bitcoin-constants.sh` compares the pinned constants against a running bridge; `publish-harvest` calls it before publishing. That is the moment the cost is incurred and the only place it *can* run — CI has no bridge.

Three exit codes, deliberately distinct:

| code | meaning | publish task |
|---|---|---|
| 0 | checked, match | proceeds |
| 1 | checked, disagree — or a constant could not be read | refuses |
| 2 | could **not** check (no bridge) | refuses, unless `HARVEST_SKIP_BRIDGE_CHECK=1`, which warns loudly |

`2` is separate from `0` on purpose: recording *"I could not look"* as *"it is fine"* is the exact defect being guarded against.

Extraction **fails closed** — rename or move a constant and the script says its own extraction has drifted and exits 1, rather than comparing empty to empty and calling it a match. That is how a check like this normally rots.

## Testing

Verified by making it **fail**, not only pass — the lesson from #27 in this same series, where a success gate was watched working once on the one path it would never meet again.

| case | result |
|---|---|
| stale constants, exactly as shipped | **exit 1**, both mismatches named — it would have caught #30 |
| unreachable bridge | **exit 2**, states the constants were NOT checked |
| constant renamed (extraction drift) | **exit 1**, "extraction has drifted" |
| current constants | exit 0 |

And the publish wiring against a stubbed check: `0` publishes, `1` refuses, `2` refuses with instructions, `2` + override warns and proceeds.

Existing `bitcoin_config` tests still pass (4/4) — as they did *before* this fix, which is the documented gap: they assert well-formedness only.

## What this is not

A stopgap. It works only where a bridge is reachable, and it trusts that bridge to describe itself. The durable answer is a **pointer record** read over the node like any other contract. `freenet/ghostkeys` already runs that pattern end to end — `pointer-records.toml`, a signer script, and a CI `check-pointer-freshness` job that fails a PR when an artifact's WASM changed and no new record was signed. That gate is exactly what is missing here, and `freenet-bitcoin` already ships `generation/pointer-v1.wasm` but has no records config.

#30 stays open for that.

[AI-assisted - Claude]

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -136,6 +136,42 @@ set -e
 export FREENET_NODE_URL
 echo "Publishing to $FREENET_NODE_URL"
 
+# The Bitcoin ADDRESSING constants are pinned at build time (the gateway CSP
+# forbids asking a bridge at runtime -- harvest#29), so a contract rebuild
+# re-keys them and this build goes quietly wrong. That happened: harvest#30,
+# stale by five generations, and it did NOT fail silent. The superseded tip
+# contract still exists and still holds its last state, so the published app
+# showed a chain tip ~400 blocks old as though it were live, and every invoice
+# named an address contract that was never published.
+#
+# So check before publishing, which is the moment the cost is incurred. This
+# cannot live in CI -- there is no bridge there.
+#
+# Exit 2 means "could not check" (no bridge reachable) and is NOT treated as a
+# pass, because recording "I could not look" as "it is fine" is the exact
+# defect being guarded against. Override deliberately if you must.
+set +e
+./scripts/check-bitcoin-constants.sh
+CHECK_RC=$?
+set -e
+if [ "$CHECK_RC" -eq 1 ]; then
+  echo "" >&2
+  echo "error: refusing to publish -- see the mismatch above." >&2
+  exit 1
+elif [ "$CHECK_RC" -eq 2 ]; then
+  if [ "${HARVEST_SKIP_BRIDGE_CHECK:-}" = "1" ]; then
+    echo "WARNING: publishing WITHOUT having checked the Bitcoin constants" >&2
+    echo "         (HARVEST_SKIP_BRIDGE_CHECK=1). If they are stale, this build" >&2
+    echo "         will show stale chain data as live. See harvest#30." >&2
+  else
+    echo "" >&2
+    echo "error: could not verify the Bitcoin constants against a bridge, so" >&2
+    echo "       refusing to publish. Start a bridge, set HARVEST_BRIDGE_URL," >&2
+    echo "       or set HARVEST_SKIP_BRIDGE_CHECK=1 to publish unverified." >&2
+    exit 1
+  fi
+fi
+
 # fdev EXITS 0 ON A FAILED PUBLISH -- observed 2026-09-06, printing
 # "Error: Failed to receive response" and returning 0. So `set -e` does not
 # catch it and the success line below would print over a failure. Confirm from
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -170,6 +170,20 @@ elif [ "$CHECK_RC" -eq 2 ]; then
     echo "       or set HARVEST_SKIP_BRIDGE_CHECK=1 to publish unverified." >&2
     exit 1
   fi
+elif [ "$CHECK_RC" -ne 0 ]; then
+  # ANY other exit code means the check did not run to a verdict at all: not
+  # executable (126), not found or bad interpreter (127), killed by a signal
+  # (128+n). Without this branch those fall through to publish, which is the
+  # precise failure the check exists to prevent -- "I could not look" recorded
+  # as "it is fine" -- reintroduced one level up in the wrapper that calls it.
+  #
+  # Found in review. Trigger is mundane: a checkout that drops the executable
+  # bit (core.fileMode=false) silently defeats the entire gate.
+  echo "" >&2
+  echo "error: scripts/check-bitcoin-constants.sh did not run to a verdict" >&2
+  echo "       (exit $CHECK_RC). Refusing to publish rather than assuming the" >&2
+  echo "       constants are fine. Check the script is present and executable." >&2
+  exit 1
 fi
 
 # fdev EXITS 0 ON A FAILED PUBLISH -- observed 2026-09-06, printing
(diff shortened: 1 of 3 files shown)


sanity:
## Review record

**Tier: Full.** Touches release tooling (the publish gate) and the constants that decide which Bitcoin contracts the app talks to.

### Lenses

| lens | outcome |
|---|---|
| Independent blind review (separate session, read-only) | **DO NOT MERGE** — one blocking finding, fixed in c6d7331 |
| Guard exercised against real failure cases | passes, tables below |
| **External non-Claude (`codex review`)** | **UNAVAILABLE** — usage limit, resets 2026-09-07 08:14; extra Claude lens substituted per the review rule |

### Blocking finding — the gate was fail-OPEN at the level it could not see

The `CHECK_RC` dispatch handled `1` and `2` and had **no `else`**. Any other exit code — `126` not executable, `127` not found or bad interpreter, `128+n` killed by a signal — matched neither branch and fell straight through to `fdev publish` with no warning.

That is precisely the failure the script's own comments say must never happen. The script distinguishes exit `2` from exit `0` specifically so that *"I could not look"* is never recorded as *"it is fine"* — and the wrapper calling it reintroduced exactly that, one level up.

The trigger is mundane, not exotic: a checkout that drops the executable bit (`core.fileMode=false`) silently defeats the whole gate, and the publish looks entirely normal.

Fixed: any unrecognised exit code is a refusal.

| stubbed check exit | before | after |
|---|---|---|
| 0 | publishes | publishes |
| 1 (mismatch) | refuses | refuses |
| 2 (could not check) | refuses | refuses |
| 2 + `HARVEST_SKIP_BRIDGE_CHECK=1` | warns, publishes | warns, publishes |
| script not executable | **published silently** | refuses |
| script absent | **published silently** | refuses |

### On the constant values — the subtle part, independently confirmed

`cd2ae741…` (A8) **is** listed in `legacy/address_contract.toml`, whose header says it records *only superseded* generations. Taken at face value that would make this fix wrong.

It is not, and the reviewer found a better citation than the one I had. freenet-bitcoin's own commit `33204f2` states it outright:

> *"the fold tie-break fix re-keys the address contract (cd2ae741… → c2273660…) … the old id stays live until the rebuilt WASM is deployed, which is a separate, deliberate act."*

So `c2273660…` is real but **deployed nowhere**. Pinning it would recreate this issue's exact symptom in the other direction — a correct-looking constant naming a contract nobody writes to. **Harvest must pin what is deployed**, which is A8.

Consequence worth stating: freenet-bitcoin's deployment is behind its own `main`, and redeploying the bridge needs a matching Harvest bump in the same window. Recorded on #30; it is the argument for pointer records being the real fix, since two independently-released repositories cannot hard-code each other's content-addressed identity without a wrong window.

### Verified by making it fail, not only pass

The lesson from #27 in this series, where a success gate was watched working once on the one path it would never meet again.

| case | result |
|---|---|
| stale constants, exactly as shipped | **exit 1**, both mismatches named — would have caught #30 |
| unreachable bridge | **exit 2**, states the constants were NOT checked |
| constant renamed (extraction drift) | **exit 1**, "extraction has drifted" |
| current constants | exit 0 |

### Non-findings, checked and cleared

- **Extraction correctness.** Both `sed` pipelines run against the real file and return the right values. The old `B24HMUFa…` appears only in wrapped prose, never in the `BitcoinNetwork::Signet => Some("…")` form the pattern anchors on; the other arms are `=> None`, so there is no wrong-arm ambiguity. The `ADDRESS_CONTRACT_CODE_HASH_HEX` range starts at that constant rather than the `TRUSTED_BRIDGE_ID_BS58` above it.
- **`set +e` / `CHECK_RC` / `set -e`** is the correct idiom for capturing the exit code under `cargo-make`.
- **Comment accuracy.** "~400 blocks stale" checks out (320955 − 320557 ≈ 398); the A3→A8 five-generation gap matches the registry; the "trust right, addressing wrong" split matches the diff, with `TRUSTED_BRIDGE_ID_BS58` unchanged and equal to the bridge's reported `bridge_id`.

[AI-assisted - Claude]

PR #33, opened 2026-09-06
fix(publish): clean the dx output and gate the archive against stale copies

### LINKED ISSUE #4: Published webapp bundle carries 10 stale UI WASM builds (37.7 MB uncompressed)
## What

The live Harvest web container (`6FzSeAUKcqJrveKyU8RJgGKc5jRB1Z2juvxXtwTA4Em9`) ships **ten** copies of the UI WASM. Fetched from the network and unpacked:

```
  898284  assets/harvest-ui_bg-dxh1ae83b4b721d32c.wasm
  973356  assets/harvest-ui_bg-dxh23638be3676654f0.wasm
 6105355  assets/harvest-ui_bg-dxh30ce78a2bebf7c54.wasm
 6235177  assets/harvest-ui_bg-dxh34e3c83329a052.wasm
 2879868  assets/harvest-ui_bg-dxh64ed56db27b719.wasm
  900327  assets/harvest-ui_bg-dxh663a5a9fcd9763a.wasm
 6233254  assets/harvest-ui_bg-dxha26b9f65b2e92f2f.wasm
 6285963  assets/harvest-ui_bg-dxhc7f633f1fd6fca8.wasm
  970531  assets/harvest-ui_bg-dxhe116274957854f88.wasm
 6234921  assets/harvest-ui_bg-dxhfed22f3cfd2ecbb.wasm
--------
37717036 bytes across 10 copies
```

Nine of those are dead weight — only one is the build the app actually loads. For comparison, fetched the same way on the same day:

| app | UI WASM copies in the live bundle | compressed bundle |
|---|---|---|
| Delta | 1 | 528,008 B |
| River | 1 | 1,083,052 B |
| **Harvest** | **10** | 1,837,144 B |

## Why it happens

`dx` writes **content-hashed** filenames, so each build produces `harvest-ui_bg-<newhash>.wasm` rather than overwriting the previous one. If the output directory is not cleaned before the archive is created, every prior build stays in the tree and gets tarred in — so the bundle grows by a full UI WASM on every publish.

This is the same bug Delta hit and fixed in **delta#70**, with `scripts/check-webapp-bundle.sh`: a clean-and-check inside the bundling task that refuses an archive carrying stale copies from earlier builds. Harvest has no equivalent, and no CI at all, so nothing has been in a position to notice.

## Impact

Every user who loads Harvest pulls this contract. The compressed penalty is smaller than the 37.7 MB uncompressed figure suggests (the duplicates are similar and xz deduplicates well), which is precisely why it has gone unnoticed for ten publishes — but it grows without bound, one UI build per publish, and it is entirely avoidable.

## Suggested fix

Port Delta's approach:

1. Clean the `dx` output directory before building the archive.
2. Add a `check-webapp-bundle.sh` equivalent that fails if the archive contains more than one `assets/*_bg-*.wasm`, and wire it into the bundling task *before* the sign/publish step, so a bad archive can never be signed.

The check is worth more than the clean on its own — a clean step can be skipped or reordered, whereas a gate that refuses the archive cannot be skipped silently.

## How this was found

Incidentally, while assessing Harvest for the pointer-record rollout ([freenet-core#5194](https://github.com/freenet/freenet-core/issues/5194)) — I fetched the live container to determine whether Harvest's contracts are actually on the network. They are: `store_contract.wasm`, `reputation_contract.wasm`, `mailbox_contract.wasm` and `harvest_delegate.wasm` all ship inside it, which means third parties can pin them and Harvest is a genuine pointer candidate.

Separately worth knowing: the working tree currently has uncommitted rebuilt WASMs for all four contracts plus the vendored ghostkey delegate, so what is on disk is not what was published.

[AI-assisted - Claude]

sanity:
Fixes #4.

## Problem

`dx` writes **content-hashed** asset filenames, so a rebuild lands *beside* its predecessor rather than replacing it. Nothing pruned the old one and `compress-webapp` tars the whole directory, so the bundle was append-only across builds — one full UI WASM per publish, indefinitely.

The live container was measured carrying **ten** copies, 37.7 MB uncompressed, where Delta and River carry one each.

The wasted bytes are the smaller cost. With several WASMs in the bundle, *"did my fix ship?"* can no longer be answered by grepping the bundle, because a grep for a new symbol hits a stale copy just as readily as the live one.

## Approach

Two changes, and the second is the one that lasts.

**1. `build-ui` wipes the dx output directory before building.** Safe because everything there is regenerated: dx re-copies `ui/public/*` (where `sync-wasm` stages `contracts/`), and the two fixed-name assets are re-copied after. Verified against a real build — contracts are present and correct in the output.

**2. `compress-webapp` gates the archive**, before `sign-webapp` can see it. An archive that has been signed already has an identity; refusing it afterwards is a worse place to find out.

**The clean should make the gate unnecessary. That is exactly why the gate is there.** A clean step can be skipped, reordered, or quietly broken by a later edit; a gate that refuses the archive cannot be skipped silently. The clean is the fix; the gate is what notices when the fix stops working. The issue makes this point and it is right.

## Why reachability, not a copy count

Ported from `delta/scripts/check-webapp-bundle.sh` (delta#70, delta#46), written against identical dx behaviour.

It asserts **transitive reachability from `index.html`** rather than "exactly one WASM":

- a count cannot distinguish a stale hashed orphan from the live one — which *is* the bug;
- reachability also catches an orphaned favicon, a stale `index.html` pointing at an old loader, a root-level orphan, and nested leftovers;
- a hardcoded "exactly 1" would break on a future dx that legitimately emits several chunks.

It **additionally** requires every non-fixed-name file to carry a dx content hash. Neither check is redundant: in delta, dx's staged *unhashed* `wasm-bindgen` output was laundered into "reachable" by a substring coincidence with the loader, which legitimately contains wasm-bindgen's default name. Hash-shape catches that by construction, whatever the matcher is later changed to.

Harvest's fixed-name set differs from delta's: `index.html`, `harvest.css`, `harvest-logo.svg`, and five `contracts/*.wasm`. `harvest-logo.svg` is REQUIRED rather than a reachability candidate because it is not content-hashed and is named from `harvest.css`.

## Testing

Verified by making it **fail**, not only pass:

| case | result |
|---|---|
| stale extra WASM beside the live one — **the #4 bug exactly** | refuses, naming the orphan and its size |
| unhashed stray (dx staged-output shape) | refuses |
| `contracts/store_contract.wasm` removed | refuses |
| real clean build via `cargo make compress-webapp` | **passes** — 10 files, 8 required, 2 reachable, 1.2M compressed |

The last row is a genuine end-to-end run, not a synthetic archive.

## Noticed while running this, not fixed here

The build log shows:

```
wasm-opt failed with status code signal: 6 (SIGABRT) (core dumped)
stdout: compile unit size was incorrect (this may be an unsupported version of DWARF)
```

dx continues and reports "Client build completed successfully", so **the shipped UI WASM is unoptimised**. It is 7,283,353 bytes — byte-for-byte the same size as the previous build, which indicates this has been failing on every build, not just this one.

Out of scope here; filed separately so it is not lost.

[AI-assisted - Claude]

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -37,6 +37,19 @@ description = "Build the Dioxus UI"
 dependencies = ["sync-wasm"]
 script = '''
 set -e
+# WIPE THE OUTPUT DIRECTORY FIRST.
+#
+# dx writes CONTENT-HASHED asset filenames, so a rebuild lands BESIDE its
+# predecessor instead of replacing it. Nothing prunes the old one and
+# compress-webapp tars the whole directory, so without this the bundle is
+# append-only across builds -- it grows a full UI wasm on every publish. The
+# live container was measured carrying TEN copies, 37.7 MB uncompressed, where
+# Delta and River carry one each (harvest#4).
+#
+# Safe to wipe: everything here is regenerated. dx re-copies ui/public/* (which
+# is where sync-wasm stages contracts/), and the two fixed-name assets are
+# re-copied below.
+rm -rf target/dx/harvest-ui/${BUILD_PROFILE}/web/public
 cd ui && dx build --${BUILD_PROFILE}
 # dx doesn't copy CSS from assets/ to the build output root, so copy
 # manually. Path is relative to `ui/` because we cd'd above and the
@@ -68,7 +81,19 @@ dependencies = ["build-ui"]
 script = '''
 mkdir -p target/webapp
 cd target/dx/harvest-ui/${BUILD_PROFILE}/web/public && tar -cJf ../../../../../webapp/webapp.tar.xz *
+cd - > /dev/null
 echo "Compressed webapp to target/webapp/webapp.tar.xz"
+
+# Gate the ARCHIVE, not the build directory, and do it here so a bad archive
+# can never reach sign-webapp -- an archive that has been signed has already
+# been given an identity, and refusing it afterwards is a worse place to find
+# out.
+#
+# The clean above should make this unnecessary. That is exactly why it is here:
+# a clean step can be skipped, reordered, or quietly broken by a later edit,
+# whereas a gate that refuses the archive cannot be skipped silently. The clean
+# is the fix; this is the thing that notices when the fix stops working.
+./scripts/check-webapp-bundle.sh target/webapp/webapp.tar.xz
 '''
 
 [tasks.sign-webapp]
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -93,7 +93,8 @@ echo "Compressed webapp to target/webapp/webapp.tar.xz"
 # a clean step can be skipped, reordered, or quietly broken by a later edit,
 # whereas a gate that refuses the archive cannot be skipped silently. The clean
 # is the fix; this is the thing that notices when the fix stops working.
-./scripts/check-webapp-bundle.sh target/webapp/webapp.tar.xz
+./scripts/check-webapp-bundle.sh target/webapp/webapp.tar.xz \
+  "target/dx/harvest-ui/${BUILD_PROFILE}/web/public"
 '''
 
 [tasks.sign-webapp]
--- a/scripts/check-webapp-bundle.sh
+++ b/scripts/check-webapp-bundle.sh
@@ -1,0 +1,179 @@
+#!/bin/bash
+# Publish gate: refuse a webapp archive that is not self-consistent.
+#
+# WHY THIS EXISTS
+#
+# dx writes CONTENT-HASHED asset filenames, so a changed build lands beside its
+# predecessor rather than replacing it. Nothing prunes the old one and the
+# publish task tars the whole directory, so the bundle is append-only across
+# builds: it grows a full UI wasm on every publish, indefinitely. The live
+# Harvest container was measured carrying TEN copies, 37.7 MB uncompressed,
+# where Delta and River carry one each (harvest#4).
+#
+# The wasted bytes are the smaller cost. With several wasms in the bundle, "did
+# my fix ship?" can no longer be answered by grepping the bundle, because a grep
+# for a new symbol hits a stale copy just as readily as the live one. Checking
+# reachability here is what keeps grep-the-bundle an honest deploy check.
+#
+# Ported from delta's scripts/check-webapp-bundle.sh (delta#70, delta#46),
+# which was written against the identical dx behaviour. Both incidents its
+# comments describe are load-bearing; read them before simplifying anything
+# here.
+#
+# WHAT IT ASSERTS
+#
+# Every file in the archive is accounted for, in one of exactly two ways:
+#
+#   REQUIRED  - a fixed-name file the app needs at runtime. These cannot be
+#               covered by reachability: contracts/*.wasm are fetched via a path
+#               the app builds at run time, so their names appear nowhere in the
+#               bundle. Their absence is checked directly instead.
+#
+#   ACCOUNTED - carries a dx content hash in its name (check 0) AND is
+#               referenced, transitively, from index.html (check 1).
+#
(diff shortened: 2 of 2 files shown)


sanity:
## Review record

**Tier: Full** (release tooling — this gates what gets signed and published).

| lens | outcome |
|---|---|
| Independent blind review (separate session, read-only) | **MERGE**, nothing blocking; one nit, acted on |
| Gate exercised against real failure cases | passes, tables below |
| **External non-Claude (`codex review`)** | **UNAVAILABLE** — usage limit, resets 2026-09-07 08:14; extra Claude lens substituted per the review rule |

### The nit I did not accept as harmless (6e24c3b)

The review flagged that the ported script's `BUILD_DIR` hardcoded `release` rather than `${BUILD_PROFILE}`, and judged it harmless since Harvest's profile is always `release`. That is true today, and I closed it anyway — for the **shape**, not the current impact.

The staleness comparison is wrapped in `if [ -d "$BUILD_DIR" ]`. Under any other profile that path would not exist, so the guard would not error — it would **skip, silently**, while the script still printed `Bundle OK` and exited 0.

A guard that measures nothing while looking like it passed is the defect this PR exists to answer. Leaving a second instance of it *inside the fix* is not a trade worth a one-line saving.

The build directory is now passed by the caller (`compress-webapp` passes the directory it actually tarred), and when passed it **must** exist — the caller just built it, so absence is an error, not a reason to skip. The guess-and-skip path survives only for running the script by hand against an archive fetched from elsewhere, and now says out loud that it skipped.

| case | before | after |
|---|---|---|
| build dir matches archive | exit 0 | exit 0 |
| build dir does not exist | **skipped silently** | exit 1 |
| archive valid but stale vs build dir | exit 1 | exit 1, names the directory |
| no build dir (ad-hoc use) | skipped silently | exit 0, prints that it skipped |

Worth noting how nearly that test proved nothing: my first attempt at the stale case removed `harvest.css`, which fails the REQUIRED check long before the staleness comparison runs. Testing the third row needs an archive that is otherwise **valid** but whose file list differs from the build tree.

### The gate itself, verified by making it fail

| case | result |
|---|---|
| stale extra WASM beside the live one — **the #4 bug exactly** | refuses, naming the orphan and its size |
| unhashed stray (dx staged-output shape) | refuses |
| `contracts/store_contract.wasm` removed | refuses |
| real `cargo make compress-webapp` | **passes** — 10 files, 8 required, 2 reachable, 1.2M |

### Cleared by the review, with evidence

- **`rm -rf` safety.** Confined to `target/dx/harvest-ui/${BUILD_PROFILE}/web/public`; `web/` holds only `public/`, no adjacent cache state. `BUILD_PROFILE` is set in `[env]`, and even an empty override degrades to a double-slash path still under `target/dx/harvest-ui/`. Everything wiped is regenerated: `sync-wasm` stages contracts into `ui/public/contracts/` *before* the wipe, dx re-copies `ui/public/*`, and the two fixed-name assets are copied back straight after `dx build`.
- **REQUIRED list matches the real archive** — the eight non-hashed files, exactly.
- **`cd - > /dev/null`.** cargo-make runs a task's script as one shell with cwd = where `cargo make` was invoked, so `cd -` returns correctly; and cargo-make halts on nonzero exit, so a failing gate genuinely blocks `sign-webapp` rather than printing a warning. Confirmed empirically, not assumed.
- **Port cleanliness** — no leftover delta strings, hash regex unchanged, REQUIRED and `BUILD_DIR` retargeted.

### Found while running this, filed not fixed

`wasm-opt` aborts (SIGABRT, DWARF version) on every UI build and dx reports success anyway, so the shipped WASM is unoptimised at 7,283,353 bytes — byte-identical to the previous build, which is what an un-run optimiser looks like. Filed as #34 with a suggested acceptance test that measures the size rather than grepping for the error string.

[AI-assisted - Claude]

PR #35, opened 2026-09-06
fix(publish): resolve the build directory before chdir, so the gate can find it

sanity:
Follow-up to #33, which broke the publish it was meant to protect.

## Problem

The gate refused the first real publish after it landed:

```
FAILED: build directory target/dx/harvest-ui/release/web/public does not exist
        -- cannot verify the archive matches the build it came from
Error while executing command, exit code: 1
```

on a directory that plainly existed.

`compress-webapp` passes that path **relative to the repo root**. The script then `cd`s into its temp extraction directory before evaluating it, so the relative path resolved against `$WORK` instead.

The archive argument already had exactly this treatment — converted to an absolute path, with the comment *"Absolute: the checks below run from inside a temp extraction dir."* The build-directory argument, added one commit later, did not get it.

**One absolute-path conversion is never the whole fix.** Every caller-supplied path that crosses that `cd` needs it, and I converted one of two.

## What went right, and is worth keeping

The gate **failed closed** on a condition it could not evaluate. It refused the publish and stopped before `sign-webapp`, so:

- nothing was signed,
- no version counter was consumed,
- no bundle reached the network.

That is precisely the behaviour the previous commit chose deliberately over `if [ -d … ]` skipping — and the choice was vindicated within the hour, against its own author. Had it kept the silent-skip form, this bug would have disabled the staleness check permanently and quietly, and the publish would have looked completely normal.

## Testing

| case | result |
|---|---|
| relative build dir from repo root — **the case that failed** | passes |
| absolute build dir | passes |
| build dir genuinely missing | fails |
| archive stale vs build dir, via a *relative* path | fails, naming the resolved absolute path |

The last row matters: it confirms the fix did not trade a false failure for a silent skip. The message now reports the resolved location rather than a relative string that means nothing from where the script is standing.

[AI-assisted - Claude]

### DIFF
--- a/scripts/check-webapp-bundle.sh
+++ b/scripts/check-webapp-bundle.sh
@@ -43,6 +43,23 @@ ARCHIVE="${1:?usage: check-webapp-bundle.sh <webapp.tar.xz> [build-dir]}"
 # Absolute: the checks below run from inside a temp extraction dir.
 ARCHIVE="$(cd "$(dirname "$ARCHIVE")" && pwd)/$(basename "$ARCHIVE")"
 
+# The build directory needs the SAME treatment, for the same reason, and
+# forgetting it is how this script first failed in anger: the caller passes a
+# path relative to the repo root, the script then cd's into $WORK, and the
+# relative path resolves against the temp directory instead. The staleness
+# check reported "build directory does not exist" on a directory that plainly
+# did, and refused the publish.
+#
+# It refused rather than skipping, which is the whole design working, but the
+# lesson is that ONE absolute-path conversion is never the whole fix -- every
+# caller-supplied path crossing that `cd` needs it.
+if [ -n "${2:-}" ]; then
+    [ -d "$2" ] || { echo "FAILED: build directory $2 does not exist -- cannot verify the archive matches the build it came from"; exit 1; }
+    BUILD_DIR_ARG="$(cd "$2" && pwd)"
+else
+    BUILD_DIR_ARG=""
+fi
+
 WORK="$(mktemp -d)"
 trap 'rm -rf "$WORK"' EXIT
 tar -xJf "$ARCHIVE" -C "$WORK"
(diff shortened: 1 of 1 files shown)


sanity:
## Review record

**Tier: Light-plus** — a one-file fix to release tooling, reviewed as release tooling.

| lens | outcome |
|---|---|
| Independent blind review (separate session, read-only) | **MERGE**, nothing blocking |
| Behaviour verified empirically | see below |

### Confirmed independently

1. **No other path crosses the `cd` unresolved.** Every path used after `cd "$WORK"` is either deliberately relative to `$WORK` (the archive's own contents — `find .`, the grep and `stat` targets, `du -sh .`) or absolute before the `cd` (`$ARCHIVE`, now `$BUILD_DIR_ARG`). The fallback guess derives from the already-absolute `$ARCHIVE`, so it is unaffected by cwd. Enumeration complete.

2. **No silent-skip gap.** `BUILD_DIR_ARG` is empty **iff** `$2` was empty. When `$2` is set, the block either produces a validated absolute path or exits non-zero. There is no path where the caller asks for the staleness comparison and it is quietly skipped — the specific thing this gate must never do.

3. **`set -e` handles an unreadable build directory correctly** — verified by running it, not by reasoning about shell semantics. A directory that exists but cannot be entered prints `cd: Permission denied` and exits non-zero; it does **not** yield an empty `BUILD_DIR_ARG` that would silently disable the check.

Non-blocking note from the review: the later `[ -d "$BUILD_DIR" ]` is now redundant for the `$2` case, though still needed for the fallback branch. Left alone.

### Testing

| case | result |
|---|---|
| relative build dir from repo root — **the case that failed** | passes |
| absolute build dir | passes |
| build dir missing | fails |
| build dir present but not traversable | fails loudly |
| archive stale vs build dir, via a *relative* path | fails, naming the resolved absolute path |

The last row is the one that matters: it confirms the fix did not trade a false failure for a silent skip.

[AI-assisted - Claude]

PR #37, opened 2026-09-06
docs(ci): correct three comments that assert things which are not true

sanity:
Three comments that assert things which are not true. Found by review reports that arrived after their PRs had merged.

## 1. "fdev EXITS 0 ON A FAILED PUBLISH" — false, and my error rather than fdev's

I read the exit status of a **pipeline ending in `tail`** and attributed it to fdev. Reproduced both ways against a dead port:

```
fdev publish ... ; echo $?             -> 1
fdev publish ... | tail -1 ; echo $?   -> 0     <- tail's status
```

fdev 0.3.295 maps a `ResponseTimeout` to 3, a conformance violation to 4, and everything else — *including* `"Failed to receive response"` — to 1 (`src/main.rs:52-72`, `src/commands.rs:298-307`).

This is the "check the operation's own output" rule, failed at the exact point of asserting what an operation did. A testable claim that is wrong is worse than no comment: the next reader tries it and finds it false, and then wonders what else here is.

**The guard stays, for a better reason than the false one.** fdev has two success arms and several failure arms, so an exit code alone cannot answer *"did the contract land"*. The output check distinguishes them and fails closed on wording it does not recognise.

## 2. `|| true` discarded fdev's deliberate exit 3

`EXIT_RESPONSE_TIMEOUT = 3` exists so automation can treat a timeout as **"submitted, may have succeeded"** rather than a hard failure. Flattening it meant a timeout printed `publish did NOT succeed` over a publish that may well have landed — pointing the reader at a blind re-publish instead of at the served-hash comparison that would actually settle it.

The exit code is now captured and a timeout says so explicitly. Verified under `set -e` that codes 0, 1 and 3 are all captured with output preserved and the script surviving:

```
if OUT=$(fdev publish ...); then FDEV_RC=0; else FDEV_RC=$?; fi
```

## 3. `dependabot.yml` claimed guards that exist do not

It still said:

> *"This repo has NO migration or wasm-hash guard workflow, so nothing here will turn red when that happens. Before merging a cargo PR, rebuild and compare the contract key…"*

Both guards exist. `contract-drift.yml` builds both sides of the PR and compares **addresses** — and is a required check — while `ci.yml`'s wasm-staleness compares the committed bytes. The comment predates them and was left behind when they landed.

**This is the worst of the three despite being pre-existing:** a stale comment about a guard, living in the file that warns about stale guards, instructing a human to do by hand what CI now does. Replaced with what the two workflows actually cover and what still genuinely needs a human — the *decision* to acknowledge an intended re-key and ship a migration.

## Also: the cost note was wrong in kind

`contract-drift.yml`'s comment said the double build is paid "on pull requests". It is paid per **event** — `labeled`/`unlabeled` are in `types:` and were previously gated by the `paths:` filter, so every label add or remove now triggers a run.

Replaced the hand-wave ("rust-cache absorbs most of it") with the measured figure — **2:00–2:25 warm-cached** across the fifteen most recent runs — and recorded why there is deliberately no `concurrency:` group: on a required check, `cancel-in-progress` leaves a cancelled non-success check that blocks the merge, which is worse than a few wasted runner-minutes.

## Testing

- `Makefile.toml` parses; extracted script passes `bash -n`.
- Exit-code capture exercised for 0, 1, 3 under `set -e`.
- Both YAML files parse.
- fdev's real exit behaviour reproduced directly, not read.

No behaviour change beyond the exit-3 message; the rest is comments that were actively misleading.

[AI-assisted - Claude]

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -84,10 +84,27 @@ updates:
 # contract and strand existing user data. ANY wasm change moves the address: a
 # direct bump, a transitive bump, or a rustc upgrade.
 #
-# This repo has NO migration or wasm-hash guard workflow, so nothing here will
-# turn red when that happens. Before merging a cargo PR, rebuild and compare the
-# contract key, and ship a migration if it moved. See the `freenet-app-migration`
-# skill and `.claude/rules/` in the freenet tree.
+# CI DOES turn red when that happens, and this comment used to say the opposite.
+# It was written before the guards existed and was left in place after they
+# landed, telling readers to do by hand what CI now does -- which is the exact
+# defect these guards are for, sitting in the file that warns about it.
+#
+# Two workflows cover it, and they are complementary rather than redundant:
+#
+#   * `.github/workflows/contract-drift.yml` -- "no contract or delegate address
+#     moved". Builds the contracts and the delegate on BOTH sides of the PR and
+#     compares their ADDRESSES, so it catches a transitive bump visible only as
+#     Cargo.lock churn, and a parameter-struct edit that re-keys with the WASM
+#     byte-identical. It runs on every pull request and is a REQUIRED check.
+#
+#   * `.github/workflows/ci.yml` -- "committed contract wasm matches a fresh
+#     build". Rebuilds and compares the committed artifacts, which is the
+#     different question of whether the checked-in bytes are stale.
+#
+# So a re-key is a red check with the old and new addresses printed. What still
+# needs a human is the DECISION: when the re-key is intended, label the PR
+# `contract-rekey-acknowledged`, and ship a migration if state was stranded. See
+# the `freenet-app-migration` skill and `.claude/rules/` in the freenet tree.
 #
 # Dependabot does not create this hazard -- a hand-run `cargo update` does the
 # same thing -- but it does deliver a steady stream of routine-LOOKING PRs, and
--- a/.github/workflows/contract-drift.yml
+++ b/.github/workflows/contract-drift.yml
@@ -87,8 +87,17 @@ on:
     # request, builds both sides to identical bytes and reports green, which is
     # the correct answer rather than the absent one.
     #
-    # The cost is real and measured: 2:00-2:25 per run with a warm
-    # `Swatinem/rust-cache`, sampled across the fifteen most recent runs.
+    # The cost is real. Measured across the 20 most recent completed runs of
+    # this workflow (`gh run list --workflow "contract drift"`, wall-clock from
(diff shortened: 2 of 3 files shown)


sanity:
## Review record

**Tier: Full** (release + CI configuration). Independent blind review, read-only, separate session.

| lens | outcome |
|---|---|
| Independent blind review | **DO NOT MERGE** — one blocking finding, fixed in 119cd48 |

### Blocking finding — I committed this PR's own defect inside the PR

The comment I added claimed the drift job costs **"2:00-2:25 ... sampled across the fifteen most recent runs."** That is false.

Measured myself after the review flagged it (`gh run list --workflow "contract drift"`, 20 most recent completed runs, wall-clock):

```
1:26 2:27 2:25 2:32 1:27 2:18 1:25 2:09 1:27 2:26
1:24 2:26 2:04 2:25 2:23 0:59 2:03 2:09 2:18 2:23

min 0:59   median 2:18   max 2:32
```

Bimodal — a cluster near 1:25 and one near 2:20 — so no narrow band describes it, however centred.

**The figure came from a reviewer's summary and I wrote it into a comment as though I had measured it.** That is the same defect as the four this PR corrects, committed inside the correction. Recorded in the comment itself, because it is the more useful half: repeating a number is not verifying it, and a number carries more apparent authority than the prose around it.

Now states the range, the median, the method, and that the distribution is bimodal so the next reader does not re-narrow it.

### Verified true, against source rather than my say-so

1. **fdev exit codes** — `main.rs:52-72`, `commands.rs:298-307`. Timeout→3, conformance→4, everything else including `"Failed to receive response"`→1. No path exits 0 on failure.
2. **The pipeline explanation** — reproduced independently: `false; echo $?` → 1; `false | tail -1; echo $?` → 0.
3. **PutResponse vs UpdateResponse depends on node mode** — confirmed in freenet-core. `operations/put/op_ctx_task.rs` constructs only `PutResponse` (zero `UpdateResponse` in the file); `contract/executor/runtime/contract_ops.rs` is reached from `contract_requests`, whose own comment says local-node mode "PUTs straight through here", and returns `UpdateResponse` on merge into an already-held contract. This is the claim I most wanted checked rather than agreed with, since I could not have asserted it from Harvest alone.
4. **Exit-code capture** — the `if OUT=$(cmd); then RC=0; else RC=$?; fi` shape tested in **both bash and dash**; captures correctly, and `set -e` does not abort because the command sits in an `if` condition.
5. **The dependabot.yml rewrite** — both workflow descriptions match what the files do, and the required-check status was re-confirmed against the live ruleset.

[AI-assisted - Claude]

Issue #5, opened 2026-08-19
Vendored ghostkeys delegate is six generations stale; resolve the pointer instead


sanity:
Not urgent: harvest is a dormant prototype and, as far as anyone knows, nobody holds a link to its container contract. Filing it so it is waiting rather than rediscovered.

## The problem

`ui/src/components/app.rs:60` vendors a copy of ghostkeys' delegate and registers it:

```rust
let gk_wasm = include_bytes!("../../public/contracts/ghostkey_delegate.wasm");
match crate::gateway::register_delegate(gk_wasm).await {
```

A delegate's address is `BLAKE3(BLAKE3(wasm) || params)`, so a vendored copy is a hardcoded key with an extra step: it pins whatever generation was current on the day it was copied, and goes stale the next time ghostkeys rebuilds.

That has already happened. The vendored copy is **six generations behind**, roughly three and a half months:

| | code hash |
|---|---|
| harvest, `origin/main` | `6c42cdb7bee8892fcff67cd8f8f70dd741518785bb3036d22fd7f5be0682b9d1` |
| harvest, local working tree | `41820886ca1db199eca2a2b229c6051d1120ea909ddbe1bd3653bf1cd586f540` |
| ghostkeys, current (2026-08-12) | `0d0e2043c5242953b6e8705179ca6847f1d83fce8e23a4ae314118d9b22ffd62` |

Both harvest copies appear in ghostkeys' `legacy_delegates.toml`, so both are retired generations rather than anything current.

## Why it matters when harvest wakes up

The failure is silent and misleading rather than an error. harvest registers the stale WASM, gets a delegate at the old key, and every request comes back against an empty namespace. A user who owns a Ghost Key is told they have none, and no error is raised anywhere, because at the protocol level "this user has no data" and "the delegate you named no longer exists" look identical.

That is [freenet/ghostkeys#21](https://github.com/freenet/ghostkeys/issues/21) exactly: ghostkeys re-keyed twice in one day, every integration broke, and the first anyone knew was a confused user reporting their key worked in the vault but nowhere else.

ghostkeys' own README tells integrators not to do this.

## The fix, when the time comes

ghostkeys now publishes a **pointer record**: a signed record at a fixed address naming the delegate's current code hash. Resolve it instead of vendoring the WASM.

- app id: `ghostkeys.ghostkey-delegate`
- pointer key (fixed, GET this): `HYBCmCSGKjr4jUUoTzivZ48UynjLscwA3nvnpKQAwi3a`
[...]

PR #43, opened 2026-09-06
fix(ghostkey): vendor the current ghostkeys delegate, six generations on

sanity:
Fixes the "Connect a ghostkey" hang. Stopgap for #5.

## Symptom

Clicking **Connect a ghostkey** in My Store sits on *"Waiting for vault…"* forever. No error, nothing in the browser console — the request simply never comes back.

## Cause

A delegate's identity is `BLAKE3(BLAKE3(wasm) ‖ params)`. Harvest and the ghostkeys vault **both register with empty params**, so the WASM alone decides which delegate instance — and therefore which **secret store** — you land in.

Harvest vendored a copy from 2026-09-02 that ghostkeys' own `legacy_delegates.toml` lists as superseded, with **six entries after it**.

The failure is not an error, which is why it presents as a hang. From the node log, at the exact second of the click:

```
Failed executing delegate delegate_key=CmDCFB19xtkUYiZeyz5Hv9uT7FNCcva6P1u3J3V8mmZ1
cause deserialize request: CBOR deserialize: Semantic(None,
  "unknown variant `RequestAnyAccess`, expected one of `ImportGhostKey`,
   `ListGhostKeys`, `GetGhostKey`, `GetCertificate`, ... `TestPermissionPrompt`")
```

The UI speaks `ghostkey-common` 0.3 and sends `RequestAnyAccess`. That variant postdates the vendored delegate, so it fails to deserialize and **never replies at all**. Nothing times out, so the button spins indefinitely.

## Verified it lands on the right store, rather than assumed

Mapped every generation in ghostkeys' registry to its delegate key (`BLAKE3(code_hash ‖ ∅)`, derivation confirmed against `freenet-stdlib-0.8.5/src/delegate_interface.rs::generate_id`) and cross-checked against the secret stores on the framework node:

| | code hash | delegate key | secret store |
|---|---|---|---|
| vault / current | `0d0e2043…` | `9g8MpAVzuPP9ZYa1Rr3UxxEaBC2cLB9YSKsQGmhBpq1U` | **20 files — the user's keys** |
| Harvest shipped | `6c42cdb7…` | `CmDCFB19xtkUYiZeyz5Hv9uT7FNCcva6P1u3J3V8mmZ1` | **no directory at all** |

The derivation was sanity-checked by reproducing the observed key: `BLAKE3(6c42cdb7…)` → `CmDCFB19…`, exactly what the node's error names.

The vendored file now hashes to `0d0e2043…` — what ghostkeys' **pointer record** names as current — and its delegate key is `9g8MpAVz…`, byte-identical to the store holding the keys.

Reproduced from source with ghostkeys' canonical `scripts/build-delegate.sh`, the build that remaps `$CARGO_HOME` so the hash is a fact about the source rather than about the machine. Confirmed the new delegate does know `RequestAnyAccess`, alongside `SignMessage`, `GetCertificate`, `ListGhostKeys` and `GrantPermission`.

## Also: correcting the reasoning that let this rot

`harvest-addresses.rs` excluded this artifact because *"nothing in this workspace can move it."* That is true, and it is the wrong question to stop at.

The risk for a vendored artifact is not that **we** move it — it is that **upstream** moves and we do not follow. A relative HEAD-vs-base comparison structurally cannot see that, because both sides are equally stale. Catching it needs an **absolute** check against upstream, which is exactly what ghostkeys' pointer record (`HYBCmCSGKjr4jUUoTzivZ48UynjLscwA3nvnpKQAwi3a`) exists for and is readable over the node like any other contract.

Same shape as the Bitcoin addressing constants fixed in #32 this morning, and the same durable answer.

## Known gaps, deliberately not in this PR

- **The UI has no timeout on this request.** Once the delegate is correct the symptom goes away, but any future protocol drift reproduces the identical silent spin. A delegate that cannot parse a request does not reply, and nothing on the UI side notices. Worth its own fix so the failure is visible instead of indefinite.
- **Nothing still watches this file for upstream drift.** This PR makes it current; it does not stop it going stale again. That is #5.

[AI-assisted - Claude]


sanity:
## Review record

**Tier: Full** — this changes which delegate instance Harvest talks to, and therefore which secret store users land in.

| lens | outcome |
|---|---|
| Independent blind review (separate session, read-only) | **MERGE**, no blocking findings |

### The load-bearing claim was independently reproduced, not trusted

The reviewer **rebuilt `ghostkey_delegate.wasm` from ghostkeys' `origin/main` via its own `scripts/build-delegate.sh` and got a byte-identical binary** to the one committed here. That is the strongest form of the check: the vendored bytes are reproducible from upstream source, not merely copied from a directory on this machine.

Confirmed independently:

- **Derivation** — `generate_id` in freenet-stdlib 0.8.5 is `BLAKE3(code_hash ‖ params)`; Harvest's `DELEGATE_PARAMETERS = &[]` and ghostkeys' `Parameters::from(Vec::<u8>::new())` are both empty, so params match and the WASM alone decides the instance.
- **Currency** — `0d0e2043…` is the `pointer-records.toml` current record. It also appears in `legacy_delegates.toml`, which is exactly the trap that caught us in freenet-bitcoin; the reviewer checked git history and found that entry was added by the routine record-at-publish step with **no later commit superseding it**. Resolves to *not stale*.
- **Ground truth on framework** — `secrets/9g8MpAVzuPP9ZYa1Rr3UxxEaBC2cLB9YSKsQGmhBpq1U/` is populated (dating to Aug 12); `secrets/CmDCFB19xtkUYiZeyz5Hv9uT7FNCcva6P1u3J3V8mmZ1/` **does not exist at all**. Unambiguous: the new key holds the real keys, the old one never held anything.
- **Variant coverage** — `RequestAnyAccess`, `GetCertificate` and `SignMessage` all exist in the vendored generation, so the mismatch does not recur in the other direction.

One immaterial discrepancy: I counted 20 files in that store, the reviewer 13 — different traversal of `.snapshots`. Nothing turns on it.

### Confirmed gap, deliberately out of scope

Neither `contract-drift.yml` nor `ci.yml`'s `wasm-staleness` covers `ghostkey_delegate.wasm`; both only walk `reputation_contract`, `store_contract`, `mailbox_contract` and `harvest_delegate`. Pre-existing, and named in the PR and in the code comment this PR rewrites. **#5 stays open** for the pointer-resolution fix that actually closes it.

### A core-level finding surfaced by this bug, filed separately

Chasing why the symptom was a *hang* rather than an error: `crates/core/src/contract.rs:641` logs the delegate execution failure and returns the accumulated messages, and the caller then sends that back as `DelegateResponse`. So the client receives an **empty successful response**, indistinguishable from "the delegate had nothing to say".

I initially read that as "the node sends nothing" and corrected it after reading the caller — the distinction matters for whoever fixes it. Not a Harvest bug, and it affects every delegate-using app.

[AI-assisted - Claude]

PR #45, opened 2026-09-06
fix(ui): install a wasm panic hook so panics say what they were

sanity:
## Problem

A Rust panic compiled to wasm surfaces in the browser as:

```
Uncaught RuntimeError: unreachable
    at harvest-ui_bg-....wasm:0x2d7af8
    at harvest-ui_bg-....wasm:0x2c3091
    ...
```

No message, no file, no line — just raw wasm offsets. The panic's own text is discarded before anything can print it.

Observed today: the first ghostkey a user ever connected panicked the UI immediately after the vault answered, and the console showed **four identical `unreachable` traces and nothing else**. The panic is in the identity-migration path, which is wasm-only and had therefore never executed anywhere before — exactly the code most likely to panic and least able to explain itself.

## Approach

`std::panic::set_hook` routed through `tracing::error!`.

- `dioxus::logger::initialize_default()` sets up **tracing, not a panic hook**, so this is not redundant with it.
- Routed through `tracing::error!` rather than adding `console_error_panic_hook`: no new dependency, and panics land in the same log stream as everything else the app reports.
- `#[cfg(target_arch = "wasm32")]` — native builds already print panics properly.

## What this does and does not do

It does **not** fix the panic. It makes that panic, and the next one, diagnosable — including on the very next load, which is how I intend to find the current one.

Verified it compiles for `wasm32-unknown-unknown`.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/main.rs
+++ b/ui/src/main.rs
@@ -13,5 +13,31 @@ mod store_link;
 
 fn main() {
     dioxus::logger::initialize_default();
+
+    // Make wasm panics say what they were.
+    //
+    // Without a hook, a Rust panic compiled to wasm surfaces in the browser as
+    // a bare `Uncaught RuntimeError: unreachable` with a stack of raw wasm
+    // offsets and NO message -- the panic's own text, file and line are
+    // discarded before anything can print them. That is not a small
+    // inconvenience: it is the difference between a bug report that names its
+    // cause and one that can only be guessed at.
+    //
+    // Observed 2026-09-06: the first ghostkey a user connected panicked the UI
+    // immediately after the vault answered, and the console showed four
+    // identical `unreachable` traces and nothing else. The migration path it
+    // panicked in had never executed before -- it is wasm-only, so no test had
+    // ever run it -- which is exactly the code most likely to panic and least
+    // likely to explain itself.
+    //
+    // `initialize_default()` above sets up tracing, not a panic hook, so this
+    // is not redundant with it. Routed through `tracing::error!` rather than
+    // `console_error_panic_hook` so it needs no new dependency and lands in the
+    // same log stream as everything else the app reports.
+    #[cfg(target_arch = "wasm32")]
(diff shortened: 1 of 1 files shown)


sanity:
## Review record

**Tier: Light** — additive, wasm-only, diagnostic. Independent blind review, read-only.

**VERDICT: MERGE**, no blocking findings. Verified against dependency sources rather than assumed:

- **Does the hook run under `panic = "abort"`?** Yes — the load-bearing question, and it checks out. `rust_panic_with_hook` always calls the hook *before* branching on panic strategy; abort vs unwind only changes what happens after it returns. Same on wasm32-unknown-unknown as anywhere.
- **Is `tracing::error!` safe in the hook?** `dioxus-logger` re-exports `tracing`, and `initialize_default()` (the line above) installs a `tracing_wasm::WASMLayer` global subscriber, so a working console-backed subscriber exists by the time the hook is registered. Tracing macros no-op rather than panic without a subscriber, so a double-panic-in-hook is unlikely.
- **Placement.** `set_hook` is a process-global registration, so it covers `launch`, initial render, and every later event callback.

**Worth knowing, not blocking:** `dioxus-web`'s `devtools` module calls `std::panic::set_hook` itself during `launch` under `#[cfg(debug_assertions)]`, which would override this one — so the fix **cannot be observed via a local `dx serve` debug build**. It is compiled out in release, which is what gets published, so production behaviour is as intended. Recording it so nobody concludes the hook is broken after testing it the one way that cannot show it.

**Known limitation:** routing through `tracing::error!` gets the Rust panic message plus file:line, but not the JS-level stack that `console_error_panic_hook` captures. Accepted — a large improvement over a bare `unreachable`, at no dependency cost.

External non-Claude (`codex review`) unavailable — usage limit, resets 2026-09-07 08:14; Claude lens substituted per the review rule.

[AI-assisted - Claude]

PR #46, opened 2026-09-06
fix(ghostkey): stop the vault response panicking, and never strand the button

sanity:
Fixes the "Waiting for vault…" hang that remained after #43. The vault answers and the key arrives; the UI then panics and never updates.

### 1. Re-entrancy — the handler runs inside a write borrow

`gateway::response_handler` invokes this as:

```rust
apply_delegate_response(&mut APP_STATE.write(), response)
```

The temporary write guard lives for the **whole statement**, so the entire `GhostKeyList` arm runs while APP_STATE is mutably borrowed. Everything it reached then took a second borrow and panicked:

```
PANIC: panicked at ui/src/gateway/migrate_ops.rs:258:10:
called `Result::unwrap()` on an `Err` value: AlreadyBorrowedMut(AlreadyBorrowedMutError)
```

Three sites — `migrate_ops.rs:258` (`start_reputation_migration`), `migrate_ops.rs:421` (`send_harvest_request`), `store_ops.rs:530` (`list_stores`).

Under `panic = "abort"` each trap kills the rest of that call. That also explains the *other* symptom in the log — "the delegate did not answer about marker …" — the handler that would have processed the answer was already dead.

**Spawning alone was not enough, and `list_stores` proves it:** it was *already* spawned and still panicked, because `spawn_local` polls the future **eagerly up to its first await**, and `list_stores` reads APP_STATE before awaiting. The follow-up task now yields (`TimeoutFuture::new(0).await`) before touching anything. That await is load-bearing, not politeness.

### 2. Ordering — the flag was cleared last

`request_any_access_in_flight = false` was the **final** statement of the arm, so any failure above it stranded the flag. The user is then told the vault has not answered, for a request that demonstrably completed.

Now cleared **first**. The vault answering is the whole meaning of that flag; everything after is follow-up work that must not be able to retract it.

## Why this survived until now

The entire path is **wasm-only and had never executed anywhere**. It needs a real ghostkey to reach, so no test, no CI job and no local run had ever entered it. It broke the first time a user connected a key — which is also why the panics were opaque until #45 added a panic hook.

## Testing

344 tests pass, clippy clean on `wasm32-unknown-unknown`.

The new test is a **source scrape**, deliberately. The invariant is an *ordering*, and reproducing the failure needs a held write guard plus a wasm-only call chain — neither exists on the host, so a behavioural test would pass just as happily under the bug.

Verified by reintroducing each fault rather than only watching it pass:

| mutation | result |
|---|---|
| flag clear moved back to the end | **FAILS**, naming both offsets |
| the yield removed | **FAILS** — "must yield before touching APP_STATE" |
| restored | passes |

The test's first draft failed for the **wrong reason** — it matched `list_stores` in the explanatory comment rather than in the code — so the scan is now scoped to the text after the spawn. Recorded because a pin test that fails for the wrong reason is worth no more than one that cannot fail at all.

## Not addressed here

Other handlers (`on_delegate_response`, `on_bitcoin_delegate_response`) run under the same write guard and could reach the same `APP_STATE.read()` re-entrancy. Nothing observed yet, but the hazard is structural rather than specific to this arm; worth an audit of everything reachable from `apply_delegate_response`.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/state.rs
+++ b/ui/src/state.rs
@@ -4095,59 +4095,17 @@ impl AppState {
             ghostkey_common::GhostkeyResponse::GhostKeyList { keys } => {
                 info!("Received {} ghostkeys", keys.len());
 
-                // Learning which identities we can act for is the first
-                // moment we can ask the delegate what stores each of them
-                // owns. We can't do this at startup: `ListStores` is
-                // scoped to a fingerprint, and no fingerprint is known
-                // until the vault shares one -- Harvest deliberately does
-                // not call `ListGhostKeys` any more (see the comment in
-                // `components::App`). Without this, `my_stores` stays
-                // empty after every page reload and the seller is offered
-                // "Create Store" for a store they already have.
-                #[cfg(target_arch = "wasm32")]
-                for fingerprint in keys.iter().map(|k| k.fingerprint.clone()) {
-                    wasm_bindgen_futures::spawn_local(async move {
-                        if let Err(e) =
-                            crate::gateway::store_ops::list_stores(fingerprint.clone()).await
-                        {
-                            dioxus::logger::tracing::error!(
-                                "Failed to list stores for {fingerprint}: {e}"
-                            );
-                        }
-                    });
-                }
-
-                // Learning the verifying key is also the first moment the
-                // store and mailbox contracts' addresses can be derived --
-                // for the current generation and for every superseded one --
-                // so this is where their migration starts. It deliberately
-                // does NOT wait for `ListStores`: the delegate's registry
-                // names the instance a store was CREATED at, and that
-                // registry is itself lost whenever the delegate re-keys.
-                // Deriving from the ghostkey needs neither.
+                // CLEARED FIRST, and deliberately not last.
                 //
-                // Started unconditionally, once per (instance, current code
-                // hash). Nothing here asks whether the current instance is
-                // empty -- see `crate::migrate`'s module docs for why that
-                // gate is the shape that silently disables a migration.
-                #[cfg(target_arch = "wasm32")]
-                for key in &keys {
-                    if let Some(vk) = key.verifying_key_bytes.as_ref() {
-                        crate::gateway::migrate_ops::start_identity_migration(&key.fingerprint, vk);
-                    }
-                }
+                // The vault has answered; that is the whole meaning of this
+                // flag. Everything below is follow-up work, and while the flag
+                // sat at the END of this arm a failure in any of it left the
+                // button reading "Waiting for vault..." forever, for a request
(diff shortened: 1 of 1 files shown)


sanity:
## Review record

**Tier: Full** — state-handling in the response path, and the failure mode is a permanently bricked `APP_STATE`.

| lens | round | outcome |
|---|---|---|
| Independent blind review | 1 | **DO NOT MERGE** — stale comment asserting a mechanism this PR disproves; pin test matching a comment not code |
| Independent blind review | 2 | **MERGE** — all findings confirmed fixed, all three handlers verified clean |
| **External non-Claude (`codex review`)** | — | **UNAVAILABLE**, usage limit resets 2026-09-07 08:14; Claude lens substituted per the review rule |

### What round 1 found, and what chasing it turned up

The stale comment was real. Fixing it exposed two worse things:

1. **The commit meant to replace that comment had prepended it to the top of `state.rs`**, above the module doc, leaving the stale block in place. Cause: the old text was sliced with an unanchored `index()` for the end marker, which matched an earlier occurrence, so the slice was empty and `str.replace("")` inserted at position 0. Both anchors now search from `start`, and the slice is asserted non-empty and asserted to contain the text being replaced. **It compiled and passed CI throughout** — a `//` comment above `//!` is legal Rust — so nothing mechanical was going to catch it.

2. **The same bug was still live in `on_delegate_response`**, which called `self.start_reputation_migration(..)` synchronously at two sites under the same write guard. Fixing only the ghostkey arm would have left the identical panic reachable as soon as an RSA key arrived. Fixed *at the method*, so it is now safe from any handler.

### Confirmed by round 2

- **Structure clean** — no stray block above the module doc, one copy of the explanatory block, no surviving claim that `spawn_local` polls eagerly.
- **`start_reputation_migration` safe from any handler** — the `self.ghostkeys` lookup stays under the held borrow (correct, no new borrow); only the call needing its own borrow is deferred. Both `on_delegate_response` sites go through the method, not the raw function.
- **All three handlers audited** — every `migrate_ops::` / `store_ops::` / `ensure_encryption_key(` call site is either inside a `spawn_local` block or behind that wrapper. `on_bitcoin_delegate_response` has no such synchronous call at all.
- **Pin test anchored on code** in every assertion, with exactly one real occurrence of each literal in the scanned range, and verified non-vacuous by deleting the deferred call and watching it fail.

344 tests, clippy clean on wasm32.

### Note taken

`published-contract/last-version` is modified in the worktree from an earlier publish and deliberately not in this PR. It will be committed with the next publish, not swept into an unrelated commit.

[AI-assisted - Claude]

PR #48, opened 2026-09-06
fix(ui): render a ghostkey's notary attestation instead of printing its JSON

sanity:
## Problem

My Store showed:

```
First Ghostkey({"action":"freenet-donation","amount":1,"delegate-key-created":"2024-08-13 15:45:36"})
```

`GhostKeyInfo::notary_info` is the raw certificate field and was printed verbatim at `my_store.rs:266`. The surrounding CSS class is `identity-tier`, so a tier was always the intent — the payload was just never unpacked.

Now reads:

```
First Ghostkey   donated, 13 August 2024
```

## Why unpack it rather than hide it

Harvest's own reputation model says *a clean record with an old, high-tier ghostkey is the best possible reputation*. The amount and the date are exactly what a buyer weighs, so this is **seller credibility, not decoration** — the right fix is to render it, not to drop it.

## Three shapes, all handled

Matching how the ghostkeys vault presents the same field (`ui/src/components/ghostkey_list.rs`), so the two apps never disagree about one key:

| stored | rendered |
|---|---|
| `{"action":"freenet-donation","amount":1,…}` | `$1 donated, 13 August 2024` |
| `donation_amount:20` (legacy) | `$20 donated` |
| `Freenet Notary` (migration adapters) | `Freenet Notary` — verbatim |

**Anything unrecognised passes through unchanged rather than being guessed at.** Showing a raw field is ugly; inventing a tier for a certificate this build cannot read would misrepresent a seller's standing, which is worse.

## On the JSON reader

Hand-rolled rather than `serde_json`, deliberately: the value is written by another project, the failure that matters is an *unexpected shape*, and every caller already treats "could not read it" as a normal outcome. It cannot panic on a malformed certificate — pinned by a test feeding it six malformed inputs and asserting none render as a donation.

## Testing

Six tests: the exact string observed in the browser, a larger donation, the legacy encoding, the migration-adapter string, a date without an amount, and the malformed set. The observed string is pinned rather than a synthetic one, since the whole bug was that nobody had looked at what the field contains.

350 tests pass, clippy clean on `wasm32-unknown-unknown`.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/my_store.rs
+++ b/ui/src/components/my_store.rs
@@ -762,23 +762,63 @@ fn extract_created_date(info: &str) -> Option<String> {
 /// Not `serde_json`: the value is a certificate field written by another
 /// project, so the failure that matters is an unexpected shape, and every
 /// caller here already treats "could not read it" as a normal outcome rather
-/// than an error. A parser that returns `None` on anything surprising is the
+/// than an error. A reader that returns `None` on anything surprising is the
 /// right shape for that, and it cannot panic on a malformed certificate.
+///
+/// Matches the key only at the TOP LEVEL of the object. An earlier version
+/// took the first textual match, so `{"nested":{"amount":42},"amount":7}`
+/// answered 42 -- the wrong donation, reported confidently. The field is flat
+/// in practice and notary-issued rather than seller-chosen, so that input is
+/// not expected; it is refused anyway, because a wrong answer here misstates a
+/// seller's standing while `None` merely declines to.
 fn extract_json_field<'a>(info: &'a str, key: &str) -> Option<&'a str> {
+    let bytes = info.as_bytes();
     let pat = format!("\"{key}\"");
-    let pos = info.find(&pat)?;
-    let after = info[pos + pat.len()..].trim_start();
-    let after = after.strip_prefix(':')?.trim_start();
-    if let Some(rest) = after.strip_prefix('"') {
-        let end = rest.find('"')?;
-        Some(&rest[..end])
-    } else {
-        let end = after.find(|c: char| !c.is_ascii_digit())?;
-        if end == 0 {
-            return None;
+    let mut depth: i32 = 0;
+    let mut in_string = false;
+    let mut escaped = false;
+    let mut i = 0usize;
+    while i < bytes.len() {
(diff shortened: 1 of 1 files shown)


sanity:
## Review record

**Tier: Light** — display-only, plus pure functions and tests. Independent blind review, read-only.

**VERDICT: MERGE**, no blocking findings.

### Verified independently

- **Cross-app agreement.** Checked the notary's own `generate_notary_keys.sh`: real certificates always use `{"action":"freenet-donation","amount":N,"delegate-key-created":"YYYY-MM-DD HH:MM:SS"}`. For that shape Harvest and the vault report the **same amount and date**, styled differently. The two theoretical divergences (fallback wording for an unrecognised string, invalid-day handling) have no production trigger.
- **Parser robustness.** Probed with nested objects, arrays, escaped quotes inside string values, key-prefix collisions, whitespace variants, negative and huge numbers, `null`, and truncated JSON — every case returns `None` or the correct value. No panics, and no wrong-but-confident answers.
- **Scope.** One display string, new pure functions, tests. No behavioural or protocol change.

### The finding I caught by probing my own code first

Before opening this for review I tried to break the reader and found the one case that returned a **wrong** value rather than none:

```
{"nested":{"amount":42},"amount":7}   ->  42
```

It took the first textual match, so a nested object shadowed the real field and the UI would have reported the wrong donation, confidently. Fixed in `4ca6c63` with a depth- and string-aware scan that matches only at the top level.

The field is flat in practice and notary-issued rather than seller-chosen, so that input is not expected. Refused anyway, because the failure modes are not symmetric: `None` declines to describe a certificate, while a wrong number **misstates a seller's standing** — and standing is the thing this display exists to convey.

Three cases pinned: nested shadowing, amount existing only nested, and the key appearing inside a string value.

### Correction

An earlier version of this commit message claimed "359 tests pass". The actual figure is **353**; amended. Same habit this repository has been correcting all day — a number carries more apparent authority than the prose around it, so it deserves more checking, not less.

External non-Claude (`codex review`) unavailable — usage limit, resets 2026-09-07 08:14; Claude lens substituted per the review rule.

[AI-assisted - Claude]

PR #50, opened 2026-09-06
fix(ghostkey): restore an already-shared identity instead of re-prompting

sanity:
## Problem

Connecting a ghostkey had to be repeated on **every page load**.

The grant was never the problem — the app simply never asked whether it had one.

## Cause

Startup deliberately skipped `ListGhostKeys`, reasoning that the vault auto-grants only the importing webapp, so for any other webapp the list is empty until the user approves a `RequestAnyAccess` prompt.

**The premise is true. The conclusion — never call it — does not follow**, and it is what forced the re-approval.

Approving `RequestAnyAccess` **persists** a grant. The delegate calls `permissions::grant_third_party`, saving `{ReadPublic, Sign}` for this requestor against that fingerprint, and `handle_list` returns exactly the keys the requestor holds `ReadPublic` on:

```rust
// Only show ghostkeys the requestor has at least ReadPublic on.
if !permissions::has_scope(ctx, fp, requestor, GhostkeyScope::ReadPublic) {
    continue;
}
```

The grant is keyed by `SignatureRequestor::WebApp(ContractInstanceId)`, and this app's contract id is fixed, so it survives a reload.

So the list is empty **only before the first approval** — exactly when an empty answer is right and the My Store empty state offers "Connect a ghostkey". Afterwards it returns the shared key with no prompt.

`RequestAnyAccess` raises a prompt every time by construction. It is the wrong call when you only want to know what you already have, and it was the only call being made.

## The way this could have backfired, checked first

If `ListGhostKeys` could itself prompt, calling it at startup would mean a dialog on every load — worse than the bug being fixed. It cannot:

- `required_scope` returns `None` for it (*"filtered per-key inside the handler"*),
- it is not a disclosure request (`discloses_private_key` = false, pinned by a delegate test),
- it dispatches straight to `handle_list`, which filters and returns, with no prompt path.

A caller with no grant gets an empty list, not a dialog.

## Failure behaviour

Deliberately not user-facing. An empty or failed list leaves the same empty state offering "Connect a ghostkey" — where the user would have started anyway.

## Testing

353 tests pass, clippy clean on `wasm32-unknown-unknown`.

The behaviour that matters cannot be unit-tested here (it needs a real vault holding a real grant), so it will be verified on the live node after publish: load, approve once, reload, and confirm the second load shows the identity **without** a prompt.

[AI-assisted - Claude]

### DIFF
--- a/ui/src/components/app.rs
+++ b/ui/src/components/app.rs
@@ -107,20 +107,60 @@ pub fn App() -> Element {
                     }
                 }
 
-                // Step 3: Register the ghostkey delegate. We deliberately
-                // do NOT call `ListGhostKeys` on startup any more: the
-                // vault now auto-grants only the importing webapp (the
-                // Ghostkey Vault itself) on key import, so for any other
-                // webapp `ListGhostKeys` returns an empty list until the
-                // user explicitly approves a `RequestAnyAccess` prompt.
-                // The "Connect a ghostkey" button on the My Store empty
-                // state triggers that prompt; the response folds the
-                // shared key into APP_STATE.ghostkeys.
+                // Step 3: Register the ghostkey delegate, then ASK IT WHAT WE
+                // ALREADY HAVE.
+                //
+                // An earlier version deliberately skipped `ListGhostKeys` here,
+                // reasoning that the vault auto-grants only the importing
+                // webapp, so for any other webapp the list is empty until the
+                // user approves a `RequestAnyAccess` prompt. The premise is
+                // true. The conclusion does not follow, and skipping the call
+                // is what made the user re-approve on EVERY page load.
+                //
+                // Approving `RequestAnyAccess` PERSISTS a grant: the delegate
+                // calls `permissions::grant_third_party`, which saves
+                // `{ReadPublic, Sign}` for this requestor against that
+                // fingerprint. `handle_list` then returns exactly the keys the
+                // requestor holds `ReadPublic` on. The grant is keyed by
+                // `SignatureRequestor::WebApp(ContractInstanceId)`, and this
+                // app's contract id is fixed, so it survives a reload.
+                //
+                // So the list is empty only BEFORE the first approval -- when
+                // an empty answer is exactly right and the My Store empty state
+                // offers "Connect a ghostkey". Afterwards it returns the shared
+                // key with no prompt at all.
+                //
+                // `RequestAnyAccess` raises a user prompt every single time by
+                // construction; it is the wrong thing to call when you only
+                // want to know what you already have. Ask first, prompt only if
+                // the answer is empty.
                 let gk_wasm = include_bytes!("../../public/contracts/ghostkey_delegate.wasm");
                 match crate::gateway::register_delegate(gk_wasm).await {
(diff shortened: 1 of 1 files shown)


sanity:
## Review record

**Tier: Full** — permission-adjacent surface, so two independent lenses rather than one.

| lens | focus | outcome |
|---|---|---|
| Independent blind review | mechanism and safety | **MERGE** |
| Independent blind review | consent and authorization | **MERGE**, one real gap surfaced |
| **External non-Claude (`codex review`)** | — | **UNAVAILABLE**, usage limit resets 2026-09-07 08:14; second Claude lens substituted per the review rule |

### Lens 1 — mechanism, verified against source

- **`ListGhostKeys` cannot prompt.** `RequestAnyAccess` is special-cased at `lib.rs:210` and this does not match it; `acts_on_a_key_it_does_not_name` matches only `SignWithDefault`; `required_scope` → `None`; `discloses_private_key` excludes it. It falls straight through to `handle_list`, which is a pure read with a per-key filter and no prompt call.
- **The grant persists.** `grant_third_party` → `grant_scopes` → `save` → `ctx.set_secret(perm_key(fingerprint), …)`, with the requestor derived from the runtime-attested `MessageOrigin::WebApp(contract_id)` — a fixed contract instance id, nothing session-scoped.
- **Multi-key restore already works.** `on_ghostkey_response` dedup-merges by fingerprint rather than assuming a single-element list — worth knowing, since the old `RequestAnyAccess` path only ever produced one.

### Lens 2 — consent, and the four ways a key could leak

All four closed by the existing per-fingerprint, per-requestor design:

| scenario | result |
|---|---|
| key granted to a *different* webapp | excluded — grants keyed by contract instance id |
| key imported *after* the grant | only the importer gets `grant_full`; no auto-grant to prior third parties |
| key deleted and re-imported on the same fingerprint | `handle_delete` calls `remove_all(fp)`, wiping every grant; re-import starts clean |
| blanket vs per-key | per-fingerprint — the prompt is a picker, only the chosen key is granted |

**On whether this weakens consent:** it does not. The consent moment is the dialog, and `grant_third_party` is explicitly a persistent grant with no TTL and no session scoping — the delegate's pre-existing model, not something this PR invents. The repeated prompt was not a second authorization layer; it was `ListGhostKeys` never being called, so the app treated *already granted* identically to *never granted*. Re-prompting for something already authorized trains people to click through, which makes consent weaker rather than stronger.

### The gap this surfaces, verified and filed separately

**There is no user-facing way to revoke one app's access.** `RevokePermission` exists as a protocol variant and a delegate handler, and **nothing in the vault UI calls it** — confirmed by grep; the only `revoke` hits there are `revoke_object_url` for blob URLs. The sole lever today is deleting the ghostkey, which wipes every app's grant on it: blunt and destructive.

This **pre-dates this PR** and is not worsened in capability. It is worsened in *visibility*: the recurring prompt was the one thing that periodically reminded a user Harvest still had access. Filing it rather than leaving it implicit.

Harvest does retain a disclosure surface — "Your Identities" lists exactly which keys it holds — but that is not a revoke.

[AI-assisted - Claude]

Issue #26, opened 2026-09-06
contract: ListingsV1::verify does not require sortedness, so two peers can converge differently


sanity:
Found while making `ListingId` content-derived in #24. Pre-existing, recorded in `docs/untested-invariants.md`, untouched there deliberately.

`ListingsV1::verify` does not require the listing set to be sorted. A peer that deserializes an unsorted state from the network therefore merges to different bytes than a peer that reached the same *set* through deltas.

Two peers holding the same listings can hold different serialized state, which breaks the merge law the contract relies on — the same class as the mailbox convergence defects fixed in #23, reached by a different route.

The fix is presumably to require sortedness in `verify` — but note the constraint that governs every such change in this repo: **`verify` rejecting existing state makes a contract permanently invalid, with no way back.** Bound by normalising in `apply_delta` rather than by rejecting in `verify`, or establish that no published state is unsorted before tightening.

[AI-assisted - Claude]

Issue #22, opened 2026-09-06
reputation: feedback is identified by a nonce the signature does not cover


sanity:
Found while sweeping for a class of defect fixed on `feat/messaging`. **Not fixed there** — it is a different contract with its own re-key, and the right repair is larger than a last-minute addition.

## The defect

`ReputationStateV1` identifies a feedback entry by `token.nonce`, but the RSA signature covers only `entry.token`. So `category`, `comment` and `submitted_at` ride alongside the signed part **unsigned**.

Anyone reading published feedback can resubmit the same token with different words. Each peer keeps whichever variant it saw first, so peers do not converge — which the contract's own documentation says is impossible.

## Why it is worse than non-convergence

A seller who reads negative feedback can push a neutered variant to peers that do not yet hold the original. Those peers then **refuse the real one**, because its nonce is already used. The genuine complaint never reaches them.

Demonstrated by execution on `feat/messaging` as `known_gap_two_feedback_variants_sharing_a_token_do_not_converge`.

## The repair

Sign the whole entry, so the variant cannot exist at all — rather than making identity content-derived, which is the fix used for the mailbox. For feedback the stronger property is available because entries are signed, and it removes the class instead of resolving collisions.

This re-keys the reputation contract and wants its own review round.

## Context

This is the fifth instance of one shape found today: **a site deciding identity by a field that is not the identity**. The other four were in the mailbox path (`verify`, `summarize`, the state-merge arm, and the migration fold) and are fixed, with a source scrape now enforcing the rule for that contract. Each was written when the nonce genuinely was the identity and was correct on the day; the identity moved in one file and nothing else was compelled to notice.

[AI-assisted - Claude]

PR #54, opened 2026-09-09
fix(contracts): encode set-valued state and summaries canonically

sanity:
## Problem

`ReputationStateV1.used_nonces` (contract **state**), `ReputationSummary` and `MailboxSummaryV2` were `HashSet`s. Their CBOR encoding was therefore not a function of their contents.

This is sharper than "two peers disagree". A `HashSet` draws a fresh `RandomState` seed per **instance**, so two sets holding identical members, inserted in the same order, in the same process, encode to different bytes. Measured:

```
HashSet, same process, different insertion order: bytes equal = false
HashSet, same process, same insertion order:      bytes equal = false
BTreeSet, different insertion order:              bytes equal = true
```

The clearest statement of the defect is inside the reputation contract itself. `apply_delta` sorts `feedback` with the comment *"Sort deterministically by nonce for CRDT convergence"* — and the set immediately beside it in the same struct re-randomised its own order on every encode, spending that sort.

## What this does and does not break

**It is not a correctness failure on the network today, and it would be overstating it to say otherwise.** freenet-core carries two mitigations, each built after a real incident, and a permutation of the same bytes is precisely the case they tolerate:

- `byte_multiset_eq` (`crates/core/src/contract/executor/runtime.rs:74`) — added after #4295, where plain byte-equality *permanently suppressed propagation*. It compares byte **multisets**, so a reordered set compares equal.
- The anti-entropy path defers to the contract's semantic `get_state_delta` before declaring a peer stale (#4857).

Harvest's own `delta()` implementations are semantic — they test set membership (`old_summary.contains(...)`), not byte equality — so the one condition that would have degraded that protection does not apply here.

**What it actually costs is avoidable overhead.** The byte-identical fast paths never fire, so redundant pushes take extra `update_state` invocations, extra state stores and extra broadcasts; anti-entropy digest mismatches force a `get_state_delta` call instead of a cheap byte comparison. Bounded and rate-limited, not a storm.

## Approach

`BTreeSet` in the three serialized positions. Local scratch sets are **deliberately left as `HashSet`**: the rule is that serialized positions must be deterministic, not that `HashSet` is banned, and collapsing the two would lose the distinction that makes the rule teachable.

Scope was checked rather than assumed. Across the workspace, these three are the only `HashSet`/`HashMap` reachable from a type that crosses the wire — `OrdersV1` already uses `BTreeMap`, `ListingsV1::summarize` returns a `Vec` from a sorted `Vec`, and there is no `HashMap` in `common/src/` at all. Every other hit is a function-local dedup accumulator, UI-only state that derives no `Serialize`, or test code.

## Testing

Four tests, **each verified red under the defect and green with the fix** by reverting only the type and re-running.

Note the shape they are forced into, because the obvious test cannot fail. `HashSet::clone` copies the hasher, so `ReputationStateV1::summarize` — which returns `used_nonces.clone()` — yields identical bytes when called twice on **one** state even under the defect. Both summary tests therefore compare two **independently built** states, which is the only shape that goes red. (The mailbox's `summarize` rebuilds with `collect()` and would have failed either way; keeping one shape across both files stops the weaker form being copied from here.)

**The existing suite was blind to this.** Under the reverted defect, 134 pre-existing tests pass and only the new ones fail. `merging_is_order_independent` asserts byte-identical mailbox *state* and never looks at a summary; reputation — the one contract whose state held the bad set — had no determinism test at all.

The fourth test pins the migration claim the registry makes, rather than leaving it as prose: predecessor bytes in any member order decode and re-encode canonically.

## Re-key

Re-keys **store, reputation and mailbox**. It does **not** re-key the delegate — the changed code is not reachable from it, so its hash is unmoved and it correctly takes no registry row. I had appended one before checking, and the rebuild caught it.

Outgoing hashes were verified against the committed artifacts out of git history (`git show 83521fc:ui/public/contracts/<artifact>.wasm | b3sum --no-names`), the method `legacy/README.md` requires. All four matched a fresh build exactly, so the committed artifacts were not stale.

Needs the `contract-rekey-acknowledged` label.

## One cost worth naming

The rebuilt WASM grew: mailbox 359,612 → 385,557 bytes (+26 KB), reputation 438,922 → 452,814 (+14 KB), store 815,838 → 816,003 (+165 B). `BTreeSet`'s monomorphised code is larger than `HashSet`'s here. Recorded as an observed fact rather than waved away; it does not change the correctness argument, but a reviewer should see it rather than discover it.

## Verification

`cargo test --workspace` 611 passed / 0 failed · `cargo clippy --workspace --all-targets -- -D warnings` clean · `cargo fmt --all --check` clean · `cargo make code-hashes` passes ("No current hash is recorded as superseded").

## Related

Same defect class as #26 (`ListingsV1::verify` does not require sortedness) — state that is not canonically encoded. #26 is not fixed here; it needs the normalise-in-`apply_delta` treatment its own issue describes.

[AI-assisted - Claude]

https://claude.ai/code/session_01Rw6QPkkPyb6aE8jE5LBbmA


sanity:
## Re-key: the moved addresses, for acknowledgement

`contract-drift` will go red on this PR. That is correct — the re-key is intended. **I have deliberately not applied `contract-rekey-acknowledged` myself**, because that label's description is *"A **human** has read the moved contract/delegate addresses and decided the re-key is intended"*, and applying it from an agent would assert something untrue and quietly spend the one guard standing between a re-key and a publish.

So here are the addresses to read. Both sides came from `cargo make code-hashes`, before and after the change.

| artifact | old address | new address |
|---|---|---|
| store_contract | `H6c5AvPmiaj2qjtMeu2Sdppx6Z2VLWZ4jg8idr6qidc4` | `2XdN5cvQsYKcxLeq9LJGbk1C5zqN8sLBACdu97Vk68T4` |
| reputation_contract | `BStztxkmJkKnpTc9tazF4rfdaBM9tygD9McH3Xii9Ycp` | `5qpLQusX2BZf7wMDbwv9uhGUkKKHNVPuz8EMKnAPbzsb` |
| mailbox_contract | `EajRA8HJnMC589wYN44nRzhoc8NKd9NzRdeHri4DvwYn` | `FGq4MsrkSnp9PVwiHcZmDeHpfYarq2ALavxtimCGDjtK` |
| harvest_delegate | `5fCPAtBF2U7WW9antkJBAWEB4ttzDv2iqguoLigv1msk` | **unchanged** |

Two things worth checking rather than taking from me:

**The delegate does not move.** `harvest-common` is compiled into it, so the usual expectation ("a change anywhere in `common` re-keys everything") would say it should. It does not, because the changed code is not reachable from the delegate and is eliminated. I had already appended a `V12` row to `legacy/harvest_delegate.toml` on that expectation before rebuilding; the rebuild contradicted it and the row was removed. `cargo make code-hashes` would have refused it anyway — the guard works — but it is worth recording that the general claim is not automatic.

**The code hash moved for all three contracts and the parameter encoding did not.** Store params are still 56B, so this is a plain re-key and needs no parameter-generation split in `ui/src/migrate.rs`. `PUBLISHED_UNDER_LEGACY_PARAMS` gets `(10, false)` on that basis, confirmed from the tool output rather than inferred from the diff.

If you agree the re-key is intended, applying the label is the remaining step.

[AI-assisted - Claude]

sanity:
## Multi-perspective review — Full tier, all findings addressed

Full tier because this touches contract wire encoding and re-keys three contracts. Five independent reviewers, each blind to the others, on `f372bde`. All findings are fixed in `929abda` except one, which is filed with its reasoning below.

**Lenses that ran:** code-first (reads the code before the description, hunting intent/implementation mismatch) · testing (hunting vacuousness) · skeptical (adversarial) · re-key and migration registry (this change's own failure mode) · **codex** `codex review --base origin/main` (external, non-Claude).

Codex passed clean. It ran the test binaries and the address derivation itself and reported the ordering change preserves the CBOR sequence shape and predecessor decoding, with the migration hashes consistent with the rebuilt artifacts.

### The finding that mattered

**Three of the four Claude lenses independently found that one of my new tests could not fail.** `predecessor_state_in_any_member_order_decodes_and_normalises` encoded a `Vec` and decoded into a `BTreeSet` *named in its own body*, so the type under test was supplied by the test rather than by the code. It was green under every mutation, including the full revert.

That is worse than a wasted test. `legacy/reputation_contract.toml` asserts the V9 fold decodes, and my commit message claimed that assertion was "pinned by a test rather than left as a registry claim". It was not pinned. The reviewer wrote a replacement, ran it, and observed it red under the revert in the same run where the shipped test passed.

Fixed: the test now decodes V9's actual on-the-wire shape into `ReputationStateV1` itself, and is verified red under the reputation revert.

Worth recording plainly: the previous commit carried a long comment about vacuousness traps while shipping one.

### Every finding, and what happened to it

| # | Severity | Finding | Disposition |
|---|---|---|---|
| 1 | High | The predecessor-decode test cannot fail; the registry's migration claim is unpinned | **Fixed.** Decodes through the real type. Verified red under mutation. |
| 2 | Medium | The new tests were inserted between a 26-line "KNOWN GAP" doc block and the `#[test]` it belonged to, moving the record of an unfixed convergence gap onto an encoding test and leaving `known_gap_two_feedback_variants_sharing_a_token_do_not_converge` undocumented | **Fixed.** Tests moved after it. |
| 3 | Low/Medium | The mechanism written into a permanent registry row is wrong for the target. `HashSet` is not seeded randomly on `wasm32-unknown-unknown` | **Fixed**, see below. |
| 4 | Medium (pre-existing) | `tests/rehearsal/src/main.rs`'s `ENCODING_BY_GENERATION` covered V1..V7 while the store lineage was already 9, so its length assertion panicked. Stale since V8 | **Fixed.** Extended to V10. |
| 5 | Medium (pre-existing) | Reputation and store `get_state_delta` cannot decode an empty summary | **Filed as #55**, not bundled. Reasoning below. |
| 6 | Low | "the fold from V9 is pure data transfer and loses nothing" is over-broad; the V9 fold has a pre-existing #22 exclusion path | **Fixed.** Scoped to what this change does. |
| 7 | Low | The `used_nonces` doc comment claimed two `to_cbor` calls on one value differ. They do not; only independently-built instances do | **Fixed.** |
| 8 | Low | 5-nonce fixtures make the reputation guards probabilistic; two small `HashSet`s can agree on an order by luck | **Fixed.** Raised to 32. |
| 9 | Low | The summary test cannot fail independently of the state test, because `summarize` returns a clone so the field type and the alias cannot disagree and still compile | **Fixed** by relabelling it documentary rather than counting it as a guard. It is kept for the clone-vacuousness trap it records. |
| 10 | Nits | Missing rustdoc separator, missing trailing periods on registry rows, an inaccurate "and nothing else" | **Fixed.** |
| 11 | Process | `contract drift` is red and the PR is unmergeable without the label | Correct and intended. See the addresses comment above; the label is deliberately not mine to apply. |

### On finding 3, because it is going into a row the rules say is never deleted

I wrote that a `HashSet` "draws a fresh random seed per instance". That is true on the host where I measured it and **false on the target the contracts actually run on.** Verified in the pinned toolchain's own std sources: `wasm32-unknown-unknown` routes to `sys::random::unsupported`, whose `hashmap_random_keys` returns a stack address and a heap address with the comment that it "isn't particularly secure, but there isn't really an alternative". No entropy at all.

The defect is still real, and the accurate statement is the stronger one. `RandomState::new` bumps a per-thread key on **every** construction, target-independently and deliberately (`keys.set((k0.wrapping_add(1), k1))`), so two peers that have built different numbers of sets encode differently. Their operation history is exactly what differs.

The reason this was worth a fix rather than a shrug: stated the old way, a future reader who checks whether `HashSet` is random on wasm would find that it is not, and could reasonably conclude the change had been unnecessary and revert it. "Measured, not inferred" is also no longer claimed for the target, because the measurement was on the host.

### Why finding 5 is filed rather than bundled

The reviewer made a real case for bundling it, citing `legacy/store_contract.toml`'s V9 row as precedent for folding a fix into a re-key "because the re-key makes it free exactly once". I did not take it, for three reasons: it is a behaviour change rather than an encoding one, it affects two contracts so a proper fix is its own change, and adding it would re-stale a review that has already run on this content. The re-key urgency also does not bite yet, since re-keys stay free until someone clicks Create Store.

### Verification on `929abda`

`cargo test --workspace` all green across 11 test binaries · clippy `-D warnings` clean · `cargo fmt --check` clean · `cargo make code-hashes` reports no current hash recorded as superseded.

**The contract WASM is byte-identical to `f372bde`** — only comments, tests and registry prose changed — so the recorded V10 hashes and the moved addresses posted above still stand, and the re-key acknowledgement does not need redoing.

[AI-assisted - Claude]

sanity:
## Review addendum — correcting my own "all findings addressed" claim

The four lenses' full reports arrived after I posted the consolidated review above, and they contained one finding I had not addressed and one factual conflict with my own measurement. **My claim that every finding was addressed was wrong when I made it.** `298ee6b` makes it true.

### The finding I missed: nothing covered the entry point

Two lenses raised it and I acted on neither. It is the sharpest form of the defect, so missing it mattered.

`summarize_state` **decodes** the state before summarizing it, and all four tests I had added build state in-process, one layer below that. The decode is where this bit hardest: every call through the entry point decodes the same unchanging bytes into a *fresh* collection, and `RandomState::new` bumps a per-thread key on every construction — so **one node, holding one unchanging state, answered two `summarize_state` calls with different summary bytes.** That is the shape behind the freenet-core incident this class of defect is known for, where a contract's anti-entropy could never agree even with itself.

`contracts/reputation-contract` had **no tests at all**, which is why none of this was covered. Two added, both verified red under the V9 revert and green with the fix:

- `summarize_state_is_byte_stable_across_calls_on_one_state`
- `two_peers_holding_the_same_nonces_summarize_identically`

`ed25519-dalek` is a dev-dependency, for the same reason and with the same comment as `contracts/mailbox-contract`: `ReputationParameters::new` is the single construction site on purpose, because the field set is hashed into the address, so a test needs a real key. A dev-dependency is not linked into the `cdylib`, so it does not move the code hash — **confirmed by running `cargo make code-hashes` and seeing `eef8685c…` unchanged, not assumed.** The contract WASM is untouched and the recorded V10 hashes still stand.

### A correction to the review record

One lens's mutation table reported that reverting `used_nonces` alone, leaving `ReputationSummary` a `BTreeSet`, compiles and isolates that single test. **It does not compile.** `summarize` returns `used_nonces.clone()` as a `ReputationSummary`, so the field and the alias cannot disagree; the observed result is `error[E0308]: mismatched types`. I re-measured rather than arguing it, and a second lens reached the same conclusion by reading. The minimal reputation mutation is both types together, which is also the real V9 shape — so the disposition I gave for finding 9 above (relabelling the summary test as documentary rather than counting it as a guard) stands.

### Verification on `298ee6b`

`cargo test --workspace` all green, now 613 tests across 11 binaries · clippy `-D warnings` clean · `cargo fmt --check` clean · `cargo make code-hashes` reports no current hash recorded as superseded, with all four hashes unchanged from `f372bde`.

Still outstanding and still not mine: the `contract-rekey-acknowledged` label.

[AI-assisted - Claude]

sanity:
## Re-key acknowledged

Label applied by me at Ian's instruction, given in conversation on 2026-09-09. Recording that provenance rather than leaving it implied, because the label's description says a human read the moved addresses: **Ian authorised the acknowledgement, and did so without reading the address list itself.** The addresses are in the comment above if anyone wants to check them after the fact.

To restate what moved, so the acknowledgement has something attached to it: store, reputation and mailbox all re-key; the delegate does not. No user data sits at any old address, because no store has ever been created.

[AI-assisted - Claude]

PR #56, opened 2026-09-09
docs: record the settled purchase process, on-chain rather than Lightning

sanity:
## Problem

**The design of record contradicted the design.** `docs/design/incentive-mechanism.md` was written around Lightning in Parts 4, 5 and 6, and Part 6 said outright that an on-chain Bitcoin payment *"produces no such secret, so a seller taking on-chain payment cannot produce usable confessions and should be visibly marked as unprotected."*

That is the reverse of the decision taken on 8/9 September, and the reverse of what the code already does: SPV-checked payment proofs, per-order addresses and a bridge-signed `Paid` transition are built and live.

**The corrections existed, and that was the problem.** All three lived in `docs/design/README.md`, a different file from the claims they corrected. So anyone reading the design of record in order got the superseded design, and the file that would have told them otherwise is one they had no reason to open.

## Approach

Fold the corrections into the body, at the steps they affect, and make the README say where they went rather than repeating them. Four documents, listed below. `docs/design.md` needed nothing: it has zero Lightning mentions and is already marked superseded.

### `docs/design/incentive-mechanism.md`

Parts 5 and 6 rewritten for on-chain. Part 4 needed nothing, being rail-agnostic. Added the twelve-character store code, ghostkey-to-complain and why it is not ghostkey-to-buy, the identity-level commitment requirement, and the block-anchor backdating correction with the two rules that neutralise it. Part 6's "why the secret proves payment" becomes why a payment can be proved **without** Lightning, and is explicit that the bridge is trusted for chain state while the proof still stops a bridge misreporting what a real transaction paid. Part 10's question is marked answered. Glossary gains Bridge, Confession and Payment proof; Preimage stays, marked unused, because earlier versions were built on it.

**The section worth reading is new, in Part 5 step 4: why the confession is not part of the public commitment.** This came out of Ian asking the question directly, having expected them to be the same thing — which is the right instinct, since both are seller-signed statements about the same order, made at the same moment, and unconditional.

They cannot be one thing, and the reason is specific to paying on chain, which is why the old document never had to argue it. A complaint needs two halves, and on chain the payment half is **public** — so the admission is the only half that *can* be private. Under Lightning the preimage supplied a private half for free. The apparent repair fails too: the commitment already carries `H(n)`, but `n` is the conversation secret and the conversation encryption keys derive from that same value by a different context, so revealing it to prove you are the buyer also hands over the shipping address.

### `docs/design/transaction-walkthrough.html`

The Lightning aside, including the clipboard and WebLN routes, is replaced by why the rail is on-chain: a Lightning invoice must be **minted** by a node at the moment of sale, which is a network call a delegate cannot make, so Lightning would have required the seller to be reachable whenever anyone places an order. An address derives offline from a key the delegate already holds. That is what keeps delegate-side auto-acceptance reachable at all, and why foreclosing it was decisive rather than incidental.

### `docs/bitcoin-integration-status.md`

Two claims were false. Corrected in place with the correction **labelled**, rather than the text quietly replaced, because the gap between a status document and the tree is the thing worth recording:

- *"no path in the UI issues an order at all"* — there are two live call sites for `store_ops::submit_order_by_id` (`ui/src/state.rs:2863` and `:4390`).
- *"Buyer-seller messaging is not implemented"* with *"no callers outside their own tests"* — four production callers (`ui/src/messaging.rs:547`, `:709`, `:861`, `ui/src/state.rs:10422`), and `StoreInfoV1::encryption_public_key` exists at `common/src/store.rs:114`.

Both verified by grep rather than inherited from the handoff that flagged them.

## One idea costed out and rejected

The previous session proposed **bucketing committed amounts**, rounding up so declared exposure stays at or above actual and the cap still binds. Recorded in Part 8 as not worth doing, with the reasoning, because it would blur nothing:

- Every order has its own Bitcoin address, and that address is published **of necessity** — without it no third party can verify the payment, which is the entire point of publishing. Bitcoin amounts are public. A reader takes the address and sees exactly what was paid.
- Separately, there is only one amount field and it is simultaneously the exposure figure and the threshold the payment must cover. Round it up and the buyer must actually pay the rounded figure. Splitting it in two does not help, since public verification needs the true amount published anyway.

It reduces to a pricing convention: a seller wanting their revenue less legible can price at round numbers today, with no protocol change.

## A gap found while costing that, which is bigger than the idea

The document said a commitment *"reveals nothing about who Bob is or what he bought."* The first half holds. **The second does not** — a commitment names its listing, and listings are public, so the ledger is itemised rather than merely valued. Corrected at the step where the claim appeared, not only in Part 8, and filed separately because deciding what a commitment *should* reveal is a design call rather than a documentation fix.

## Testing

Documentation only, no code. HTML tag balance checked on the rewritten walkthrough (`div`, `h2`, `h3`, `p`, `table`, `tr`, `td`, `th` all balanced). Swept all of `docs/` afterwards for any surviving assertion that Lightning is required: none. The four remaining Lightning mentions in each document are deliberate contrasts explaining why it was rejected.

Em-dashes in added prose: none, per house style. The 56 remaining are in existing text and are left alone rather than churned into the diff.

[AI-assisted - Claude]

https://claude.ai/code/session_01Rw6QPkkPyb6aE8jE5LBbmA

### DIFF
--- a/docs/bitcoin-integration-status.md
+++ b/docs/bitcoin-integration-status.md
@@ -4,6 +4,15 @@ What works, and — more usefully — what does not, including pre-existing
 Harvest gaps that block the full buyer-side scenario and are unrelated to
 Bitcoin.
 
+> **Two claims in this document were false and are corrected in place below,
+> on 9 September 2026.** It said no path in the UI issues an order, and that
+> buyer-seller messaging is not implemented with no callers outside its own
+> tests. Both had been overtaken by the code and neither was updated, so a
+> reader using this file to decide what to build next was being pointed at work
+> already done. Corrected rather than deleted, because the gap between a status
+> document and the tree is the thing worth recording: a status file nobody
+> re-checks is worse than no status file, since it is read as current.
+
 ## Working
 
 - Orders in the store contract, with a monotonic status lattice and merge laws
@@ -55,11 +64,12 @@ mutable *state* instead would have been worse: `OrdersV1::verify` re-checks
 every order on every state validation, so rotating a shared mutable list would
 retroactively invalidate the whole historical order book.
 
-**What remains open is a different gap:** no path in the UI issues an order at
-all, so nothing yet chooses a bridge set at invoice time. When that path is
-built it has to supply one; an order that names none is unpayable, exactly as
-before, but now that is a per-invoice mistake rather than a permanent property
-of the store.
+**This is no longer open, and the text above is corrected.** It said no path in
+the UI issues an order at all. There are now two live call sites for
+`store_ops::submit_order_by_id` (`ui/src/state.rs:2863` and `:4390`), so orders
+are issued and a bridge set is chosen at invoice time. The requirement it named
+still holds: an order that names no bridges is unpayable, and that is now a
+per-invoice mistake rather than a permanent property of the store.
 
 The buyer side of the same move is already in place. Because the bridge set is
 per-invoice, checking a store's address once no longer tells a buyer who will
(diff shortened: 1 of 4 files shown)


sanity:
Heads up: #61 rewrites section 2 of `docs/bitcoin-integration-status.md` ("A watch is recorded and nothing is ever asked to synchronize it"), including the "Note this does **not** affect order-driven payment watching..." paragraph this PR's hunk uses as context. Whichever lands second will need that section reconciled by hand.

[AI-assisted - Claude]

sanity:
Another overlap to resolve by hand: #63 closes #57, so published orders no longer carry `listing_id`. Two passages in this PR's rewrite of `docs/design/incentive-mechanism.md` will become false once #63 lands:

- the status banner, which lists "a commitment reveals which listing it bought" as an open gap against the code;
- the Part 5 step 3 text saying the commitment "does reveal what he bought, because the commitment names the listing".

When rebasing, drop the banner item and say the listing is now carried only in the encrypted conversation (the buyer's request and the seller's acceptance), not in the published order. Please don't let automatic conflict resolution keep the old wording.

[AI-assisted - Claude]

Issue #60, opened 2026-09-16
build: a freenet-bitcoin bump re-keys the store contract even when no source changed


sanity:
Found while bumping `freenet-bitcoin-common` 0e5b9d9 -> a037181 for #59. The store contract re-keyed (`97db0a2e` -> `52107dcf`), recorded as V11 in `legacy/store_contract.toml`, although **no source it compiles in changed**:

- `common/` in freenet-bitcoin is byte-identical between the two commits: `git diff 0e5b9d9 a037181 -- common/` is empty.
- Nothing in Harvest's own contract source changed.

## Why the bytes moved anyway

`scripts/build-contract-wasm.sh` remaps `$CARGO_HOME` to `/cargo` (line 83), which strips the machine path. But Cargo checks a git dependency out under a directory that names its **commit**, and that segment survives the remap. The store WASM therefore embeds a panic-location string like:

```
/cargo/git/checkouts/freenet-bitcoin-ca471e53e3ade0e6/a037181/common/src/spv.rs
```

Before the bump it carried `.../0e5b9d9/...` instead. `strings` on the built WASM confirms it: the old commit appears zero times, the new one once.

Only the store moved because it is the one artifact that verifies SPV payment proofs and so reaches `spv.rs`. `reputation_contract`, `mailbox_contract` and `harvest_delegate` kept their hashes.

## Consequence

Every bump of the freenet-bitcoin git dependency re-keys the store, whether or not anything it runs changed. That is cheap today, since no store exists, but after launch each one strands every store's data at the old address until migrated, for a change that did nothing.

## Fix

Remap the commit segment of each git checkout ahead of the broader `$CARGO_HOME` rule (rustc applies the first matching prefix), e.g. `$CARGO_HOME/git/checkouts/<name>/<commit>` -> `/cargo/git/checkouts/<name>`. The locked commits can be read from `Cargo.lock`, or from `cargo metadata`, rather than globbed.

One cost to plan around: the fix itself changes the embedded paths, so it re-keys the store (and anything else embedding a git checkout path) **once**. Batch it with the next deliberate re-key rather than landing it alone. #57 (if option 2 or 3 is taken) and the #22/#26 batch are both candidates.

[AI-assisted - Claude]

PR #61, opened 2026-09-17
feat(payments): register order addresses with the bridge over Freenet, resolved through generation pointers

### LINKED ISSUE #29: payments: the published app cannot reach the Bitcoin bridge — CSP blocks it, and the failure is a console warning
Confirmed by publishing Harvest and loading it on a **second machine's** Freenet node. Not reachable by any test in the repo.

## What happens

`ui/src/gateway/bitcoin_bridge_http.rs` fetches bridge status over HTTP:

```
Connecting to 'http://127.0.0.1:8431/v1/status' violates the following Content
Security Policy directive: "connect-src http://127.0.0.1:7509 blob: data:".
Fetch API cannot load http://127.0.0.1:8431/v1/status.
```

The published app is served with `connect-src http://127.0.0.1:7509` — its own gateway and nothing else. So **a published Harvest can never reach a Bitcoin bridge over HTTP, on any machine, regardless of whether one is running.** It is not a configuration problem and no bridge endpoint setting can fix it.

The app degrades rather than breaking: it logs `WARN … Failed to fetch bridge status` and carries on. But that warning is in a browser console, so from the user's side the Payments surface is simply wrong with no explanation.

## Why this is not the same as the known watch-path gap

The architectural constraint is documented — a delegate has no HTTP variant in `OutboundDelegateMsg`, and a published page's CSP blocks direct fetch. What is new here is **confirmation that the app actively tries anyway, on every load, and reports the failure only to a console.**

Worth being precise about what it does and does not break:

- **Payment verification is unaffected.** Proofs are assembled from bridge-signed claims read out of the address contract *through Freenet*, not from the bridge over HTTP.
- **Bridge status display and anything needing a live bridge query is dead** in a published build, and looks like a bug rather than a limitation.

## What to decide

Either stop attempting the fetch in a published context and say plainly on the Payments surface that live bridge status is unavailable there — the same honesty the messaging and watch surfaces already apply — or route the query through the node the way everything else goes, if that is possible at all.

What should not stand is the current shape: a fetch that cannot succeed, attempted on every load, whose failure only a developer sees.

Found alongside #28 by loading the published app for the first time. Both are CSP-shaped, both invisible to the test suite, both about ninety seconds of actually running it.

[AI-assisted - Claude]

### LINKED ISSUE #30: bitcoin: build-time addressing constants are stale, so payment silently cannot complete

### LINKED ISSUE #59: payments: nothing registers an order's address with a bridge, so no order can ever reach Paid
## Problem

**An order can never reach `Paid`, on the published app or anywhere else.** Nothing in Harvest ever asks a bridge to watch an order's payment address.

A bridge only scans scripts on its own watch list. On the live bridge that list holds only its configured demo addresses. So when a buyer pays an order's per-order address, the bridge never looks at it, no claim is ever published to that address's contract, and the order stays `AwaitingPayment` however long anyone waits.

Verified by reading the code, 2026-09-10:

- The only `freenet_bitcoin_common::WatchRequest` constructions in Harvest are in tests: `delegates/harvest-delegate/src/bitcoin.rs` (`a_stored_labels_watch_never_leaks_into_a_bridge_watch_request`) and `common/src/bitcoin_delegate.rs` (`the_bridge_watch_request_has_nowhere_to_put_a_private_label`). Both only check that a private label cannot leak into the wire format.
- Nothing calls a bridge's `/v1/challenge` or `/v1/request`.
- The bridge adds scripts to `watched_scripts` only from those HTTP requests or from its own `always_watch` config (`freenet-bitcoin`, `bridge/src/service.rs`, `bridge/src/main.rs`).

The "built and working" payment path was true piece by piece (SPV proof, claim verification, the `Paid` transition), and the link between them was never built. It was never caught because no store has ever existed, so nothing has run end to end.

## Why it could not simply be wired up

A bridge takes watch requests only over HTTP, and neither a published webapp (the gateway's `connect-src` allows only its own node) nor a delegate (no HTTP capability) can reach one. That is being fixed in freenet-bitcoin, where the request side moves onto a Ghost Key gated inbox contract: freenet/freenet-bitcoin#3.

## Work on the Harvest side, once that lands

- When an order's address is derived, send a watch request for it through the bridge's inbox, dated by the Bitcoin mainnet tip height from the bridge's tip contract. The UI can do this as a client immediately; the seller's delegate takes it over when delegate-side auto-accept is built.
- Read the bridge's status contract instead of the compiled-in constants in `ui/src/gateway/bitcoin_config.rs` (#30).
- Delete `ui/src/gateway/bitcoin_bridge_http.rs`, which can never succeed in a published build (#29).
- Confirm success through the address contract's per-bridge scan watermark, and resend if it never appears.

This is the real blocker to a first sale. #29, which was previously thought to be one, is not: payment verification already reads bridge claims through Freenet.

[AI-assisted - Claude]

sanity:
## Problem

No order could ever reach Paid (#59). The bridge only scans scripts someone has asked it to watch, and nothing in Harvest asked. Two related problems blocked any fix:

- #30: the UI carried build-time addressing constants (address contract code hash, tip contract id) that go stale whenever the bridge redeploys. A stale constant fails as "payment never confirms", with nothing on screen to say why.
- #29: the only path to the bridge was an HTTP call that the published app's CSP blocks, and the failure showed up only as a console warning.

## Approach

Everything now goes over Freenet, and nothing about the bridge's contracts is compiled in.

**Generation pointers (#30).** The bridge signs a pointer for each of its artifacts (address contract, request inbox, tip contract). `bitcoin_generation` resolves them with `freenet-migrate`'s `PointerResolver`, and `gateway/bitcoin_generation_ops` does the GETs, timeouts and backed-off retries. Settled pointers are refreshed every 10 minutes from their settled floor. A bridge redeploy is followed, an older record is refused, and a refresh that learns nothing keeps the settled answer. There is deliberately no fallback: until the address generation resolves, `order_for_invoice` refuses to issue an invoice and tells the seller why. The tip contract is derived from the resolved tip generation, and a replaced tip or inbox contract is retired.

**No HTTP path (#29).** `bitcoin_bridge_http.rs`, `scripts/check-bitcoin-constants.sh` and the Makefile publish gate that guarded the constants are removed. `bitcoin_config` is now trust policy only.

**Watch requests (#59).** A seller's tab asks the bridge to watch the payment scripts of its own unpaid, anchored orders that name the bridge, through the bridge's request inbox contract:

- Each request is sealed to the bridge, bound to the store's verified seller key (`GhostkeyId`), signed by the ghostkey delegate through the existing `SignResult` matching, and submitted with the floor it was dated against.
- **Allowance:** the planner stays within the inbox's 2 unread entries per Ghost Key, counting entries in the inbox, still landing, and awaiting signature. Each request carries up to 32 scripts, newest orders first.
- **Renewal:** a watch lasts about a day, so it is renewed at 12h. A request that left the inbox unread is sent again, judged only against a state received after its landing grace. A read the bridge was seen to make is remembered, because the floor later prunes the evidence.
- **Freshness:** the inbox is fetched only by a node with an order to watch, refetched every 10 minutes, and not planned against if stale. A request that left the inbox unread is sent again. If a key's requests for scripts it still wants go unread for two hours, the seller is told once for as long as it lasts. That covers requests that keep expiring and being resent, and entries sitting in this node's copy whoever sent them. The clock restarts after the tab has not been running.
- **Containment:**
  - Requests are signed only with a Ghost Key the vault has listed for this app, one key at a time.
  - They wait, for up to 10 minutes at a stretch, while the seller signs anything of their own.
  - A refusal that can only concern a watch request clears nothing of the seller's.
  - Any refusal, silent timeout or unusable signature stops that key for the session, with one notice. The usual cause is a grant revoked while the tab was open, which prompts once. Retrying would raise the dialog again. A late answer to a timed-out prompt is ignored.

Tracking is in memory: a reload sends every wanted request once more, as an early renewal.

**Store contract re-key.** Moving `freenet-bitcoin-common` to the commit the inbox crates resolve from changed the store WASM (V11 in `legacy/store_contract.toml`). No source the contract compiles in changed. The difference is a panic-location string that includes the git checkout's commit. Details and the follow-up (#60) are in the registry entry. The reputation, mailbox and delegate addresses are unchanged, as the drift check confirms.

**Known limits** (also in `docs/bitcoin-integration-status.md`):
- A payment mined before the bridge reads the request is not found. The bridge ignores `scan_from_height` (freenet-bitcoin#7).
- Pointer floors are not persisted, so the first resolve after a load can take a superseded record a peer serves.
- The public inbox links a store's seller key to its bridge, and shows request timing and script counts. Payment addresses were already public.
- Manual watches (`WatchForm`) are still not sent to the bridge.

## Testing

- `bitcoin_inbox`: tests against the real inbox contract logic under a throwaway Ghost Key authority. They cover admission and bridge unsealing, real signed removals and floor pruning, the per-Ghost-Key allowance, landing and awaiting entries, per-network requests, `scan_from_height`, strictly increasing `made_at_ms`, renewal, state-based landing grace, freshness, fetch backoff, and the unread notice (runs carried across resends, entries from other tabs, wanted scripts only, sleep gaps).
- `bitcoin_generation`: resolution against real signed records, including live pointer ids and the live signet tip derivation. Refresh, stale and forged records, withdrawal, and retry arithmetic.
- `state`: end to end on real pieces. A served inbox queues the request, the delegate's answer is matched through `on_ghostkey_response`, the submission is admitted by the inbox, and the bridge opens it to the order's script and anchor. Also covered:
  - every guarded refusal arm, and a refusal naming another key
  - wrong-key signatures and silent timeouts
  - the bounded hold, and unlisted keys
  - replaced and withdrawn contracts, tip retirement
  - buyers not fetching, and order selection
  - the store-state trigger
- Mutation-checked by hand: 25 before review, then 18, 20, 11, 6, 9 and 5 in the review rounds. Survivors led to stronger tests, or to removing a condition that turned out to be redundant.
- Not tested: the wasm-only glue in `gateway/bitcoin_generation_ops.rs` (the repo has no wasm test harness), and a live run against the bridge.
- `cargo test --workspace`, `cargo clippy --workspace --all-targets` and wasm32 clippy are clean. `cargo make code-hashes`: store 52107dcf, others unchanged.

Reviewed at Full tier in six rounds so far (review comments on this PR). Codex was unavailable for quota reasons, so Claude lenses were substituted.

Closes #59
Closes #29
Closes #30

[AI-assisted - Claude]


sanity (commented):
## Comprehensive PR Review: #61 (round 1)

### Summary
- **PR Title:** feat(payments): register order addresses with the bridge over Freenet, resolved through generation pointers
- **Type:** feat
- **CI Status (reviewed head 50ca30d):** fmt/clippy/test, wasm build and committed-wasm checks pass. "No contract or delegate address moved" fails, as expected, for the store re-key recorded as V11.
- **Linked Issues:** #59, #29, #30
- **Review tier:** Full (signing and sealing, inbox wire format, contract WASM re-key)
- **Reviewers run:** code-first, testing, skeptical, big-picture (freenet reviewer agents), plus two substitute lenses: protocol compliance against the inbox contract, bridge and pointer code, and crypto/trust/privacy.
- **External model:** `codex review` was tried twice and failed both times with "You've hit your usage limit ... try again at Sep 19th". Per the review rule, two extra Claude lenses were substituted for it. These share the authoring model's blind spots, so a Codex pass is worth running once quota returns.

### Code-First Analysis
**Alignment:** matches. The description is accurate, including its stated gaps.
**Gap:** a signature returned for an inbox the tab had since replaced was still submitted to the old inbox.

### What held up (verified)
- Pointer records are verified against the configured bridge key and bound to their artifact. A NotFound or forged reply cannot redirect resolution, only delay it.
- An inbox signing payload cannot be confused with a listing, store-info or order payload (distinct domain prefix, and matching is on exact bytes).
- The sender obeys the inbox rules: one entry per delta with its floor, `sender_height`, the per-Ghost-Key cap of 2, 32 scripts per request, strictly increasing `made_at_ms`, and canonical certificates.
- The V11 re-key: `common/` is identical between the two freenet-bitcoin commits, the lockfile only adds packages, and CI shows reputation, mailbox and delegate unchanged.
- No APP_STATE or RefCell re-entrancy on the new paths.

### Findings and disposition

Everything below was verified against the code and is fixed in 7d2d46a unless marked otherwise.

#### Must Fix
1. **Background signing could destroy the seller's work.** Watch requests shared `pending_signatures` with listings, invoices and store info. Every ghostkey refusal clears the whole queue and pending store creation/edit, pushes a notification, and the request came due again within a minute. **Fixed:** no watch request is queued while the seller has a signature, store creation/edit or access prompt outstanding. A refusal that can only concern a watch request is handled by a guarded arm that clears nothing of the seller's, notifies once, and backs off (5 min doubling to 6 h). Tests: `a_refused_watch_signature_is_contained`, `a_refusal_with_the_sellers_own_signature_outstanding_is_theirs`, `no_watch_request_is_queued_while_the_seller_is_signing`.
2. **Pointers resolved once per tab.** A bridge redeploy while a tab stayed open left invoices naming the old address contract and requests stuck in an inbox nothing reads, filling both Ghost Key places for good. **Fixed:** settled pointers are refreshed every 10 minutes from their settled floor. Older records are refused, a refresh that learns nothing keeps the settled answer, a withdrawal is lifted only by a newer record, and a new inbox generation replaces the tracker. Six new tests in `bitcoin_generation`, plus `a_new_inbox_generation_replaces_what_was_sent_to_the_old`.
3. **Status doc contradicted the PR.** `docs/bitcoin-integration-status.md` still described #59 as open and cited the deleted HTTP module. **Fixed:** the section is rewritten, including known limits and what the public inbox reveals.

#### Should Fix
4. A signature that never returned held a Ghost Key place for the whole session. **Fixed:** given up after 5 minutes.
5. A failed first inbox GET disabled watching for the session (the id was marked subscribed before the GET, and NotFound was ignored). **Fixed:** fetches are tracked in the tracker, retried after 60 s, and refetched every 10 min in case a subscription ended quietly.
6. No floor freshness check. **Fixed:** nothing is planned against a state older than 15 min.
7. Every visitor, buyers included, subscribed to the inbox. **Fixed:** only a node with an order to watch fetches it.
8. A signature for a replaced inbox was still submitted. **Fixed:** not sent.
9. Wrong-key or unusable signatures looped. **Fixed:** they back off and notify once.
10. Pointer GETs were sent before the response loop started, so slow delegate registration timed out first attempts, and late pointer answers fell through as app state. **Fixed:** resolution starts just before the loop, and any answer at a pointer id is consumed.
11. Retries had no backoff or jitter. **Fixed:** doubling from 30 s to 10 min, with ±20% jitter.
12. Orders with no anchor were watched indefinitely. **Fixed:** not watched.
13. Test gaps: `register_inbox_contract`, the store-update trigger, and a landing entry being counted twice. **Fixed:** tests added. The store trigger turned out to be testable through `on_contract_state`, contrary to the PR description.
14. A doc comment was detached from `spawn_store_info_signature`, and the `WatchSyncStatus` docs were stale. **Fixed.**

#### Consider (not changed, with reason)
- **Payment before the watch applies (protocol lens).** The bridge ignores `scan_from_height` (freenet-bitcoin#7, open), so a payment mined before the bridge reads the request is missed. That is a bridge-side fix. It is documented in the status page and the `WatchWanted` docs, not worked around here.
- **Pointer floor not persisted.** A peer can serve a superseded pointer on the first resolve after a load, until the next refresh. It is documented as a follow-up.
- **`made_at_ms` across reloads and devices, and "read" not meaning "acted on"** (e.g. over the bridge's per-sender cap). Documented on `InboxTracker`. Neither can be seen from this side without bridge feedback.
- **Unwatch when paid:** not sent. The bridge lets an interest lapse after a day.
- **Inbox state taken from the local node without `verify`:** the local node validates contract state, and a full verify costs RSA checks on every update.
- **Privacy:** the inbox links a store's public seller key to the bridge it uses, and shows request timing and script count. Payment addresses were already public. Documented, and padding would be a freenet-bitcoin change.
- **`docs/privacy.md`** is referenced from `common/src/bitcoin_delegate.rs` but does not exist. This is pre-existing and was not touched here, because editing that file risks moving the delegate WASM.

### Testing
Every new behaviour has a unit test, most against the real inbox contract logic and real signed pointer records. Hand mutation checks: 7 on the due rule, 18 on the planner and state wiring, and 18 on the review fixes. Each mutation was confirmed to fail a test. Not exercised: the wasm-only glue in `gateway/bitcoin_generation_ops.rs` (no wasm test harness in the repo), and a live run against the bridge.

### Process
- `contract-rekey-acknowledged` was applied by the authoring agent, not a human. That was after checking the drift job's output (only the store moved, matching the V11 registry entry) and on the maintainer's standing guidance that pre-launch Harvest re-keys are cheap. @sanity, remove it if you want to make that call yourself.

### Verdict
**State:** Needs Changes, Re-review Required After Fix. Must-fix findings were addressed in 7d2d46a, so that head needs a fresh Full-tier review, which is running now.
**HEAD SHA reviewed:** 50ca30d80e1a4d61c7db3ccb1fd535cbd19ae39b

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #61 (round 2, re-review of 7d2d46a)

### Summary
- **Review tier:** Full (re-review required: round 1 had must-fix findings)
- **HEAD SHA reviewed:** 7d2d46aa6b66e557ee34b1b50dc54c67147f2b5a
- **Reviewers run:** code-first, testing, skeptical, big-picture, plus substitute lenses for protocol compliance/timing and crypto/trust/privacy. `codex review` is still out of quota ("try again at Sep 19th"), so Claude lenses stood in again.

### Round-1 findings
All 14 were confirmed resolved in the code by more than one reviewer. Three were resolved only partly; the residuals are below. No reviewer found a must-fix in round 2.

### What held up (verified)
- **Pointer refresh:** a peer cannot roll a tab back, alternate two genuine records, or pin a withdrawal. Stale, competing and forged records keep the settled answer, and a re-served tombstone yields Withdrawn, not the zero hash.
- **Pointer ids:** they are derived from the bridge key and the artifact, so consuming answers at them cannot suppress app state.
- **Guards:** no re-entrancy, and no early return skips `start()`.
- **Big picture:** closing #59 is honest. The remaining gap (a payment mined before the bridge reads the request) is freenet-bitcoin#7, which is open, predates this PR, and is linked from the status page.

### Findings and disposition (fixed in acd8c6c unless noted)

#### Should Fix
1. **Background signing could raise a vault prompt.** Confirmed in the ghostkey delegate (`request_user_permission` for a key without a Sign grant). **Fixed:** watch requests are signed only with a key the vault has listed for this app. Listing requires a grant, and a third-party grant includes Sign.
2. **Refusals were tracked tab-wide.** One broken Ghost Key could starve or spam another store's watches. **Fixed:** backoff and notification are per Ghost Key, and a refusal naming a fingerprint drops only that key's requests.
3. **A silent signature timeout had no backoff.** **Fixed:** it counts as a refusal of that key.
4. **Leaked seller state stopped watching for the session.** A store creation, prompt or signature that never clears could hold requests back indefinitely. **Fixed:** the hold is bounded at 10 minutes.
5. **The freshness gate measured arrival, not the floor.** A node serving a copy it stopped following looked fresh, and the tab's own dropped entries filled both places. **Fixed:** when the floor has not advanced for 90 minutes, planning pauses, the seller is told once, and the inbox is refetched with backoff.
6. **Land grace was judged by the clock.** A quiet inbox made new requests look dropped. **Fixed:** absence is judged only against a state received after the grace period.
7. **A tip generation change left the old tip contract feeding the same `TipView`, and a replaced or withdrawn inbox stayed in use.** **Fixed:** replaced contracts are retired and their late states ignored, a withdrawn inbox is dropped along with what was waiting for it, and old-inbox requests no longer count against the new inbox.
8. **Status doc errors.** It said requests were "sealed to the store's seller key", presented "about a minute" as a bound, said the bridge reads its inbox "each block", overstated what a refresh corrects, and did not mention what fetching the inbox reveals. **Fixed.**
9. **Test gaps:** four of the five guarded refusal arms untested, wrong-key backoff not asserted, two outstanding requests under one refusal untested, retry arithmetic untested. **Fixed:** there is now a parameterised test over all five arms, plus the other cases, and the retry arithmetic moved into a pure function with a test.
10. **The per-minute check still wrote state for any seller with a store.** **Fixed:** it writes only when something could be watched.

#### Consider (not changed, with reason)
- **A user action started while a watch signature is outstanding, and then refused, still goes to the ordinary arm.** With listed-key gating, `SignMessage` does not prompt and is answered immediately, so the window is milliseconds. No change.
- **Buyers still resolve the three pointers.** Buyers need the tip generation to judge anchors and depth, and the address and inbox GETs are a few hundred bytes every 10 minutes. No change.
- **Refresh interval has no jitter.** Tabs only refresh in step if they loaded together, and the cost is three small GETs. No change.
- **A late valid pointer answer that arrives after a slot failed is consumed rather than used.** It is retried within the backoff. No change.
- **The wasm-only glue** (timers, `deliver_*`, settle) still has no tests, because the repo has no wasm test harness. The decisions it makes are in tested pure code.

### Mutation checks
There were 20 new hand mutations. One (removing the listed-key filter) was first applied as a mutation that could not fail; redone properly, it is caught. Another (ignoring retired contracts) survived and led to a stronger test. All are caught now.

### Related work
PR #56 edits the section of `docs/bitcoin-integration-status.md` this PR rewrites, so it will need rebasing; a note has been left there.

### Verdict
**State:** Needs Changes, Re-review Required After Fix. Several high-risk-area should-fixes changed in acd8c6c, so a third pass on that head is running.
**HEAD SHA reviewed:** 7d2d46aa6b66e557ee34b1b50dc54c67147f2b5a

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #61 (round 3, re-review of acd8c6c)

### Summary
- **Review tier:** Full. A re-review was required because round 2 changed several high-risk areas.
- **HEAD SHA reviewed:** acd8c6c
- **Reviewers run:** skeptical, code-first, testing, and a combined substitute lens for protocol, trust and user safety. `codex review` is still out of quota.

### Findings and disposition (fixed in e5e6c4c)

#### Must Fix
1. **A revoked grant still raised a vault prompt from the background loop.** `AppState.ghostkeys` only grows, so a key revoked while the tab was open still passed the listed-key gate. The delegate then prompts (`request_user_permission`). Per-key backoff re-raised the dialog every 5 minutes to 6 hours, and clicking Allow to make it stop would quietly undo the revocation. **Fixed:** any refusal, silent timeout, or wrong-key/unusable signature stops watch signing for that key until the page is reloaded, with one notice. There is no retry, so the dialog appears at most once.

#### Should Fix
2. **An unnamed refusal (`Error`, `AccessDenied`, `NoIdentityAvailable`) still backed off every key with an outstanding request**, so one broken key could starve another. **Fixed:** requests go out one Ghost Key at a time, and none while one is still with the delegate, so an unnamed refusal is always about exactly one key.
3. **The frozen-floor pause was wrong for a real case, and its refetch was not backed off.** During a bridge mainnet-node outage the floor stands still, but the bridge still reads entries dated against it, so pausing lost those watches. The refetch also ran every minute for the whole outage. **Fixed:** the pause is removed. Instead, this tab's own request left unread for 30 minutes (the bridge polls every 30 s) is sent again and stops holding its place. That covers a node serving a copy it stopped following, at the cost of one signature per half hour while the fault lasts.
4. **The hold on watch requests while the seller signs added up across separate actions.** **Fixed:** it restarts whenever a check finds nothing of the seller's under way.
5. **A state did not reset the fetch spacing.** **Fixed.**
6. **A withdrawn inbox was only logged.** **Fixed:** the seller is told.
7. **Docs overstated isolation and "never raises a prompt", and code comments still said "sealed to the seller key".** **Fixed.**
8. **Test gaps** (hold reset, the pending-request case of `watch_check_could_act`, and others). **Fixed:** there are tests for every new behaviour. 11 hand mutations were checked, and the two survivors (queuing while a request is out, and fetch-spacing reset) led to stronger tests.

#### Consider (not changed, with reason)
- **The listed-key assumption.** That listing implies a Sign grant rests on the ghostkey delegate's grant functions (`grant_full`/`grant_third_party` only ever write sets that include Sign). If that changed, the first background request would prompt once and then stop the key, so the damage is bounded.
- **A retired tip contract keeps `TipView` at its last values until the new contract answers.** This fails safe (depth is under-counted), and the old contract stops advancing anyway once the bridge moves off it.
- **Retired contracts stay subscribed**, and their updates trigger a re-GET that is then ignored. Harvest has no unsubscribe path. The cost is bounded to generation changes.
- **A sleep of more than 5 minutes with a request outstanding stops that key**, via the timeout, until reload. This is rare, and the notice says to reload.
- **The wasm-only glue** is still untested (no wasm harness).

### Verdict
**State:** Needs Changes, Re-review Required After Fix. A must-fix was addressed in e5e6c4c, so a fourth pass on that head follows.
**HEAD SHA reviewed:** acd8c6c

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #61 (round 4, re-review of e5e6c4c)

### Summary
- **Review tier:** Full (re-review after a round-3 must-fix)
- **HEAD SHA reviewed:** e5e6c4c
- **Reviewers run:** skeptical, testing, and a substitute lens for protocol, trust and user safety (checked against the ghostkey delegate, the inbox, the bridge and freenet-core). `codex review` is still out of quota.
- **Must-fix:** none found.

### Findings and disposition (fixed in 231c976)

#### Should Fix
1. **The 30-minute resend did not do what it claimed, and could stall a key.** Two reviewers found this independently. In a node serving a copy it stopped following, the resend was dated against that copy's passed floor, so peers that had moved on dropped it. `record_sent` overwrote each script's record, so earlier entries stopped counting as this tab's. They then held both places, and a resend losing the equal-height tie against its own earlier entry left the key blocked. A bridge that is not running is not helped by resending either. **Fixed:** the resend is removed. Every sent entry is tracked, and the seller is told once if one sits unread for two hours. That is well past the bridge's legitimate deferrals: its per-sender read budget, and a few blocks before the floor passes an entry.
2. **A refusal of a background prompt could still reach the ordinary arm.** A Deny given after the 5-minute timeout, or while the seller had started an action, cleared the seller's state. In the mid-action case the key was not stopped, so the dialog came back. **Fixed:** a refusal naming a key with a watch request outstanding stops that key even mid-action. A late refusal for a key already stopped is ignored when nothing of the seller's is under way.
3. **Status doc overclaims.** It said "signing raises no prompt", which is false for a grant revoked mid-session (that prompts once). It said the bridge "would not allow" 30 minutes unread, which budget deferral and slow blocks both allow. **Fixed.**
4. **The hold-restart test passed for the wrong reason.** It never produced the tick where nothing is wanted. **Fixed:** that tick is now covered, and moving the reset back to its old position fails the test.

#### Consider (not changed, with reason)
- **A host-level delegate error or node overload that loses an answer stops the key until reload.** Harvest has no reconnect path, and the delegate returns `Error` for a granted key only when that key or its certificate is missing. It is rare, and the notice says to reload.
- **A key whose submissions keep failing to send** (`update_contract` error) is re-signed about once per state arrival, without prompting. That costs signatures, not user attention.
- **Re-listing keys before signing or stopping** would avoid even the single prompt after a revocation. It is left as a follow-up, since it adds a round trip and the current worst case is one dialog.
- **One key at a time:** a key whose request is stuck behind a prompt delays other keys by up to 5 minutes, then it is stopped.
- **If a refusal naming a stopped key arrives while the seller's own work is also under way**, it is handled as the seller's, because the two cannot be told apart.

### Mutation checks
6 new hand mutations were checked, and each was caught.

### Verdict
**State:** Needs Changes, Re-review Required After Fix. A fifth pass on 231c976 follows.
**HEAD SHA reviewed:** e5e6c4c

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #61 (round 5, re-review of 231c976)

### Summary
- **Review tier:** Full (re-review)
- **HEAD SHA reviewed:** 231c976
- **Reviewers run:** skeptical, testing, and a substitute protocol/trust lens (checked against the inbox, the bridge, the ghostkey delegate and freenet-core). Codex is still out of quota.
- **Must-fix:** none.

### Findings and disposition (fixed in 2f3990e)

#### Should Fix
1. **A request that keeps expiring unread was resent every ~30 minutes forever, with no notice.** In a copy whose floor still moves, an unread entry is pruned after about 3 blocks, before the 2h notice could see it. This covers a submission that never reaches the bridge's node, and an entry the bridge will not read. **Fixed:** each script's unread run is tracked across resends and ended by a read, and the notice fires on the run.
2. **After a reload, a stale copy's entries held the key's places with no notice.** The notice only looked at this tab's own sends, and its advice to reload led straight to that state. **Fixed:** entries of the key's in the node's copy are tracked by when they were first seen, whoever sent them. The notice no longer suggests reloading.
3. **Status doc contradicted itself** ("sent again if it left the inbox unread" and "nothing is resent"). **Fixed.** It also now says that a revoked grant the seller allows again resumes watching.
4. **Doc comment for `only_watch_signatures_outstanding` was glued onto `named_refusal_is_a_watch_request`.** **Fixed.**
5. **Test gaps:** no plan check at or after the notice threshold, no `KeyNotFound` late-refusal case, no test that the notice re-arms. **Fixed.** Hand mutations were checked: one survivor was a redundant condition, now removed, and the other led to a new assertion.

#### Not changed (disputed)
- **The seller's own prompt overwriting a background watch prompt, so the seller's Deny is taken for the watch request's and their save hangs** (skeptical lens). The protocol lens showed this cannot happen, and freenet-core confirms it. `crates/core/src/contract/delegate_park.rs` (#5544) queues further requests to a delegate while it has a prompt parked, and before #5544 the loop awaited prompts inline. Answers therefore come back in request order, and a prompt auto-denies after 60 s. A `PermissionDenied` naming a key with a watch request outstanding is that request's.

#### Consider (not changed)
- **A read the bridge did not act on** (over its per-key cap, or bridge database loss) stays quiet until the 12h renewal. This side cannot see it.
- **An unnamed `Error` from a broken vault** (the key's signing key or certificate missing) arriving while the seller is mid-action goes to the ordinary arm. That needs a broken vault.
- **Two queued seller requests for a stopped key, both denied:** the second one's notice is swallowed. Only a duplicate is lost.

### Verdict
**State:** Needs Changes, light re-check. A scoped sixth pass on 2f3990e follows.
**HEAD SHA reviewed:** 231c976

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #61 (round 6, scoped re-review of 2f3990e)

### Summary
- **Review tier:** Full, scoped to the round-5 commit.
- **HEAD SHA reviewed:** 2f3990e
- **Reviewers run:** skeptical, testing, and a substitute protocol lens. Codex is still out of quota.
- **Must-fix:** none. Planning and sending are unchanged by 2f3990e; the findings are all in the unread notice.

### Findings and disposition (fixed in c838c92)

#### Should Fix
1. **A script that stopped being wanted mid-run latched the notice.** Found by two reviewers independently. `request_long_unread` counted a `!read` record for any script. An order paid, cancelled or expired while its request was going unread left a run that never ended. That gave a false notice to a seller whose order was already Paid, and because the condition never cleared, a later real fault was never told. **Fixed:** only scripts still wanted count.
2. **False notice on waking from sleep.** Send an invoice, close the lid before the removal is seen, and the bridge's read and its removal are pruned overnight. The first check on wake judged hours-old evidence. **Fixed:** the unread clocks restart after a gap in checking longer than 5 minutes (checks come every minute while anything is wanted), and nothing is judged on a state older than 15 minutes.
3. **The notice did not re-arm while nothing was wanted.** The reset sat after the early return, so a fault on a later order after an earlier notice was never told. **Fixed.**
4. **Test gaps:** a renewal after a read was untested, and the re-arm test only checked the flag, not a second notice. **Fixed.** There are tests for a renewal after a read, a script no longer wanted, the sleep gap, stale state, and a second fault being told. 5 hand mutations were checked, and each was caught.

#### Consider (not changed)
- **A requested change to the set of keys during a continuing fault can repeat the notice.** It needs a store re-fetch or a key stop mid-fault, and the result is a duplicate, not a wrong notice.
- **Requests that cannot be sent at all** (the inbox never served) are not covered by this notice. The status doc now says so.
- **The notice text did not mention a bridge that reads but will not act.** The text and doc now include a bridge that refuses requests.

#### Confirmed
- The round-5 disposition of the "prompt overwrite" finding holds for Harvest's path. Client `DelegateRequest`s go through `dispatch_delegate_request`, which queues first-in-first-out behind a parked prompt (freenet-core `contract.rs`, `delegate_park.rs`). Re-registration queues too. A full queue returns an error naming no key, so it cannot be taken for a watch request's refusal.
- A 2-hour threshold is safely beyond every legitimate deferral in the bridge (per-key read share, global budget, a mainnet node outage, the 2-minute floor hold on restart).

### Verdict
**State:** Needs Changes, light re-check. A scoped check of c838c92 follows.
**HEAD SHA reviewed:** 2f3990e

[AI-assisted - Claude]

sanity (commented):
## PR Review: #61 (round 7, scoped light re-check of c838c92)

- **Reviewers run:** skeptical and testing, scoped to c838c92. Codex is still out of quota.
- **HEAD SHA reviewed:** c838c92
- **Must-fix:** none upheld.

### Findings and disposition
1. **Should fix: the freshness gate silenced the notice when a node stops serving the inbox** (skeptical). That is one of the faults the notice names. The gate was also redundant, because `note_check` restarts the clocks before anything is judged after the tab has not been running. **Fixed in 6a7ac91:** the gate is removed, and the test that locked it in now expects the notice.
2. **Reported must-fix: entries seen in the inbox are not filtered to wanted scripts** (testing). **Not changed, with reason.** An entry's scripts are sealed to the bridge, so the tab cannot know them. An entry sitting unread in this node's copy for two hours means the inbox is not being read: the bridge polls every 30 s, and the floor passes an entry within a few blocks. That affects every script the key still wants, so the notice is correct, not false. The skeptical lens reached the same conclusion independently. The docs and the function's doc comment now say this precisely.

### Confirmed
- `note_check` runs at least once a minute while anything is wanted: the timer gate matches its reach, and the state-arrival path calls it too. So a running tab never restarts the clocks by accident.
- The new tests pin their behaviour for the right reasons, including that the second notice comes from an independent fault.

### Verdict
**State:** a light re-check of 6a7ac91 (gate removal and docs) follows.
**HEAD SHA reviewed:** c838c92

[AI-assisted - Claude]

PR #63, opened 2026-09-17
fix(privacy): stop published orders revealing which listing was bought

### LINKED ISSUE #57: privacy: an order commitment reveals which listing it bought, which the design says it does not
Found while costing out a proposal to bucket committed amounts for privacy. The amount turned out not to be the binding leak, and this is.

Filing publicly rather than privately because nothing is exposed today: no store has ever been created, so there are no orders and no buyers. It is also visible in the type signature rather than being a method anyone needs explained. If a store is created before this is settled, that calculus changes.

## The gap

`docs/design/incentive-mechanism.md` said, of the order commitment:

> It reveals nothing about who Bob is or what he bought.

The first half holds. The second does not.

`Order.listing_id` (`common/src/payment.rs:311`) is part of the published commitment, and `ListingsV1.listings` (`common/src/store.rs:233`) is public store state. So anyone who can read the store can join the two and see **which item each order bought**, alongside its amount and its approximate time.

The sentence is corrected in #56. What that PR does NOT do is decide whether the behaviour should change, which is why this is separate.

## Why it is more than a documentation slip

Part 8 of the design accepts a real, structural privacy cost on the seller side and argues the trade is not optional, because the public commitment is what makes exposure countable. That argument is sound for the **amount**: a stranger must be able to add up what a seller owes.

It does not obviously extend to the **listing**. Counting exposure needs the value of each open order. It does not need to know which item it was. So the itemisation may be cost without corresponding benefit, which is the part worth a decision.

It also changes what the ledger is. Part 8 describes "an itemised, timestamped ledger of a seller's business: order values, order frequency, revenue rate, growth, quiet periods". With the listing id included it is additionally a per-product sales record: which lines sell, at what rate, to how many distinct buyers, and when a product stops selling. For a seller that is competitive information. Combined with the buyer binding it also groups a single buyer's purchases within one conversation.

To be precise about "public": reading a store's contract needs its key, which comes from the twelve-character code in the store link (#52). There is no directory of stores. But the link is the thing a seller hands out to attract buyers, so for any store doing business this is effectively public, and on Freenet a peer hosting the record receives every update pushed to it for free and undetectably.

## Options, not a recommendation

I have deliberately not picked one, since this is a design call.

1. **Accept and document it.** Cheapest. Part 8 gains a sentence saying the ledger is itemised per product, and the privacy analysis follows. Arguably right if Harvest is for ordinary commerce, which is the settled answer to Part 10.
2. **Replace `listing_id` with a commitment to it.** The buyer can verify their own order names the listing they asked for, because they know which one that is; a stranger counting exposure does not need to. Needs care: the order id already hashes the whole order, so this is a change to what is published rather than to what is signed, and any such change re-keys the store contract.
3. **Drop it from the published order entirely** and keep it only in the encrypted conversation. Simplest to reason about, but worth checking against whatever currently reads `listing_id` off a published order, which is at minimum the buyer's own request-matching (`Self::requested_listings`).

Option 2 or 3 should be batched with any other store-contract re-key rather than taken alone, since a re-key is free only while no store exists.

## Related

- #56 corrects the sentence.
- #52 is the twelve-character store code, which is what makes "public" mean "anyone with the link".
- Part 8 of `docs/design/incentive-mechanism.md` is where the privacy trade is argued, and where option 1 would land.

[AI-assisted - Claude]

sanity:
## Problem

A published order carried `listing_id`, and a store's listings are public, so anyone who could read a store could join its orders to its listings (#57). That exposes a per-product sales record: which lines sell, at what rate, and when a product stops selling. Counting a seller's exposure needs each order's amount, not which item it was. Decision on #57: stop publishing it.

## Approach

`Order` no longer carries `listing_id`. It carries `listing_tag`: a digest of the listing id, keyed by the buyer-seller conversation's seller-direction key (`harvest_common::mailbox::listing_tag`). Only the order's two parties hold that key, so a reader of the store cannot test a listing against the tag, and the same listing in two conversations gives unrelated tags. `None` means an invoice that answers no request. The field is omitted from the encoding when absent.

**Why a tag, not simply dropping the field.** The first revision moved the listing into the seller's encrypted acceptance message instead. Review found two problems:
- "has this request been answered" then depended on a mailbox message that can be lost or evicted, so a seller could be offered Accept again and publish a second order for one request;
- an acceptance naming no listing silently switched off the buyer's wrong-listing check.

A keyed tag on the seller's signed, published record avoids both, and still reveals nothing to anyone else.

What reads it:
- **Seller, `unanswered_requests`:** a request counts as answered when a published order carries its binding and its listing tag, recomputed with that conversation's keys. This depends only on published state, as before.
- **Buyer, `PaymentBlocker::CommitmentNotRequested`:** when the conversation asked for something, the order's tag must equal the tag of one requested listing. An untagged order is refused, not skipped.
- **Invoice, `order_for_invoice`:** writes the tag whenever the invoice answers a request. It refuses to build one if the conversation's key is missing.

The acceptance message is unchanged from main.

**What remains inferable, and is now documented.** The amount is still public, so an order whose amount matches a uniquely priced listing can still give the listing away. `order_binding` groups one pseudonymous buyer's orders within a conversation. Neither can be fixed without bucketed amounts or the ledger contract in issue 8.

## Re-keys

- **Store contract V12:** a real record change. Order ids and signature preimages change. Predecessor-generation orders still decode but no longer verify, and because expiry doesn't remove orders from state, a predecessor that ever held an order is refused permanently. That is acceptable only because no store exists. `StoreParameters` is unchanged at 56 bytes.
- **Harvest delegate V12:** codegen only. One monomorphized serializer instance is gone; the delegate uses only `OrderId`.
- **Reputation and mailbox:** unchanged.

Both V12 rows are recorded with `git show` provenance, and the migrate tests' hash lists and encoding table are extended. `cargo make code-hashes`: store ddac8dd5, delegate 08eac64a. The committed WASM matches a fresh build.

## Testing

- **Tag derivation:** `the_listing_tag_derivation_is_pinned` checks against the `b3sum --derive-key` CLI output, and `a_listing_tag_differs_by_conversation_and_by_listing` covers keying.
- **Invoice:** `an_invoice_answering_a_request_tags_its_listing_privately` covers:
  - the tag is written;
  - the listing id's bytes appear nowhere in the published order;
  - no key means no invoice;
  - an unprompted invoice is untagged.
- **Seller:** `a_request_already_answered_is_not_offered_again` and `a_request_is_answered_only_by_an_order_with_its_binding_and_tag`. The second checks that each of these still leaves the request open: another buyer's binding, another listing's tag, an unbound order, an untagged order, a tag from another conversation's key, and no keys.
- **Buyer:** `a_commitment_for_a_listing_never_requested_is_refused` and its converse, `an_untagged_order_for_a_conversation_that_asked_is_refused`, and `a_listing_tag_from_another_conversation_is_refused`.
- **Pins:** the order-id pin and the wire-compat CBOR fixture are regenerated from code, and `orders_without_a_payment_hash_still_decode` also covers an old record carrying `listing_id`.
- **Mutation checks by hand:** 6 of 7 mutations of the new logic failed a test. The 7th, building an invoice without the conversation key, never applied because formatting changed the target text. The invoice test asserts that case directly.
- **Build checks:** `cargo test --workspace`, `cargo clippy --workspace --all-targets` and wasm32 clippy are clean.

**Reviewed:** Full tier, with four freenet reviewers plus a privacy/wire lens substituted for Codex, which is out of quota. The first round's findings are addressed in b39cb2a, and a re-review follows.

Closes #57

[AI-assisted - Claude]


sanity (commented):
## PR Review: #63 (round 1, head 6596f58)

- **Tier:** Full (contract WASM re-key, message wire format, privacy).
- **Reviewers:** code-first, testing, skeptical, big-picture, plus a privacy and wire-compatibility lens substituted for Codex (still out of quota; per the updated rule, not waited for).

### Findings and disposition (fixed in b39cb2a)

**Must fix**
1. **"Answered" came to depend on a mailbox message.** (skeptical; privacy lens) Moving the listing into `OrderAccepted` meant a request was only recognised as answered when the seller's acceptance could be read back. That acceptance is sent fire-and-forget, and can be lost or evicted, including by a buyer who floods the mailbox. When it was missing, the seller was offered Accept again, which reopens the double-order hazard `a_request_already_answered_is_not_offered_again` exists for. **Fixed** by redesign. The published order now carries `listing_tag`, a digest of the listing id keyed by the conversation. Only the two parties can compute it, and "answered" rests on published state again.

**Should fix**
2. **An acceptance naming no listing switched the buyer check off.** (code-first; skeptical; privacy lens) **Fixed:** an untagged order for a conversation that asked for something is refused.
3. **Docs overclaimed.** (privacy lens; skeptical) A public amount matching a uniquely priced listing still reveals it, and `order_binding` groups a buyer's orders within a conversation. **Fixed:** both are now documented as residuals in the design doc, the untested-invariants doc and the accept panel text.
4. **The V12 store row said orders expire out of state.** (skeptical) They don't: `OrdersV1` prunes only by count. **Fixed:** the row now says a predecessor that held any order is refused permanently, which is acceptable only because no store exists.
5. **Test gaps in the acceptance design** (testing): the old-acceptance wire compat and the end-to-end listing delivery. Moot now, because the acceptance message is unchanged from main. The equivalent coverage for the tag is in b39cb2a.

**Process**
6. **#56 contradicts this change** (big-picture): it still describes the listing leak as an open gap. A note naming the passages to fix when rebasing is posted on #56.

### Verified
The re-key registry rows were recomputed against git (store `52107dcf`, delegate `4739e019`, delegate key), and no reader of a published order's listing remains.

[AI-assisted - Claude]

sanity (commented):
## PR Review: #63 (round 2, head b39cb2a, keyed listing tag)

- **Reviewers:** skeptical, testing, and a privacy/cryptography lens standing in for Codex (still out of quota).
- **Maintainer decision (@sanity):** the keyed tag (issue #57 option 2) is accepted in place of a plain removal.
- **Must-fix:** none upheld.

### Findings and disposition (fixed in 360a8f5 unless noted)

**Should fix**
1. **The tag was computed directly under the AES direction key** (crypto lens). **Fixed:** it now uses a dedicated `listing_tag_key` derived from that key, re-pinned against `b3sum`. Either party can prove which listing a signed order named by revealing the tag key, without exposing the conversation.
2. **A missing conversation key was caught only after an address index was spent** (skeptical). **Fixed:** `issue_invoice` refuses first.
3. **Residuals were incomplete** (crypto lens; skeptical). **Fixed:** all of these are now documented:
   - quantity multiples and fiat approximation;
   - tag equality within a conversation;
   - the listing claim now checkable by the two parties only;
   - fulfilment reading the request from the mailbox;
   - unprompted invoices recording no listing;
   - the buyer check needing its own request.
4. **Test gaps** (testing). **Fixed:** there is now a delegate-recalled conversation tag agreement test, and an app-state test of the real issue path, including the early refusal.
5. **Leftover `OrderAccepted { .. }` patterns, stale fixture comments, and an overconfident delegate V12 cause** (skeptical). **Fixed.**

**Dismissed**
- **"The V12 registry rows record the wrong hashes"** (testing). Not applicable. The rows record the **superseded** artifacts this PR replaces, main's store `52107dcf` and delegate `4739e019`, as `legacy/README.md` requires ("Superseded hashes only ... the current generation is ... deliberately never written down"). The round-1 big-picture reviewer recomputed both against `git show` and confirmed them. The HEAD artifacts (store `ddac8dd5`, delegate `08eac64a`) are correctly unrecorded.

### Mutation checks
The new early key check and the derived-key use each fail a test when mutated.

[AI-assisted - Claude]

sanity:
Light re-check on 360a8f5 by a separate Claude reviewer (Codex was not used, per the standing substitution rule). Nothing blocking. Two nits are fixed in ecaa805: a test's doc comment is reattached to the right test, and the accept-panel module note now says "a multiple of". One nit is dismissed. It asked for a public helper that computes a tag from the tag key alone. The derivation is two BLAKE3 derive_key calls, documented in mailbox.rs and pinned by test, and the reviewer reproduced both pins with the b3sum CLI. That is enough for a third party to check a revealed key. ecaa805 changes only a comment and a doc line in the UI crate, so no contract or delegate hash moves.

[AI-assisted - Claude]

PR #65, opened 2026-09-17
fix(ui): say what keeps an invoice's status current, now that Harvest asks the bridge

sanity:
## Problem

Two pieces of UI copy predate #61:

- The invoice lists (seller's My Store, the invoice form, and the buyer's store view) said "Harvest is not yet watching these addresses for you" and told sellers to check their own wallet.
- The manual "Watch a Bitcoin address" form said "Harvest has no route to a bridge from inside a published app".

Since #61, each invoice's address goes to the bridge through its request inbox. A live signet test on 2026-09-17 took an invoice from issue to Paid that way (test store on a throwaway peer, 10,000 sats from nova's signet wallet), so the old wording now tells sellers something false.

## Approach

The note now says what does matter: the seller's Harvest renews the watch request only while it is open, the bridge stops watching about a day after the last renewal (`RENEW_AFTER_MS` renews at 12h), and a payment made while it is not watching is not picked up later, because the bridge does not backfill (freenet-bitcoin#7). So a seller waiting to be paid should open Harvest at least once a day. The note is shown to buyers too, since they are the ones who lose out.

The manual watch form now says only invoice addresses are sent to a bridge. The `invoice_form.rs` module doc is rewritten to match; it still described the HTTP path and freenet-core#5542 as the blocker.

## Testing

Copy and doc changes only. `cargo fmt --check`, `cargo clippy -p harvest-ui --all-targets -- -D warnings`, and `cargo test -p harvest-ui` (441 passed) are clean.

[AI-assisted - Claude]


sanity:
Light review by a separate Claude reviewer (skeptical lens) on 5d2bc39. It checked every factual claim against the code: renewal only while the seller's tab is open at a 12h interval, the bridge's 24h `WATCH_LIFETIME_MS`, no backfill (freenet-bitcoin#7), buyer placement, and the manual-watch statement. All were accurate. It raised two low-severity wording issues and one nit, all fixed in 49e79cf:

- "shows Paid once the bridge sees the payment" now says "confirmed", because Paid also needs the invoice's confirmations.
- "at least once a day" is now "more than once a day". A 24h lapse with renewal due at 12h leaves no slack for opening at slightly different times each day.
- The note keeps its warning style, since a missed payment costs the buyer.

[AI-assisted - Claude]

PR #68, opened 2026-09-17
feat(ui): drop the payment-instructions field, render the store description as markdown

sanity:
## Problem

Ian, from a screenshot of the store-creation form: the third field was labelled **Payment Instructions** with the placeholder *"How should buyers pay? e.g. BTC: bc1q..., or contact me to arrange"*. Both halves are wrong. Harvest bills in on-chain Bitcoin only (`ui/src/state.rs:960`, and `payment.rs` makes Paid reachable only under `Authority::BitcoinEvidence`), and it takes a **fresh address per invoice** derived from the seller's account key, so there is nowhere for one store-wide address to live. An address pasted there invites the one payment route Harvest cannot prove: no order covers it, so no bridge is asked to watch it, nothing reaches Paid, and neither party has evidence or an order for a complaint to attach to.

Then his second call, which is what this PR actually does: a store does not need a second freeform text box beside its description at all. The field was a fossil of the era when payment really was out of band, which is still visible in `docs/design.md`'s original walkthrough ("the seller privately hands over payment instructions").

## Approach

**`payment_instructions` is removed** from `StoreInfoV1` and every UI path that carried it, leaving a store with a name and a description.

**The description is rendered as markdown**, which is what makes one field enough: headings, lists, emphasis, code, quotes and links. `ui/src/markdown.rs` parses to a small `Block`/`Inline` tree and renders real elements; it never produces an HTML string. Markdown allows raw HTML by definition, so the `dangerous_inner_html` route would put a seller's `<script>` inside the app's own origin, next to the delegate connection. Anything the tree cannot represent cannot reach the page:

- raw HTML is dropped,
- headings are demoted (`#` becomes `h4`) so a seller cannot outrank the store's own name,
- links are http/https/mailto only, and a refused link keeps its text rather than vanishing,
- an image contributes only its alt text, rather than fetching a tracking pixel from every viewer's browser.

Listing descriptions render the same way, sharing the same styles; leaving them plain would have been a split a seller notices the first time they try a heading in a listing. The form now says markdown works, since a feature nobody is told about is one nobody uses.

**This re-keys the store contract**, `ddac8dd5` to `3f47ab79`, and `legacy/store_contract.toml`'s V13 entry records what that costs. `AuthorizedStoreInfoV1::verify` re-serializes `self.info` and compares it against the signed payload (`common/src/listing.rs:264`), so a store published earlier still *decodes* but no longer *verifies*, and the contract refuses the state. A seller whose store predates this must publish their details again, which re-signs in the new shape. Acceptable only because no store but a test one has ever been created, the same trade V12 made. Parameters did not move (56 bytes), and the delegate did not re-key.

`docs/design.md` also listed payment integration as an open question answered with "payment is out of band"; it now records what was settled, and its banner no longer vouches for the out-of-band walkthrough below it.

**Separately reported by Ian, and included here:** the store's share link and an invoice's payment address were readonly inputs wearing `.form-input`, so they looked editable rather than like values to copy. They get a `.copy-field` style (monospace, plainly read-only), as does the conversation backup, where a mistyped character is worst. A copy button is deliberately not the answer: `bitcoin_view.rs` records that the clipboard API is unreliable in the gateway's sandboxed iframe, and a button that silently does nothing is worse than a field that visibly works.

## Testing

`cargo test --workspace`: 713 passed. `cargo check --manifest-path tests/rehearsal/Cargo.toml`, clippy `-D warnings` and fmt clean. `cargo make code-hashes` accepts the new hash and reports store parameters still 56 bytes.

New tests in `ui/src/markdown.rs` pin the parts that matter:

- raw HTML a seller typed never reaches the tree, over four shapes including ``,
- a `javascript:` or `data:` link is not clickable but keeps its text, and `safe_href` refuses mixed case, an embedded newline, relative and protocol-relative URLs,
- headings are demoted,
- an image contributes only alt text,
- a fenced block is not parsed as markdown,
- lists, ordered lists, emphasis, inline code, quotes and rules survive,
- plain text is left alone, wrapped lines read as one paragraph.

In `common/src/store.rs`: store info written while the field existed still decodes, and the re-encoding differs by exactly the removed key and nothing else (so the break is the decided one, not a second one hiding behind it).

[AI-assisted - Claude]


sanity:
Light review by a separate Claude reviewer (skeptical lens). It verified the claims the new copy makes, and they hold: on-chain only (`payment_hash` is `None` at all twelve construction sites, and `payment.rs:1490` makes Paid reachable only under `Authority::BitcoinEvidence`, so there is no seller-say-so route), a fresh address per invoice (`bip32.rs:34,188-211`, with `next_index` advanced on hand-out), `payment_instructions` display-only and absent from the store contract, and the watch-request claim correct. It also confirmed the new line does not overstate: a Gift listing cannot be invoiced at all (`state.rs:3843` rejects a zero amount), so there is no order that could be shown as paid.

Two findings, both addressed:

**Medium, fixed in the follow-up commit.** Answering the Open Questions entry left four passages in the same file saying the opposite: the overview's "pay using whatever method the seller specifies", and steps 3, 5 and 6 of the walkthrough. Worse, the banner scoped supersession to the feedback tokens and explicitly vouched for the rest, so the file certified the out-of-band account while denying it later. The banner now covers payment and points at where the real account is, the overview says what the app does, and the walkthrough is marked as the original design with its payment steps named. The steps are left as written, because they are history and rewriting them would erase why the bridge exists.

**Low, deliberately not fixed, and the reviewer's reasoning was wrong on this one.** `common/src/store.rs:80` still says `/// Payment instructions (freeform, e.g. "BTC: bc1q...")`. The review argued the doc comment cannot re-key anything because the WASM is a committed artifact. I tried it: adding the replacement comment moved the store contract hash from `ddac8dd5…` to `762977ec…`, because it shifts panic locations, and CI rebuilds and compares ("committed contract wasm matches a fresh build"). So fixing that comment means a re-key, a legacy registry entry and the acknowledgement label, which does not belong in a copy-only change. It should ride along the next change that already re-keys the store contract.

Also took "SETTLED" down to "ANSWERED", since freenet-bitcoin#7 is still open.

[AI-assisted - Claude]

sanity:
Big-picture review of 62c8c65 by a separate Claude reviewer. It caught a CI failure before CI did, and three real gaps. All fixed in 07fb414, which is green:

- **`tests/rehearsal/src/main.rs` still set the removed field.** It is its own cargo workspace, so `cargo test --workspace` never compiles it, and CI checks it as a separate required job. That is the second time today a separate workspace has hidden a break from a local test run.
- **Listing descriptions were left plain** while the store description directly above them rendered markdown. A seller would find that split the first time they tried a heading in a listing, so listings render the same way now, sharing the styles.
- **The messaging fallback still said "the store's notes"**, naming a field that no longer exists; it says "description".
- **Nothing told a seller markdown works.** There is a line under the description now. A feature nobody is told about does not earn the care in `markdown.rs`.

It also found the PR body describing the earlier, smaller version of this change, including the claim that nothing re-keys, which is false for this head. The body is rewritten.

One thing it raised that I did not change: it confirmed `legacy/store_contract.toml`'s V13 entry and the migrate-test rows are consistent with the diff, and that `docs/design/migratability.md` is a process doc rather than a per-generation log, so it needs nothing.

A security-lens review of the renderer is still running; I will post it when it lands.

[AI-assisted - Claude]

sanity:
## Security review of the markdown work

A separate Claude reviewer attacked `ui/src/markdown.rs` directly: 400,000 fuzz iterations over a hostile alphabet, measured nesting depths, and a read of freenet-core's shell and navigation interceptor to see what a link can actually do. It found one Must Fix and several real hardening gaps. Everything below is fixed in f22fc74 and 099e0b8.

### Must Fix

**Nothing capped nesting depth, and both the renderer and the tree's own `Drop` recurse per level.** Measured: `">".repeat(50_000)` parsed in 12ms into a quote tree 50,001 levels deep, and pulldown-cmark has no block-nesting limit. wasm links with a 1 MiB stack, where an overflow is an unrecoverable trap rather than a catchable panic, so a hostile description would kill a visitor's whole Harvest app, not just the description. The text lives in contract state, so a reload re-reads it and dies again: a persistent denial of service from a store link, and after listing descriptions render too, from one listing inside an otherwise ordinary store. `Drop` recursing is why a render-time guard would not have been enough.

`parse` now caps the source, flattens containers past `MAX_DEPTH`, stops at a node budget, and says so in the output rather than trimming in silence. Runs of text are coalesced too: the same review measured `"[".repeat(50_000)` producing 50,001 inline nodes, since the parser emits an event per unmatched punctuation character, so a megabyte was about a million DOM nodes. Both are needed; neither covers the other. Tested at 200,000 characters of each shape.

### Also fixed

- **An HTML block swallowed the whole description.** In CommonMark a block runs to the next blank line and every line arrives as raw HTML, so a stray `<div>` at the top dropped everything, with no preview for the seller to notice in and nothing said to the buyer. Raw HTML is kept as literal text now: it cannot execute, so the safety property is unchanged, and the module doc no longer claims nothing disappears while that was false.
- **A link to the machine the app runs on is refused.** Relative links already were, on the grounds that they resolve inside the app's own origin; an absolute URL naming that origin did the same thing and was allowed. It matters because the shell carries `allow-popups-to-escape-sandbox` and the interceptor returns early on a non-`_self` target, so a `target="_blank"` link opens a real top-level tab at the node's origin: a one-click load of any contract webapp, a look-alike Harvest included. Loopback, private, link-local and `.local` hosts are refused; the residual (a node on a public hostname) is stated in the function's doc rather than implied away.
- **A link's text could claim a different address than its target.** Plain text made a URL inert; a link does not, on the screen where somebody decides whether to pay a stranger. A link whose text looks like an address that disagrees with the href now names the real destination inside the link.
- **The re-encode test had been weakened to a length check**, which passes a length-preserving change to the signature preimage: `serde_bytes` on `reputation_contract_id` turns `array(32)` into `bytes(32)`, same two bytes, every ghostkey signature dead. That field's own doc flags it as the change somebody will try. Back to byte-exact.
- **The payment address is a textarea**, so a buyer sees all of it. At this width 42-62 characters of bech32 scrolled sideways, and a partial selection pastes a truncated address, which sends coin where no order can recognise it.
- **Long unbroken strings wrap** instead of pushing the certificate warning off screen, **each store's link carries its own aria-label** (it replaces the visible label rather than adding to it, so two stores sounded identical), and the decode fixture is the shape a live store has.
- **The renderer had no test at all**, so an edit adding a raw-HTML sink or dropping `rel` would have passed every test in the file. It has a source pin now. Naming the attribute in that test's own doc comment made its count read two on the first run, which is the self-matching-grep trap in miniature and is recorded there.

### Dismissed, with reasons

- **No contract-side cap on description length.** The UI caps what it renders, which protects this client; a cap in the store contract is a re-key and a decision about what a store may hold, not a rendering fix. Worth doing deliberately rather than bundled here.
- **Seller content is not visually contained** (a border or background to separate it from the app's chrome). Text-level spoofing of the app's voice was possible before this change, and the containment is a design question for Ian rather than a security fix.
- **Bidi control characters pass through.** Pre-existing for all seller text, not introduced here, and the same class as the point above.
- Noted for the record, out of scope: `ui/src/document_title.rs` posts the seller-controlled store name to `window.parent` with `targetOrigin: "*"`, untouched by this change.

`cargo test --workspace`: 719 passed. clippy `-D warnings`, fmt, and the rehearsal harness check all clean.

[AI-assisted - Claude]

sanity:
## Re-review of the hardening round

Same reviewer, fresh extraction of 099e0b8, ~700k fuzz iterations plus an instrumented build carrying a shadow stack over the depth accounting. One blocking finding, and it was introduced by the previous round's fix. All fixed in ba1e531.

### Blocking, and mine

**Hitting the node budget inside a list or quote threw the container away.** The loop broke out with containers still open, and both their contents and the truncation marker go to the innermost open container while `finish` returns the outermost. Measured: `"> a\n>\n".repeat(2048)` rendered **nothing at all**, and `"- a\n".repeat(2048)` rendered the marker alone. That is the same silent loss this module had just fixed for HTML blocks, reintroduced by the fix for the denial of service, and the module doc claimed the opposite.

Content pushes respect the budget now, structural ones do not (dropping a quote drops everything it holds), the event stream is allowed to close what it opened, and the marker goes to the document rather than to whatever was open. Tested at 4096 repetitions of three shapes, asserting both that content survives and that the reader is told.

### The test that could not see what it tested

`parse(&">".repeat(50_000))` was the input the whole depth argument rested on, and it passes with `MAX_DEPTH` deleted: empty quotes are dropped by `End(BlockQuote)` either way, so the tree is depth 1 in both builds. With a trailing character it is 4001 without the cap and 17 with it. Fixed, and worth recording as the shape to watch for: the assertion was true, stable, and about nothing.

### Host parsing was bypassable by one extra slash

A browser skips redundant slashes and backslashes after a special scheme, so `http:////127.0.0.1:50509/…` is the loopback address and was **allowed**. So were `http://2130706433/`, `http://0x7f000001/`, `http://0177.0.0.1/`, several spellings of IPv6 loopback, `http://localhost./`, and fullwidth characters that map to ASCII. The function's doc had claimed its only residual was a node on a public hostname.

Now: an http(s) URL whose authority cannot be read fails closed, addresses are parsed rather than prefix-matched (including the decimal, octal and hex forms), and a non-ASCII host is refused rather than compared, which also stops the destination note echoing a homograph back as reassurance.

### The destination note had four silent evasions

Two asterisks (it read only top-level text), a bare domain (how most people write link text), userinfo (`[https://freenet.org@evil.example]` — reading the host the way a browser does is exactly what hides the spoof), and a `mailto:` target, which has no host. All four say something now.

### The app's own voice

The marker and the note were plain text a seller could reproduce exactly, which trains a buyer on a signal the attacker controls. Both are their own node with their own style now, and a test checks that a seller typing the marker does not get it.

### Also

The store-details warning still promised a field that no longer exists; the payment address tracks re-renders through `value` rather than an initial text child (the reviewer asked whether the update path had been seen, and it had not); and the re-encode test works its byte count out rather than stating it in a comment.

### Dismissed, with reasons

- **No page-level node cap** across a store's many listings. The per-description cap plus the store contract's own limits are what this change can reasonably carry; a cap on how many listings a store may hold is a contract decision, and the same one deferred last round.

`cargo test --workspace`: 722 passed. clippy `-D warnings`, fmt, and the rehearsal check clean.

[AI-assisted - Claude]

sanity:
## Rounds 3 and 4

Round 3 attacked the host logic against a real WHATWG parser (Node's `URL`), round 4 checked the fixes and, importantly, the false-refusal direction. Both found something real, and in both cases the defect was in code the previous round had just added.

### Round 3, blocking: a percent-encoded host reached loopback

A browser percent-decodes the host **before** parsing it, so `http://%31%32%37.0.0.1/` is 127.0.0.1 and `http://loc%61lhost/` is localhost, and comparing the undecoded text sees neither. The same trick defeated the non-ASCII guard, since the URL text stays ASCII while the decoded host is fullwidth. That was the loopback hole round 2 closed, reachable through a different encoding, and the doc claimed it was covered.

Decoding here would mean reimplementing the URL host parser, including the domain-to-ASCII mapping, so a host that is not plainly letters, digits, dots, hyphens (or a bracketed IPv6 literal) is simply not linked. That also closed a second finding: `host_of` split on punctuation but not spaces, so `[a.b](<https://Verified seller. Payment is safe.>)` made "the host" a sentence, which then rendered in the app's own muted note style. A seller could word the app's voice.

Also from round 3: IPv6 unique-local and link-local were allowed while their v4 equivalents were refused; the note fired on any dotted text (`v1.2.3`, `$4.99`, `Fig.1`), which teaches a buyer to ignore it, and fell silent when the address sat inside a sentence or used a fullwidth dot; and the truncation marker claimed a cut when the depth cap had only flattened nesting, where every character survives.

### Round 4, blocking: my fix silenced the two most natural spoofs

The claim token was tested **before** its punctuation was trimmed, so a trailing full stop or quote landed in the last label and the token was never considered. Measured against the previous head: `freenet.org.`, `"freenet.org"`, `(freenet.org)` and three more went from firing to silent. Trimming first restores them, and the new test is mutation-checked against the old ordering.

Two more it found in the same code: the doc claimed the last-label rule excluded `readme.md`, which it did not (`.md` is two letters, and Moldova), so filenames are now excluded explicitly; and the note fired on an honest `mailto:` link in text identical to the spoofed one, which makes it unactionable either way.

### What round 4 cleared

It ran 38,500 differential cases against Node's parser and found **no** URL that Harvest allows and a browser resolves to loopback, RFC1918, link-local or `.local`, and confirmed the host guard is strictly host-scoped: percent-encoding, spaces and non-ASCII in a path, query, fragment or userinfo are all still allowed, as are punycode, ports, trailing dots, long hosts and uppercase. Underscore hosts were a genuine false refusal and are now allowed. It also re-measured the caps (worst case about 5,300 nodes in 1.1ms at the source cap) and confirmed the `suppressed` counter is correctly LIFO-paired.

### Left as they are, with reasons

- **A refused link degrades to bare text with no marker**, so a false refusal (an IDN, say) loses the address silently. Worth a note eventually; it is a behaviour question rather than a defect in this change, and the refusals are deliberate.
- **Both caps can fire at once**, and the cut message wins over the flatten message. Neither is false; saying both would be noise.
- **`flattened` can be set for a container that held nothing** (`">".repeat(40)` alone), so the note appears about content that never existed. Harmless and rare.
- **A `mailto:` note cannot distinguish two mail addresses.** Now suppressed for the honest case, which is the half that was misleading.

`cargo test --workspace`: 728 passed, 28 of them in the markdown module. clippy `-D warnings`, fmt and the rehearsal check clean. CI green on e2fdabe.

Merging on this. Four rounds, and three of the four blocking findings were introduced by the previous round's fix, which is worth recording: the failure mode of this module is not the parser, it is each new guard's edges.

[AI-assisted - Claude]

PR #73, opened 2026-09-18
fix(ui): keep asking for an unsettled order's payment address

### LINKED ISSUE #67: an order can never recover from a stale address-contract read, so a confirmed payment reads as awaiting payment forever
Found in a live signet test today (2026-09-17). Two invoices, same seller, same vpub. The first reached Paid. The second (order 5yxYeT, 7,000 sats, `tb1qr9hesff3e0lljr7r30qzx0ljjlmxu4lvg7lv6d`) was confirmed in block 322547 and the bridge published its claims seconds later, and its own end-to-end check agrees the proof is sufficient:

```
--verify tb1qr9hes… --network signet --prove-payment-of 7000 --confirmations 1
YES -- settled.
1 signed claim(s) constitute the proof; an order carrying them
would transition AwaitingPayment -> Paid
```

The UI still showed AWAITING PAYMENT twenty minutes later, and a reload did not help.

## Root cause is in the node, but Harvest cannot escape it

The app's peer had fetched that address contract five minutes before the payment, then missed the UPDATE, and answered every later GET from its 507-byte cached copy while three other peers held the full 4846 bytes. Filed as freenet-core#5703.

What makes it permanent for a user is that Harvest reads each address contract once and never questions the answer:

- `ui/src/state.rs:3102` guards the GET on `bitcoin.subscribed`, which is insert-only (`ui/src/state.rs:5851`) and never cleared, so a tab asks once.
- `ui/src/gateway/response_handler.rs:104-128` does not route `NotFound` for an address contract anywhere, so a miss leaves no trace and no retry.
- Settlement is attempted only when address or store state arrives (`ui/src/state.rs:1841`, `:1945`), never on a timer, and `settled_orders` returns early when the address has no view (`ui/src/state.rs:2937-2943`).

So there is no staleness check and no periodic re-read. The app's own reasoning in `common/src/payment.rs` says a client should cross-check the live address contract; today it checks once and believes the answer forever.

## What would fix it here

Re-read an order's address contract on a timer while the order is unsettled, rather than only when the node volunteers state, and treat a `NotFound` or an unchanged-but-stale read as a reason to ask again. That also covers the ordinary case of an address contract that does not exist yet when the invoice is issued. freenet-core#5703 asks for a way to demand a network read; even without it, re-asking is what gives the node a chance to have healed.

Workaround today: open Harvest against a peer that has not already cached that address contract.

Found while running the end-to-end test that took the first order to Paid, so the happy path does work; this is the recovery path.

[AI-assisted - Claude]

sanity:
## Problem

Closes #67.

Harvest learns that an invoice was paid by reading the Bitcoin address contract the order names. It asked for that contract **once**, when the order first appeared, and then waited for the node to volunteer every later version.

That is enough while the node keeps up, and the node does not always keep up. A peer that misses one live fan-out serves its own copy until anti-entropy repairs it. In the live signet test on 2026-09-17 an order that had been paid, confirmed in a block, and attested by the bridge sat on AWAITING PAYMENT for the rest of the session. A reload did not help, because a reload asks once too.

The node-side latency is a known residual (freenet-core#5527: 300s minimum, 40-75 minutes in the field; the peer in that test converged after about 18 minutes, and freenet-core#5703 was closed because it does converge). This change does not touch it. It makes Harvest recover from it, which is the half that is ours.

## Approach

**Keep asking while an order is unsettled.** `ui/src/address_reread.rs` is the decision half, testable without a browser: an address never asked about is due at once, each consecutive ask doubles the wait to a five-minute ceiling, and each address carries an offset derived from its own id so a seller's invoices do not all fire on one tick. Arriving state does not reset the spacing -- a stale copy answers every ask, so resetting on arrival would hold the fastest cadence for exactly the case the widening exists to bound.

The timer decides under a **read** of the state and takes it for writing only when something is due. Taking it for writing re-renders every component, and most ticks have nothing to send.

What bounds it is the caller: only addresses of orders that are ours, still `AwaitingPayment`, and still inside the payable anchor window. That window is now one shared `worth_watching_for_payment` helper rather than a second copy of the rule `watches_wanted` already applied. **The bound needs a chain tip** -- without one this node cannot say the window has closed, so it keeps asking; that is deliberate, and said plainly in the module doc.

Three things came out of review and are part of the fix rather than extras, because without them the re-read cannot deliver a settled order:

- **A stale answer must not be able to make a settlement *wrong*.** `apply_address_state` replaces the view wholesale and a reorg retraction is an *additional* claim, so applying an older copy would un-retract a reversed payment. A copy that scanned less far than the one already held is now refused. This narrows the window rather than closing it: an equal watermark is still admitted, which is what a same-height reorg produces.
- **A settlement update that failed to send is published again.** The submitted-set was recorded before the send and never withdrawn, so one failed dispatch stranded a confirmed payment for the session -- #67's own symptom, one step further down the path.
- **A buyer's tab no longer attempts a publish.** Only the seller can publish (the contract takes an update signed with the store's key), so on a buyer's tab the send always failed and apologised. With re-reads arriving every few minutes that apology would have piled up on the screen of the buyer whose payment worked.

## Testing

24 tests, all native, plus a mutation matrix. Every guard was verified by making it **fail**, not by watching it pass: deleting the unsettled/payable filter, the spacing, the id registration, the send loop, the staleness guard (both directions -- too loose *and* refuse-everything), the ownership gate, the withdrawal, the stagger, the backwards-clock branch, and narrowing the watch set to the re-read set each turn a specific test red.

Three of those mutations survived CI before this PR added the tests that catch them. Two bugs were found by writing the tests: the ask counter is 1-based so shifting by it doubled every wait, and the first stagger spread addresses over one tick of the timer that reads them, so 255 ids in 256 convoyed anyway.

An arriving `BitcoinAddressStateV1` is driven through `on_contract_state`, so the routing is tested rather than the map entry.

Full workspace suite green (751 tests), `cargo fmt` and `cargo clippy -D warnings` clean.

## What this does not do

It does not make a stale answer fresh, and cannot: a read is served from the node's local copy, and `hosting-invariants.md` invariant 1 settles that there is no freshness flag to ask for. It shortens the time between the node healing and the screen saying so, from "never, until the tab is reloaded at the right moment" to a few minutes.

It also does not re-read the **tip** contract, which has the same read-once shape. Deliberate: a tip contract gets an update per block, so its own stream heals a missed fan-out within minutes, whereas an address contract may receive one update in its entire life. That asymmetry is why the address contract is the pathological case.

Known residuals, all recorded in `docs/untested-invariants.md` rather than implied away: the equal-watermark case above; the watermark being a total order only while one bridge is trusted; and the interval registration itself, which no test can reach because that file is type-checked and never executed.

[AI-assisted - Claude]


sanity (commented):
## Comprehensive PR Review: #73

### Summary
- **PR Title:** fix(ui): keep asking for an unsettled order's payment address
- **Type:** fix
- **CI Status:** green on `8c803c6` (the reviewed SHA); fixes pushed as `b43e051`, re-review below
- **Linked Issues:** #67
- **Review tier:** **Full** -- the change gates when an order is shown as Paid, and adds periodic network requests
- **Reviewers run:** `code-first`, `testing`, `skeptical`, `big-picture`, plus two substituted lenses (browser-runtime, end-to-end efficacy)

### External model substitution (recorded, per `~/.claude/rules/multi-model-review.md`)

`codex review --base origin/main` failed twice with `You've hit your usage limit ... try again at Sep 19th, 2026 10:34 AM`. `gemini` failed with `IneligibleTierError: This client is no longer supported for Gemini Code Assist for individuals`. Both retried once, then treated as down; the review was **not** held for the reset. Two extra Claude lenses with distinct adversarial framings were substituted: **browser-runtime** (the wasm/Dioxus paths no native test reaches) and **end-to-end efficacy** (tracing "node heals" to "screen says Paid").

---

### What the reviewers confirmed works

The end-to-end chain is complete and reaches the party who was actually stuck. The re-read registers the id before asking, so the answer routes to `apply_address_state`; nothing upstream in `response_handler` intercepts it (the migration-probe and bridge-pointer interceptors key on maps an address id is never in); `publish_settled_orders` then runs for every browsing store; and the seller's own store is in `browsing_stores` via the same store-state path that feeds My Store. Only the seller can publish a settlement at all (`submit_order_by_id` requires the store to be theirs), so this had to work on the seller's tab, and it does.

Better than the PR claimed: `OrderCard` renders "Payment seen on chain" from the live address view, so the pill flips on the response itself rather than waiting for a store-contract round-trip.

Also verified sound: the `spacing` arithmetic (clamp-before-shift, no overflow, and the off-by-one genuinely fixed); `worth_watching_for_payment` is byte-for-byte the filter it replaced; no borrow panic from the spawned GETs (the browser `WebApi::send` has no await points, so no guard is held across a suspension); one interval per page, not one per reconnect; and the scoping means no stranger's payment address is ever asked for.

### Must Fix -- all addressed in `b43e051`

1. **A stale answer could make a settlement wrong, not just late.** `apply_address_state` replaces the view wholesale, and a reorg retraction is an *additional* signed claim, so applying a copy from before the retraction un-retracts a reversed payment -- with `publish_settled_orders` on the next line. This PR is what makes applying possibly-stale state routine, so the guard belongs to it. **Fixed:** a copy that scanned less far than the one already held is refused, pinned by `an_answer_that_scanned_less_far_is_ignored` (mutated red).
2. **A failed settlement publish could never be retried.** `settlements_submitted` was recorded before dispatch and never withdrawn, so a failed send left a confirmed payment reading as awaiting payment for the session -- #67's own shape, one step further down the path, where the re-read cannot reach it. **Fixed:** withdrawn on dispatch failure.
3. **`watches_wanted` lost its doc comment** to the extracted helper via a missing blank line, leaving a `pub fn` undocumented and a boolean helper carrying a description of grouped payment scripts. The repo has written this trap up three times already. **Fixed:** helper moved below, with its own doc.
4. **The tick decided under a write**, re-rendering every component every 60s for the life of any unsettled order (new for buyers, who previously had no periodic write at all), and ran the expensive walk -- conversation AEAD decrypts plus a signature verify per purchase -- twice per tick. **Fixed:** decided under a read, written only when something is due, walked once.
5. **An ask that was never sent burned a doubling step**, so a disconnected tab reached the 5-minute ceiling having sent nothing and was at its slowest on reconnect. **Fixed:** withdrawn on send failure, matching `spawn_inbox_entry_signature`.

### Should Fix -- addressed

6. **`subscribe: true` on every re-read.** Core documents the per-client interest registration as non-idempotent. **Fixed:** re-reads now read without re-subscribing; the subscription is still made once when the order appears.
7. **A backwards wall-clock step silently disabled the feature** until real time caught up. **Fixed:** treated as due.
8. **No jitter:** every address of a batch was first asked on one tick and stayed in lockstep forever. **Fixed:** a deterministic per-address stagger from the id.
9. **`default_network()` where the order's own network was in hand.** **Fixed.**
10. **Three mutations survived CI** (measured, not read): deleting the send loop, narrowing `address_contracts_to_watch` to the re-read filter, and -- after it was added -- the staleness guard. **Fixed:** all three now fail, and an arriving `BitcoinAddressStateV1` is driven through `on_contract_state` so the *routing* is tested rather than the map entry.
11. **Doc overclaims.** "Nothing announces the repair" was an inference stated as fact; the tens-of-minutes window was attributed to serve-DURING when that is a different clause; and the payable-window bound silently requires a chain tip. **Fixed:** all three say what is observed and what is inferred.

### Consider -- deliberately not done, with reasons

- **The tip contract has the same read-once shape** and is not re-read. Left alone deliberately: a tip contract gets an update per block, so its own stream heals a missed fan-out within ~10 minutes, whereas an address contract may receive one update in its entire life. That asymmetry is why the address contract is the pathological case; it is now stated in the module doc. Worth a follow-up issue, not this PR.
- **`retain` only prunes on a tick that has something to send**, so the last settled order's entry is pruned later rather than at once, and an order that flaps out of a stale store state and back is asked at the base wait again. Both cost a few dozen bytes or one GET. Documented rather than defended against.
- **With no chain tip there is no bound at all.** Deliberate -- the alternative stops watching an order that is still being paid -- and now said plainly in both the module doc and the helper's.
- **The interval registration itself is still unexercised**: deleting the whole block leaves the suite green, because that file is type-checked and never executed. Recorded as its own row in `docs/untested-invariants.md` rather than papered over, and the tick rate is now pinned against `FIRST_RETRY_MS` by a `const` assertion.

### Verdict
**State:** Needs Changes -- Re-review Required After Fix (findings 1-5 were blocking)
**HEAD SHA reviewed:** `8c803c6`

All eleven findings are addressed in `b43e051`. That is well past the re-review threshold (blocking findings, and far more than 30 lines), so a fresh pass on the new head follows this comment.

[AI-assisted - Claude]

sanity (commented):
## Re-review: #73 at `7cd964e`

Four further review rounds since the first report. Recording them here because the interesting part is not the findings but where they came from: **most of the blocking findings in each round were introduced by the previous round's fix**, which is the argument for re-reviewing a fix rather than trusting it.

**Reviewers run:** `code-first`, `testing`, `skeptical`, `big-picture` across rounds, plus substituted browser-runtime and end-to-end lenses. `codex` remained out of quota (reset Sep 19) and `gemini` remains retired (`IneligibleTierError`); retried once each, then substituted, never waited for.

### Round 2 fixes were reviewed, and three of them were wrong

- **The `settlements_submitted` withdrawal apologised to buyers on a loop.** Only a tab owning the store can publish, so on a buyer's tab the send always fails; withdrawing on failure turned that permanent failure into a retry on every address arrival, each pushing "your payment was seen on chain, but the order could not be updated to say so" onto a list nothing clears. Fixed by not attempting a publish that cannot succeed.
- **The stagger staggered nothing.** It spread addresses across the first wait, which is exactly one tick of the timer reading it, so 255 ids in 256 rounded to the same tick. The test passed because it sampled at arbitrary milliseconds rather than at ticks.
- **The staleness guard's accepting direction was untested.** Mutating `>=` to `>`, or making it refuse everything once a view was held, both left the suite green -- and both recreate #67 inside the guard written to make #67 safe.

### Then the guard itself turned out to be the wrong mechanism

A watermark comparison cannot order two copies of one scan window: the bridge stamps its watermark only when the height advances, so a copy missing a claim compares **equal**. Review demonstrated both directions by probe against the code as it stood:

- paid copy, then an equal-watermark copy with no payment: **50,000 confirmed sats erased, order un-settled**
- retracted copy, then an equal-watermark copy without the retraction: **a reversed payment published as `Paid`**

Claims are signed and additive, so a copy that is behind differs from a current one only by absence. Folding them needs no ordering at all, which is the property the watermark could not supply. That replaced the guard.

### And the fold had a hole of its own

`from_claims` fails on the first claim that does not verify, and the error path used the **arriving** state -- so one piece of junk, signed by nobody, collapsed the fold back to replace and restored the erasure in full. Reproduced by probe. Claims are now filtered before folding, and a fold that fails anyway keeps the view as it stands: refusing an update costs a late settlement, replacing the view costs a payment.

### Where it stands

- **24 tests**, every guard verified by making it fail. Mutations that CI survived before this PR and now do not: deleting the send loop, narrowing the watch set to the re-read set, the staleness guard in both directions, the ownership gate, the withdrawal, the stagger, the backwards-clock branch, and the fold.
- **Two bugs were found by writing tests**: the ask counter is 1-based, so shifting by it doubled every wait; and the first stagger was defeated by tick granularity.
- **Two follow-ups filed** rather than left in this thread: #74 (the tip contract has the same read-once shape and gates every settlement) and #75 (`Paid` is authorized by evidence and the contract accepts it from anyone, but `owned_store_key` resolves the key from `my_stores`, so a buyer's publish has never succeeded -- which is why an order currently reaches `Paid` only when the seller opens the app).
- **Residuals are recorded in `docs/untested-invariants.md`**, including the honest one: the interval registration itself is unreachable by any test, because that file is type-checked and never executed.

### Verdict
**State:** Ready to Merge
**HEAD SHA reviewed:** `7cd964e`, CI green on that SHA (all five checks).

[AI-assisted - Claude]

PR #80, opened 2026-09-18
fix(ui): make "Publish details" actually publish when only the key is missing

### LINKED ISSUE #78: "Publish details" does nothing, so a store keeps telling buyers they cannot message the seller
Found while re-taking screenshots on a clean peer, 2026-09-17.

## What happens

A store created through the normal flow shows this on My Store:

> This store publishes no encryption key, so buyers who open it are told they cannot message you. Publishing the details below adds the key your delegate already holds; nothing else about the store changes.
>
> [Publish details]

Clicking **Publish details** does nothing observable:

- no vault prompt appears (`#__freenet_perm_overlay .fn-card` count is 0 after 25s)
- no error in the browser console, and no notification in the app
- the notice is still there afterwards, and is still there on reload

Reproduced three times on a fresh node (port 7651), fresh browser profile, and a freshly created store, on the build published this evening.

## Why it matters

It is the only thing the UI offers a seller to fix a state it has told them is a problem, and it is phrased as a live defect in their store. A seller who clicks it and sees nothing happen has no next move and no reason to believe anything is wrong with the app rather than with their store.

Worth noting the notice may itself be wrong: the buyer's view of the same store rendered "Contact Seller" normally in the same session, so it is possible the key is published and the notice is stale. Either way one of the two is lying to somebody.

## Where to look

Whatever handles that button in `ui/src/components/my_store.rs`, and whether it reaches a signature request at all -- the absence of a vault prompt suggests it returns before asking the delegate to sign, since everything else the seller signs does prompt.

Two possibilities worth separating first:

1. The button's handler is not wired, or returns early on a condition that is always true for a store created this way.
2. It does publish, and the notice's condition is wrong or reads state that is never refreshed.

The second would make this a display bug rather than a broken action, and the fix is different.

[AI-assisted - Claude]

sanity:
## Problem

On My Store, a store with no published encryption key shows a repair notice ("This store publishes no encryption key, so buyers who open it are told they cannot message you...") with a "Publish details" button. Clicking it did nothing observable: no vault prompt, no error, no notification, and the notice persisted after reload.

## Root cause

Neither of the two hypotheses in #78. Instead: the button labelled "Publish details" only ever toggled open an edit/repair form (`ui/src/components/my_store.rs`, the button's `onclick` at what is now line ~349-368 on `main`); the actual publish happens on a *second*, undisclosed "Publish" button inside that form (`StoreDetailsForm`'s `on_submit`). For every other gap this is reasonable, since the seller has to type something. But for `StoreDetailsGap::NoEncryptionKey` specifically, the gap's own message (`state.rs:676-680`) promises publishing "adds the key your delegate already holds; nothing else about the store changes" -- there is nothing for the seller to fill in. A single click that only opens a form with nothing new to type in it is indistinguishable from the button being broken: no network call is made, so no vault prompt and no notification, and the notice is unchanged.

The notice's own condition (`store_details_gap` in `state.rs`) is correct, not stale: a normally-created store legitimately publishes without its encryption key whenever the async `InitEncryptionKey` delegate round-trip loses the race against the initial store PUT (`ui/src/components/my_store.rs`'s `initiate_store_creation`, step 3, whose own comment already anticipates this: "a store that publishes without it is a store buyers cannot message, which `store_details_gap` reports and re-publishing repairs"). The buyer-side "Contact Seller" button the issue cites as counter-evidence (`store_view.rs`) is unconditional -- it toggles the messaging panel regardless of whether the seller published a key -- so it renders identically whether or not the key is actually present, and isn't evidence the key was published.

## Fix

Extracted `store_details_button_action(gap: Option<StoreDetailsGap>) -> StoreDetailsAction` (`PublishNow` | `ToggleForm`). It returns `PublishNow` only for `NoEncryptionKey`, since that's the one gap where the details already on record are sufficient to republish as-is. Every other gap (`NeverPublished`, `NoName`, `NoReputationLink`) and the ordinary "Edit details" click (no gap) still open the form, since those genuinely need seller input.

For the `PublishNow` case the button now calls `publish_store_details` directly with the current details, which always leaves a notification (success or a specific error) via the existing error handling in `publish_store_details` -- no silent no-op on any path.

## Testing

Added `store_details_button_tests` in `ui/src/components/my_store.rs`:
- `no_encryption_key_gap_publishes_immediately` -- the regression test.
- `gaps_needing_seller_input_open_the_form`
- `no_gap_opens_the_edit_form`

Verified the regression test fails without the fix by temporarily reverting `store_details_button_action` to always return `ToggleForm` (matching the pre-fix behavior) -- `no_encryption_key_gap_publishes_immediately` failed with `left: ToggleForm, right: PublishNow`; restored the fix and it passes again.

`cargo test --workspace` (499 UI tests + contract/delegate crates, all green), `cargo clippy --workspace --all-targets` clean, `cargo fmt --check` clean.

Fixes #78.

[AI-assisted - Claude]


sanity:
## Review (1 adversarial Claude lens; Light tier — small, self-contained, no high-risk surface)

An independent reviewer (blind to my diagnosis, given only the diff + surrounding code) checked the gap-matching, staleness, notification guarantee, and test quality. Summary:

- **No blocking bugs.** Gap-matching in `store_details_button_action` is correct (`NoEncryptionKey` is only reachable after `NeverPublished`/`NoName`/`NoReputationLink` are ruled out, so `card.details` is always valid non-default at the moment `PublishNow` fires — no risk of publishing empty details). `card.details`/`card.gap` are rebuilt fresh from `APP_STATE` on every render, so no staleness. `publish_store_details` notifies on both `Ok` and `Err` — no silent no-op on this path.
- **Non-blocking finding, accepted as a tradeoff:** while a store's only gap is `NoEncryptionKey`, the button no longer opens the edit form, so a seller who wants to edit the store name/description *at the same time* as fixing the key can't do both in one action — they'd publish the key fix first, then use "Edit details" afterward once the gap clears. Not fixing this: the notice's own text says publishing "adds the key... nothing else about the store changes," so this button is deliberately a pure one-click repair rather than an edit surface, and the combined-edit path was never a documented feature to begin with (previously it required opening the form and clicking Publish inside it — same number of steps, just in one modal). Editing remains fully reachable via "Edit details" once the gap resolves.
- **Nit, not fixed:** `store_details_button_action(card.gap)` is called twice per render (once for the closure, once for the label). Harmless (pure, cheap), left as-is for readability.
- Test coverage confirmed adequate for the fixed logic (pins the exact regression + the "every other gap still opens the form" invariant), though it's scoped to the pure decision function rather than full component rendering, which is consistent with the rest of this file's test style (no DOM-level tests exist elsewhere in this codebase either).

[AI-assisted - Claude]

sanity:
## Addressed: independent skeptical review findings (commit 84f07d0)

1. **Fixed — stale-details race.** `store_details_button_action` ignored whether this store's edit form was already open. Sequence: gap `None`, seller clicks "Edit details" (form opens), `EncryptionKeyReady` arrives asynchronously (`state.rs` ~4317, requested on every session load — a normal race independent of the seller), gap becomes `NoEncryptionKey`, and the outer button would silently relabel to "Publish details" and publish `card.details` (the old, on-record values) while the form with unsaved edits stayed open underneath.

   Fix: `store_details_button_action` now also takes `is_editing` and always returns `ToggleForm` (Cancel) when this store's form is open, regardless of gap — an open form is always closable, and the gap is read fresh the next time the button renders. Added `an_open_form_always_cancels_even_if_the_gap_is_now_no_encryption_key` and `an_open_form_always_cancels_regardless_of_gap`; verified both fail without the fix (reverted to ignore `is_editing`, got `left: PublishNow, right: ToggleForm`), then restored.

2. **Fixed — no in-flight guard.** A double-click on the one-click `PublishNow` path could fire two real publishes with duplicate notifications.

   Fix: `AppState::store_publish_in_flight(store_contract_id)` reports whether a publish for that store is still waiting on the certificate (`pending_store_edit`) or the delegate's `SignResult` (`pending_signatures`) — both pre-existing fields, no new state invented. The button is `disabled` while true, and the click handler re-checks fresh state (not the render-time snapshot) before publishing, closing the race where two clicks land before Dioxus re-renders the `disabled` attribute. Both backing fields are already cleared on every completion and failure path (`SignResult` match, send failure, every ghostkey `Error`/`AccessDenied`/`NoIdentityAvailable`/`PermissionDenied` arm), so this cannot get stuck reporting `true`. Added `publish_in_flight_covers_the_certificate_and_signature_wait` and `publish_in_flight_clears_when_the_vault_denies_the_prompt`; verified both fail with the check hardcoded to `false`, then restored.

`cargo test --workspace` (503 UI tests + contract/delegate crates) green, `cargo clippy --workspace --all-targets` clean, `cargo fmt --check` clean. CI green at 84f07d0.

[AI-assisted - Claude]

sanity:
## Re-review outcome (commit 1cc0d8a)

- **Finding 1 (network-phase duplicate)** — accepted as non-blocking. \`store_publish_in_flight\` covers the certificate/sign phase only: the \`pending_signatures\` entry for a queued \`StoreInfo\` is removed as soon as \`SignResult\` matches it, before \`submit_store_info_by_id\`'s network write even starts, so a second click landing during that in-flight write can queue a second, real publish. That duplicate carries identical content at the next sequenced version, so it's wasted work, not data corruption — not worth the added complexity of extending the in-flight window through the network write for this PR's scope.
- **Findings 2 and 4** (single global \`pending_store_edit\` slot silently drops a concurrent second-store edit; no timeout on the certificate/sign wait, so a non-responding delegate leaves the button stuck disabled for the session) — moved to follow-up **#84**, since both are pre-existing defects in the certificate/sign wait, not something this PR introduces or worsens.
- **Doc comment corrected**: \`store_publish_in_flight\`'s doc comment overstated the guarantee ("can never get stuck reporting true"); now states plainly what it does and doesn't cover, per the above.

[AI-assisted - Claude]

Issue #81, opened 2026-09-18
contracts: unsigned certificate fields and equal-version store info still let two peers keep different state


sanity:
Found by `fdev verify-merge` while fixing #22, #26 and #55. Pre-existing, and deliberately not bundled into that PR: each needs a design decision rather than a mechanical fix, and one of them (the listing certificate) changes what a seller signs.

## The defect

Three places where two valid states hold different values in a field the merge does not resolve deterministically. Each is first-writer-wins, so two peers that receive the two values in opposite orders keep different bytes forever. `fdev verify-merge` (0.3.295) reports `state_commutativity` and `reconciliation_cycle` for each, against the contract WASM built on the fix branch:

| Where | What differs | Who can produce the second value |
|---|---|---|
| `AuthorizedListing::certificate_pem` (store contract) | The PEM next to a listing. It is not covered by the listing's signature, and `ListingId` does not include it, so a second copy with a different PEM has the same id and still verifies. `ListingsV1::apply_delta` skips an id it already holds, and the summary names only ids, so the two copies are never exchanged. | Anyone who has read the listing. |
| `ReputationStateV1::owner_certificate_pem` (reputation contract) | Unsigned. The merge fills it only when the local copy is empty. | Anyone, on a peer that saw the state with an empty certificate first. |
| `AuthorizedStoreInfoV1` at an equal `version` (store contract) | Two different details the seller signed at one version. `apply_delta` ignores an update whose version is not higher, and the summary is only the version. | The seller only (equivocation, or two tabs publishing at once). |

Reproduction: two states differing only in that field, then `fdev verify-merge --wasm <contract.wasm> --params <params.cbor> --state a.cbor --state b.cbor`. The generator used is not committed; the PR for #22/#26/#55 lists the commands and results.

## Why it matters

The listing case is the one users would see: `ui/src/state.rs::unverified_listings` verifies each listing's own certificate, so a third party who swaps in a junk PEM can make a genuine listing show as unverified on every peer it reaches first, and those peers never recover.

## Why it was not fixed with the rest

A deterministic tie-break over the bytes (keep the smaller, as the order and feedback merges do) would converge, but for an UNSIGNED field it hands the choice to whoever grinds the smallest value, which spreads the junk PEM everywhere instead of to some peers. The options worth weighing:

1. Cover the certificate with the signature (listings: move it into the signed terms; reputation: have the owner sign it). Changes what a seller signs, so it re-keys and changes the UI's publish path.
[...]

Issue #53, opened 2026-09-08
the purchase flow stops at Paid: no despatch, no feedback send path, no way to cancel


sanity:
## Summary

A Harvest purchase currently has no ending. Tracing the flow end to end against `origin/main`, everything from "buyer opens a link" through "the order becomes `Paid`" is implemented and works — and then it stops. There is no despatch, no delivery, no completion, and **the accountability mechanism the whole reputation design rests on has no send path**.

This is one issue rather than three because they are one hole: the transaction has no second half.

## What works today, for context

Nine steps, of which only one structurally requires the seller:

1. Buyer opens `#store=<id>` — GET with `subscribe:true`, no ghostkey or delegate needed (`store_link.rs:141-163`).
2. Buyer checks the certificate, ghostkey tier/age and reputation contract.
3. Buyer sends an encrypted `OrderRequest` (shipping address inside the AEAD) into the seller's mailbox — ephemeral X25519, ECDH against the published `encryption_public_key` (`state.rs:2067-2095`, `messaging.rs:305-335`).
4. Seller reads it; their delegate derives the conversation key fresh.
5. **Seller issues the invoice** — delegate derives the next address from an xpub that never leaves it, ghostkey signs the `Order`. `verify_terms` exists so a buyer cannot create one (`invoice_form.rs:6-9`).
6. Buyer pays from an external wallet.
7. Tip contract and per-order address contract are subscribed; both content-addressed, so either party computes the id.
8. `AwaitingPayment → Paid`, authorised by `OrderPaymentProof` and **signed by nobody** — any peer holding the bridge claims can publish it (`payment.rs:1195-1197`, `state.rs:2757-2761`).

Step 5 is the only point requiring the seller to act. That is worth keeping in view, because it is also the step a resident delegate could take over.

### 1. Nothing records despatch or delivery

`OrderStatus` has exactly four variants — `AwaitingPayment`, `Cancelled`, `Paid`, `PaymentReversed` (`common/src/payment.rs:190-204`). There is no `Shipped`/`Despatched`/`Fulfilled`, and a repo-wide search finds no equivalent anywhere.

`Fulfilled` was **deliberately removed**, for a good reason recorded at `payment.rs:220-231`: a seller-signed status there would let a seller bury a genuine reorg under a status they issued themselves. That reasoning is sound and should not be reversed casually — but the consequence is that a paid order and a delivered order are indistinguishable, to both parties, forever.

[...]

PR #82, opened 2026-09-18
fix(contracts): sign whole feedback entries, canonical state, associative caps, empty state (#22, #26, #55, #85)

### LINKED ISSUE #22: reputation: feedback is identified by a nonce the signature does not cover

### LINKED ISSUE #26: contract: ListingsV1::verify does not require sortedness, so two peers can converge differently

### LINKED ISSUE #55: contracts: reputation and store cannot decode an empty summary, so a new subscriber's first exchange fails
Found by the review on #54. Pre-existing, not introduced there, and deliberately not bundled into it: it is a behaviour change rather than an encoding one, it affects two contracts, and bundling would re-stale a review that had already run.

## The defect

`get_state_delta` in the reputation and store contracts cannot decode an empty summary, so the first exchange a new subscriber makes is answered with a decode error instead of state.

Sequence:

1. A peer holds state. A new subscriber has none.
2. The subscriber calls `summarize_state`, which correctly returns `StateSummary::from(vec![])` for empty state (`contracts/reputation-contract/src/lib.rs:110-111`).
3. The holder is asked for the delta with that zero-byte summary.
4. `from_reader::<ReputationSummary, &[u8]>(&[])` at `contracts/reputation-contract/src/lib.rs:134` returns `UnexpectedEof`, so the call fails with `ContractError::Deser`.

An empty `BTreeSet` encodes as `0x80`, never as zero bytes, so the two conventions never meet. Line 132 has the mirror-image omission: a holder whose own state is empty fails on `from_reader::<ReputationStateV1>` before the summary is reached.

`contracts/store-contract/src/lib.rs:277-287` has the same pair of omissions.

## This is not hypothetical

The mailbox contract already guards both cases, at `contracts/mailbox-contract/src/lib.rs:152-153` and `157-162`, and its comment records the behaviour as observed rather than predicted:

> before this the very first exchange a new subscriber makes -- it summarizes its absent state, which is zero bytes, and asks a holder for the difference -- was answered with a decode error rather than the mailbox.

## The fix

The mailbox's own pattern, applied to both contracts: treat zero bytes as "there is no state here yet" rather than as a malformed encoding, for both the state and the summary.

```rust
if state.as_ref().is_empty() {
    return Ok(StateDelta::from(vec![]));
}
let old_summary = if summary.as_ref().is_empty() {
    ReputationSummary::default()
} else {
    from_reader::<ReputationSummary, &[u8]>(summary.as_ref())
        .map_err(|e| ContractError::Deser(e.to_string()))?
};
```

## Worth doing before the next re-key rather than after

Not urgent, but cheap now and not later. Both contracts re-key on any `harvest-common` change, and while nobody has clicked Create Store a re-key costs a registry row and nothing else. `legacy/store_contract.toml`'s V9 row sets the precedent for bundling a fix into a re-key on exactly this reasoning.

## What is not established

Whether current freenet-core still calls `get_state_delta` with a zero-byte summary was not verified against core; the mailbox comment above is the evidence, and it is the team's own observation rather than a reading of core. If core no longer does this, the change becomes a consistency cleanup rather than a bug fix, and is still worth making so the three contracts agree.

[AI-assisted - Claude]

### LINKED ISSUE #85: contracts: mailbox and order caps are not associative; mailbox verify accepts non-canonical state
Found by an `fdev verify-merge` (0.3.295) baseline run against `main` (`f1d8614`), with corpora built from the repo's own types and signing. These are the same defect class as #26: merge laws the network relies on do not hold. Each one re-keys a contract, so they are being fixed in #82 alongside #22, #26 and #55.

## 1. Mailbox byte cap is not associative

`enforce_message_cap` (`common/src/mailbox.rs`) walks messages in rank order and SKIPS any message that does not fit the byte budget. A message skipped while over budget never comes back.

**Counterexample.** fdev found this in the built WASM (`state_associativity`, 2 of 2 cases):
- P = 501 messages totalling `MAX_MAILBOX_BYTES - 500`
- Q = one old 600-byte message x
- R = one new 600-byte message r1

With these, `(P+Q)+R` loses x, and `P+(Q+R)` keeps it.

**A prefix walk does not fix this.** A prefix walk stops at the first message that does not fit instead of skipping it, which was the rule before 2026-09-05. It fails with the same shape at a budget of 10 bytes:
- a (rank 1, 5 bytes), z (rank 2, 8 bytes), y (rank 3, 3 bytes)
- `({a}+{z})+{y}` = `{a}+{y}` = `{a, y}`
- `{a}+({z}+{y})` = `{a}+{z}` = `{a}`

Neither rule is path independent. A size budget is a knapsack constraint, and a greedy selection over a fixed ranking is path independent only when the constraint is a matroid. A count cap is one, and a byte budget over variable sizes is not. The fix has to change the shape of the bound, not the walk.

## 2. Store `MAX_ORDERS` cap is not associative

`enforce_order_cap` (`common/src/store.rs`) drops terminal orders first. Terminal-ness is not monotone in the per-key merge's status rank:

| Status | Rank | Terminal? |
|---|---|---|
| Awaiting | 0 | no |
| Cancelled | 1 | yes |
| Paid | 2 | no |
| Reversed | 3 | yes |

So a key's keep-priority changes when its record is merged. A key pruned on one peer can also come back from a peer that still holds an older version of it.

**Counterexample**, reproduced natively (fdev skips states over 4 MiB without saying so):
- P = {x Awaiting, newest}
- Q = {x Cancelled}
- R = 4096 older Awaiting orders

With these, `(P+Q)+R` drops x, and `P+(Q+R)` keeps x as Awaiting.

## 3. Mailbox messages in non-canonical order validate

An unsorted mailbox passes `verify`, and a state is not equal to itself merged with itself (`state_idempotence`, plus `state_commutativity` against sorted states). Same class as #26.

## 4. Mailbox `verify` accepts an oversized message that `apply_delta` refuses

`verify` accepts a message over `MAX_MESSAGE_BYTES`, but `apply_delta` drops it. So whether the message survives a merge depends on which side it was on, and the merge is not commutative.

[AI-assisted - Claude]

sanity:
Fixes #26, fixes #55, fixes #22, fixes #85.

**#22 has no user-visible effect until feedback submission is wired (#53).** No UI path builds or publishes a `FeedbackEntry` today. This PR makes the reputation contract correct before that path exists.

Four contract fixes in one re-key, because each re-keys a contract and a re-key is free only while Harvest has no users. #85 was added in a second commit, after an `fdev verify-merge` baseline on `main` found further merge-law violations of the same class.

## What changed

**#22, feedback identified by a nonce the signature does not cover.** `FeedbackToken` gains `entry_key`, an Ed25519 verifying key the buyer generates for that token alone. The seller's blind signature covers the token, so it covers the key, and `FeedbackEntry` gains `entry_signature`, made by that key over every other field of the entry: token, RSA signature, category, comment, timestamp. The RSA signature is inside the preimage on purpose, because the seller can mint new PSS signatures on a token at will. So a third party can no longer build a variant at all. The buyer still can, by signing twice, so two entries for one token resolve to the smaller CBOR encoding. The summary now names entry digests instead of nonces, because a nonce summary told two peers holding different entries for one token that each already had the other's, so neither was ever sent. `ReputationStateV1::merge` is now the single merge used by the contract's `UpdateData::State` arm and by the migration fold. The fold's exclusion detector and all five `nonce-identity-waiver` markers are removed with the defect, and the scrape now expects 0 waivers.

The token's nonce must be `FeedbackToken::nonce_for(entry_key)`, a domain-separated BLAKE3 of the key, so a token's one slot belongs to its key. This was added after the Full-tier review (Must Fix 1): the seller holds the RSA key, so without the binding it could mint a token for a buyer's published slot with a key of its own, grind that key until its entry sorted first, and replace the buyer's feedback on every peer.

**#26, `ListingsV1::verify` does not require sortedness.** `verify` now requires listings strictly ascending by id, which rejects both an unsorted list and a listing held twice. Both used to verify. `ListingsV1::normalize` (sort, then drop a repeated id) runs in `apply_delta`, in the store contract's `update_state` before it encodes, and in the migration fold. The last two are needed because the scaffold skips a child's `apply_delta` when a merge brings nothing, so a stored non-canonical state would otherwise be written back as it came. The issue's constraint (rejecting existing state in `verify` is irreversible) is met by the re-key: the new generation starts empty, and the fold normalises what it carries. The reputation contract gets the same canonical-form check: feedback strictly ascending by nonce, and `used_nonces` exactly those nonces.

**#55, empty summary cannot be decoded.** In reputation and store `get_state_delta`, zero bytes means "no state here", following the mailbox's pattern. An empty state answers with an empty delta. An empty summary is treated as the summary of the empty state, so a new subscriber gets everything the holder has.

**#85, caps not associative, mailbox verify too permissive.**

- **Mailbox byte cap.** The walk skipped any message that did not fit the byte budget, and a skipped message never came back when a later merge freed the room. So `(P+Q)+R` and `P+(Q+R)` differed.
- **A prefix walk is not the fix either.** Stopping at the first message that does not fit fails the same law another way, and `the_byte_bound_is_associative_with_a_large_message_mid_ranking` pins that. A byte budget over variable sizes is a knapsack constraint, not a matroid, so no greedy walk over a fixed ranking is path independent under it.
- **What replaces it.** The byte bound is now a count cap per size class, `SIZE_CLASS_CAPS`: 512/128/64/24 for the 1/4/16/64 KiB buckets, under the existing 512-message `MAX_MESSAGES`. Class caps nested inside an overall cap form a laminar matroid, so the greedy pick is path independent. A compile-time assertion keeps the class caps within `MAX_MAILBOX_BYTES`, at about 3.7 MiB.
- **Order cap.** It dropped terminal orders first. Terminal-ness is not monotone in the status rank the per-key merge maximises, so a key's keep-priority moved as it merged. The cap now keeps the newest `MAX_ORDERS` by signed `created_at`, then id, which no merge can change. The cost: an old Paid order can now be dropped before a newer Cancelled one. Only the seller can create orders.
- **Mailbox `verify`.** It now requires exactly what `apply_delta` leaves behind. Every message must be within `MAX_MESSAGE_BYTES` (it used to accept ones `apply_delta` refuses), every size class within its cap, and the messages strictly ascending in `canonical_order`. It used to accept unsorted and duplicate states.
- **Empty state, found by the empty corpora.** The store and reputation `update_state` `State` arms failed to decode a zero-byte state. All three contracts also answered zero bytes merged with zero bytes with the encoded default, so `merge(A, A) != A` for the empty state. Now zero bytes in, with nothing but zero bytes merged in, gives zero bytes out; anything non-empty arriving is encoded.

**From the Full-tier review, beyond Must Fix 1:**

- Every `validate_state` refuses a state that does not re-encode to its own bytes (trailing bytes, unknown keys, non-minimal encodings).
- The order summary names each order by the full 32-byte BLAKE3 instead of 8 bytes.
- The mailbox fold reports messages pruned to the count caps.
- Seeded, deterministic merge-law property tests cover every state type. They add about 3 minutes to CI's test job (2 minutes before, 5.5 now). They check the algebra (idempotence, commutativity, associativity) over random states, not the exact cap values, so the exact-cap tests (`the_byte_bound_is_associative_at_the_budget`, `one_class_cap_alone_is_associative_and_exact`, `the_order_cap_is_associative_when_a_status_changes_at_the_cap`) stay load-bearing.
- The seller-controlled `created_at` eviction risk is documented (see Re-keys).

**From the re-review:**

- Version-0 store info must be exactly the default. Before this, anyone could inject a name, certificate or encryption key into a store whose seller had not published details. The buyer side resets version 0 to the default and still keeps it loaded, so the seller's "never published" repair prompt stays reachable. The migration fold resets it too, so a predecessor holding injected version-0 content still carries its listings, and an unsigned reputation link is never followed.
- `verify_strict` for feedback entries is pinned by a test.
- The nonce binding is checked before the RSA signature.
- The order summary encodes its id and digest as CBOR byte strings, about 280 KiB at `MAX_ORDERS` instead of about 512 KiB. `OrderId` and order records are unchanged.
- `docs/design.md` no longer claims the contract checks a token's target.

## Re-keys

All four artifacts move. Rows appended for the outgoing hashes and pinned in `the_recorded_hashes_are_the_ones_derived_from_git_history`:

| Artifact | Outgoing (now legacy) | New | Cause |
|---|---|---|---|
| reputation_contract | `eef8685c...` (V11) | `58c9632d...` | its own: #22, canonical verify, #55 |
| store_contract | `c4e21292...` (V15, the #83 build) | `95f464c4...` | its own: #26, #55, #85 order cap |
| mailbox_contract | `5c0145d2...` (V11) | `f78ff5a2...` | its own: #85 size-class caps and canonical verify, empty state |
| harvest_delegate | `a1118c09...` (V14, the #83 build, key `4ad013d5...`) | `da3c9ca1...` | `FeedbackToken` is a wire type it stores |

**Known limit of the order cap.** The cap keeps the newest orders by `created_at`, which the seller signs and nothing bounds. So the seller can evict old orders, Paid ones included, by creating `MAX_ORDERS` newer ones, or pin slots with far-future dates. Ranking by anchor height was considered and does not help, because the contract cannot check an unpaid order's anchor against the chain either. This is recorded in `docs/untested-invariants.md` and in the store V15 row.

Rebased onto #83 (`1b0d3c3`). #83's own rows (store V14 `3f47ab79`, delegate V13 `08eac64a`) are kept as written, and this PR's rows follow them as store V15 and delegate V14.

Parameter structs are unchanged (`cbor(StoreParameters)` still 56 bytes), so no derivation split is needed. Migration consequences, all in the registry rows:

- **Reputation:** a predecessor entry has neither new field, so a predecessor state holding any feedback would not decode. None exists, because no UI path builds or publishes a `FeedbackEntry`. Every published reputation state is a certificate plus an empty list, and that decodes unchanged.
- **Store:** V13 records still verify. A predecessor state that is unsorted or holds a listing twice is normalised by the fold rather than refused.
- **Delegate:** no `TransactionRecord` exists to break, because the UI never sends `BeginTransaction`.
- **Mailbox:** a V10 state decodes unchanged and is normalised by the fold. A mailbox holding more of one size class than its new cap (for example more than 24 top-bucket messages) is pruned to that cap by the fold. Only test mailboxes exist.

## Merge-law verification (`fdev verify-merge`, fdev 0.3.295)

**The full baseline corpus**, from the coordinator's generator. The states are built with the repo's own types and signing. The generator was adapted only for #22's new entry fields, and to stop asserting that adversarial states verify. Each corpus runs twice, once from `--state`/`--transition` and once from a bundle that carries delta steps, and then once per property. "Before" is `main` at `f1d8614`. "After" is this branch's committed WASM, rebased onto #83 (`1b0d3c3`). The generator's fixture orders now carry an anchor (height 90, payments at 100) so they pass #83's payment window.

```
(cd gen && HARVEST_WASM_DIR=<repo>/target/wasm32-unknown-unknown/release cargo run --release -- ../corpus)
./run.sh <repo>     # fdev verify-merge --wasm ... --params ... --state ... --transition ... --max-cases 2000 --json
```

| Corpus | Before (main `f1d8614`) | After (this PR) |
|---|---|---|
| store | 1316/1316 held | 1316/1316 held |
| store-empty | 1311 held, **115 inconclusive** (zero-byte decode errors) | 1425 held, 1 inconclusive (no delta path) |
| store-adv | **34 violations** | 6 violations, all #81; the unsorted and duplicate states are refused by `validate_state` |
| store-cap | 15/15 held | 15/15 held |
| store-cap2 (review; tied `created_at` at the cap) | 82/82 held (on `cf714c7`) | 82/82 held |
| store-noncanon (review) | **2 violations** (on `cf714c7`) | 0 violations; the trailing-byte and unknown-key states are refused (41 inconclusive, "not in canonical CBOR encoding") |
| reputation | 297 held, 1 inconclusive (no delta path) | same |
| reputation-empty | 296 held, **56 inconclusive** | 349 held, 3 inconclusive (no delta path) |
| reputation-adv (+ a seller-minted token for a published slot) | **24 violations** | 8 violations, all #81 (`owner_certificate_pem`); the third-party rewrite, the seller-minted slot and the unsorted feedback are refused, and the buyer-signed second entry converges |
| mailbox | 298/298 held | 298/298 held |
| mailbox-empty | **1 violation** | 0 violations, 1 inconclusive (no delta path) |
| mailbox-adv | **9 violations** | 0 violations; both states refused |
| mailbox-cap | **2 violations** (`state_associativity`) | 32/32 held |
| mailbox-cap2 (review; class-3 cap with a tie) | 85/85 held (on `cf714c7`) | 85/85 held |
| mailbox-cap3 (review; 512-message total cap) | 56/56 held (on `cf714c7`) | 56/56 held |
| mailbox-noncanon (review) | **2 violations** (on `cf714c7`) | 0 violations; refused as non-canonical |
| store-v0 (re-review; unsigned version-0 details) | **6 violations** (on `50fb3fa`) | 0 violations; refused ("store info at version 0 must be empty") |
| store-rr (re-review; full cap, boundary order Paid with differing proofs) | 108 held, 1 inconclusive (on `50fb3fa`) | same |
| reputation-rr (re-review; small-order entry keys, non-minimal CBOR) | 0 violations (on `50fb3fa`) | 0 violations; weak-key forgeries and non-minimal encodings refused |
| mailbox-rr (re-review; non-minimal encoded default) | 0 violations (on `50fb3fa`) | 0 violations; non-minimal encoding refused |

The remaining violations were attributed state by state with `minimise.sh`. Every one involves an adversarial state that is already filed as #81.

**On store-cap:** fdev silently skips any case over 4 MiB, so the order-cap counterexample was checked natively by the generator. `(P+Q)+R == P+(Q+R)` is `false` on `main` and `true` here. `the_order_cap_is_associative_when_a_status_changes_at_the_cap` pins it in `cargo test`.

**The earlier targeted runs from the first commit** (my own generator, before/after) still stand. The unsorted, duplicate and same-token-variant states violated commutativity before and do not after.

## Tests, and each new check made to fail

Every new check was mutated and observed red, then restored:

| Mutation | Red test |
|---|---|
| listings `verify` canonical check removed | `verify_refuses_unsorted_or_repeated_listings` |
| listings `>=` weakened to `>` | same |
| `normalize` without dedup | `merging_normalises_a_non_canonical_state` |
| `normalize` without sort | `listing_merge_is_commutative_including_non_canonical_inputs` |
| store `update_state` without `normalize` | `update_state_writes_canonical_listings_when_the_merge_brings_nothing` |
| migration fold without `normalize` | `a_fold_normalises_a_non_canonical_base` |
| reputation ordering check removed | `verify_refuses_non_canonical_state` |
| reputation `used_nonces` equality removed | same |
| Ed25519 entry check skipped | `a_third_party_cannot_rewrite_an_entry` |
| comment dropped from signed bytes | `every_field_of_an_entry_is_signed` |
| RSA signature dropped from signed bytes | same |
| tie-break replaced by first-writer-wins | `two_signed_entries_for_one_token_converge` |
| summary keyed on nonce again | `peers_holding_different_entries_for_one_token_exchange_them` |
| `merge` skips `apply_delta` for an empty other side | `apply_delta_normalises_the_state_it_is_applied_to` |
| each of the four #55 guards removed (2 per contract) | `an_empty_summary_is_answered_with_everything_held` / `an_empty_state_answers_with_an_empty_delta` in that contract |
| store/reputation `State` arm zero-byte guard removed (pre-fix code) | `merging_a_zero_byte_state_changes_nothing` in each |
| empty-state rule: flag never set / not cleared by a State / not cleared by a Delta, x3 contracts (9 mutations) | `the_empty_state_is_idempotent_and_the_rule_is_commutative` (first two) and `a_delta_applied_to_the_empty_state_is_encoded` (third), in each contract |
| order cap ranks terminal orders first again | `the_order_cap_is_associative_when_a_status_changes_at_the_cap`, `the_order_cap_obeys_the_merge_laws_at_the_cap` |
| mailbox skipping byte walk restored | `the_byte_bound_is_associative_at_the_budget` |
| mailbox prefix byte walk restored | same, plus `the_byte_bound_is_associative_with_a_large_message_mid_ranking` |
| class cap ignored (overall count only) | `the_mailbox_is_bounded_in_bytes_and_not_only_in_count` |
| mailbox `verify` order check / oversized check / class-cap check removed | `verify_refuses_non_canonical_order` / `verify_refuses_an_oversized_message` / `verify_refuses_a_state_the_cap_would_prune` |
| mailbox `apply_delta` final canonical sort removed | 5 tests, including `verify_refuses_non_canonical_order` |

Locally: `cargo fmt --check`, `cargo clippy --locked --workspace --all-targets -D warnings`, `cargo test --locked --workspace`, the `address-guard` clippy and test, wasm32 clippy of the UI, `cargo check --locked --manifest-path tests/rehearsal/Cargo.toml`, and `scripts/check-code-hashes.sh`. All pass.

`docs/untested-invariants.md` gains a section for these mechanisms, and it records two honest "no" rows. First, `fdev verify-merge` is manual and not in CI. Second, the buyer has nowhere to keep the entry key's secret yet: feedback submission is unwired (#53), and whoever wires it has to add that storage.

Two existing mailbox tests changed meaning. `verify_accepts_an_over_budget_state_so_an_existing_mailbox_is_never_stranded` is inverted as `verify_refuses_an_over_budget_state_now_the_rekey_starts_empty`. Its stranding argument was about state held at the same address, and the re-key starts every mailbox empty. Separately, the comments on `dedupe_identical_entries` and `apply_delta`'s final sort described those two as mutually redundant. That stopped being true once the cap began sorting on every call, and the comments now say so.

## Not done

Not reviewed yet and not published. Full-tier review is still to run.

[AI-assisted - Claude]


sanity (commented):
## Comprehensive PR Review: #82

### Summary
- **HEAD SHA reviewed:** `cf714c7d635d339bcb68e874b0306cbc5906017d`. CI is green.
- **Linked issues:** #26, #55, #22, #85.
- **Review tier:** Full (contract WASM, wire format, migration).
- **Reviewers run:** code-first, testing, skeptical, big-picture, and a dedicated merge-law/migration lens that re-ran `fdev verify-merge` with five extra boundary corpora.
- **External model:** Codex is over its usage limit (resets 2026-09-19), so diverse Claude lenses were substituted, per the review rule.

### What holds
- **Merge laws:** every changed merge is commutative, associative and idempotent over validly signed content, including at the caps. The merge-law lens proved each one. For the order cap, pruning is top-N over a ranking taken from the signed terms. The mailbox is a laminar matroid.
- **New boundary corpora:**
  - store-cap2 (4096 orders with tied `created_at`): 82/82 hold.
  - mailbox-cap2 (class-3 boundary with a tie): 85/85 hold.
  - mailbox-cap3 (512-message total cap): 56/56 hold.
- **Remaining violations:** all of them trace to the #81 states.
- **The byte-budget prefix walk is non-associative:** confirmed independently with a second counterexample.
- **Registry:** the rows and outgoing hashes match main's committed WASM byte for byte.
- **Mutation checks:** the testing lens made 7 mutations of its own, and none survived.

### Must Fix
1. **The seller can replace any buyer's feedback on every peer, permanently.**
   - **Why it's possible:** the slot is keyed on `token.nonce` alone (`common/src/reputation.rs:326`), and nothing ties the nonce to `token.entry_key`.
   - **Attack:** the RSA key holder, the seller, can mint and sign a token with a published nonce and their own `entry_key`. They then sign a neutered entry and grind a key whose CBOR sorts first. `canonical_rank` then selects it on every peer.
   - **Regression:** before this PR, a variant only won on peers that had not seen the original.
   - **Fix:** bind the slot to the key, e.g. require `nonce == H(domain || entry_key)`, or key the slot by the token digest. Add a test where the RSA key holder mints a different-key token for a published nonce.
   - **Docs:** correct `docs/design.md:122` ("nobody but Bob can change what it says").

### Should Fix
2. **Byte-level canonicality.**
   - **Where:** every `validate_state` decodes leniently and never checks that re-encoding gives the same bytes (`contracts/store-contract/src/lib.rs:83`; the mailbox and reputation contracts have the same shape).
   - **Effect:** a state with a trailing byte or an unknown map key is accepted, and merge then rewrites it. The summaries are equal, so no delta ever repairs it. That is the #26 defect one layer down.
   - **Evidence:** `fdev verify-merge` reports `state_idempotence` on the new store-noncanon and mailbox-noncanon corpora.
   - **Fix:** reject when `to_cbor(decoded) != bytes`. It's free during this re-key.
3. **An in-repo merge-law property test.**
   - **Gap:** CI only pins the specific counterexamples already found, and `fdev verify-merge` is manual, so a new kind of violation would go unnoticed.
   - **Fix:** add a seeded, deterministic random test per state type (no new dependency) that asserts commutativity and associativity byte-for-byte over a few hundred small valid states, including states at the caps.
4. **Widen the order summary digest.** It is truncated to 8 bytes (`common/src/store.rs:371`, `:572`). A birthday collision between two same-rank Paid variants, about 2^32 work, would stop those peers exchanging them. Widening it is free during this re-key.
5. **The order cap ranks by `created_at`, which the seller chooses and nothing bounds.**
   - Far-future orders pin slots.
   - Flooding with Cancelled orders can evict Paid records, which are the buyer's evidence.
   - Only the seller can create orders (`buy_view.rs:406`).
   - The contract has no clock, so bounding it isn't trivial. At minimum, document the seller-side eviction of buyer evidence in `docs/untested-invariants.md` and in the V15 row. Consider ranking by anchor height instead.
6. **Stale comments and docs:**
   - `ui/src/migrate.rs:900-908` and `ui/src/migrate/tests.rs:1576` still say `verify` "tolerates an over-budget state".
   - `enforce_message_cap`'s doc (`common/src/mailbox.rs:852-855`) says a small-message flood "no longer touches a larger class", which is false while class 0's cap equals `MAX_MESSAGES`.
   - Add a comment at the inverted mailbox test: its justification holds only for a re-key that starts empty.
7. **Mailbox fold:** report messages pruned to the class caps, as oversized ones already are.
8. **Class-cap test depth:** the per-class cap is caught by a single test. Add an associativity test isolating one class's cap.

### Consider
- The PR body should say up front that the #22 fix has no user-visible effect until feedback submission is wired (#53).

### Policy decisions for a human
- The order cap now prunes by age regardless of status, so an old Paid order can go before a newer Cancelled one.
- The mailbox size-class caps are 512/128/64/24, so at most 24 top-size messages.

Both are sound for the merge laws. The specific choices are product decisions.

### Verdict
**State:** Needs Changes. Re-review is required after the fix, triggered by Must Fix 1 and Should Fix 2 and 4, which change contract code again.

[AI-assisted - Claude]

sanity:
**Policy sign-off (Ian, 2026-09-18).** Both trade-offs flagged in review are accepted as implemented:
- The order cap prunes the oldest orders whatever their status. An old Paid order can be dropped before a newer Cancelled one, but only past 4,096 orders in one store.
- The mailbox size-class caps are 512/128/64/24.

[AI-assisted - Claude]

sanity:
## Review findings and how each was resolved

Head: `50fb3fa4c10a4d3c8e55d97f1c39eb1a31c56394`, rebased onto current main (#86 and #87 touch neither the contracts nor the WASM). CI is green at that SHA.

**Must Fix 1: the seller could take over a buyer's feedback slot.**
- **Fix:** the token's nonce must now be `FeedbackToken::nonce_for(entry_key)`, a BLAKE3 of the key with its own domain label. `FeedbackEntry::verify` refuses any other nonce. Tokens are built with `FeedbackToken::new`.
- **Why this option:** keying slots by the whole token's digest also closes the hole, but it gives up "one token, one slot". The replay check and the tie-break would then only apply to byte-identical tokens, which is harder to reason about. With the nonce bound to the key, putting a second entry in a published slot needs a preimage.
- **Test:** `the_rsa_key_holder_cannot_mint_a_token_for_a_published_slot`. The RSA key holder mints a token for the buyer's published nonce with its own key, grinds that key until the entry sorts first, and applies it in both orders. The buyer's entry survives. The test goes red with the check removed.
- **Docs:** `docs/design.md` (the step list, what the contract validates, and the struct) is corrected. A nonce-identity scrape waiver is added for the derivation check, which compares a nonce against a hash rather than deciding identity, and the expected waiver count is updated with the reason.

**Should Fix 2: byte-canonical `validate_state`.**
- **Fix:** all three contracts now refuse a state whose bytes differ from its own re-encoding, via `harvest_common::is_canonical_cbor`.
- **Test:** `validate_state_refuses_non_canonical_bytes` in each contract covers a trailing byte and an unknown key. Each goes red with the check removed.
- **fdev:** store-noncanon and mailbox-noncanon now report 0 violations, where both had 2 before. The non-canonical states are refused as "not in canonical CBOR encoding".

**Should Fix 3: seeded random merge-law tests.**
- **What:** a xorshift checker in `harvest_common::merge_laws`, with no new dependency. It checks idempotence, commutativity and associativity byte for byte.
- **Coverage:**
  - Mailbox: 200+ states, including states at the top size class's cap and more top-size messages than fit in the byte budget. Checked through both the whole-state merge and the summary/delta exchange.
  - Reputation: genuinely signed entries, including two buyer-signed entries for each of two tokens. Checked through both paths.
  - Store: signed store details, listings, and orders at every status, including two valid Paid proofs for one order.
  - The order set at `MAX_ORDERS`: tied `created_at` values, and keys held at different statuses.
- **Mutations:** red under the prefix byte walk, the status-ranked order cap, first-writer-wins feedback, and an unsorted listing normalise.
- **What it misses:** the skipping byte walk survives it, so `the_byte_bound_is_associative_at_the_budget` stays the guard for that one.
- **Cost:** CI's test job went from 2 minutes to 5.5.

**Should Fix 4: order summary digest.** It is now the full 32-byte BLAKE3. `the_order_summary_digest_is_the_full_hash` goes red with an 8-byte truncation.

**Should Fix 5: seller-chosen `created_at`.** Documented, not fixed.
- **Where:** `docs/untested-invariants.md` (a new "policy" row) and the store V15 row.
- **Anchor height was considered and rejected.** The contract cannot check an unpaid order's anchor against the chain, so it is as free as `created_at`. Only a Paid order's anchor is bounded, by its payment window, and Paid orders are exactly what a flood evicts. Status cannot take part in the ranking without breaking associativity.

**Should Fix 6: stale comments.**
- `migrate.rs` (the fold rationale) and `migrate/tests.rs` (the drop test's doc) now describe both what held before and after #85.
- The `enforce_message_cap` doc now says a small-message flood still evicts every class, through the smallest class's cap, which equals `MAX_MESSAGES`.
- The inverted mailbox test now says its argument holds only because this re-key starts every mailbox empty.

**Should Fix 7: fold reports cap pruning.** `MailboxFold::pruned_by_cap` counts each pruned message once, even when it is on both sides, and appears in the warning. Test: `a_fold_that_prunes_to_the_caps_reports_it`, red with the count zeroed and with the warning dropped.

**Should Fix 8: one size class's cap, isolated.** `one_class_cap_alone_is_associative_and_exact` overfills only the 4 KiB class. It checks the laws, an exact cut to that class's cap, and that the other classes are untouched. It goes red when that class is given the smallest class's cap.

**Consider: the PR body now says up front** that #22 has no user-visible effect until #53.

**Re-keys:** the WASM is rebuilt: store `416abc98…`, reputation `08f26db6…`, mailbox `f7a4c329…`. The delegate is unchanged at `da3c9ca1…`. No new registry rows, since these are the same unpublished generations; the store V15, reputation V11 and mailbox V11 descriptions are updated.

**verify-merge:** every corpus, including the reviewer's five, is in the PR body's table. All remaining violations trace to #81 states, which I checked state by state with `minimise.sh`.

[AI-assisted - Claude]

sanity (commented):
## Re-review: #82 at `50fb3fa`

**Lenses:** skeptical, testing (in its own worktree), merge-law/migration (re-ran `fdev verify-merge` on 16 corpora, including new adversarial ones), and code-first/big-picture. **External model:** Codex is over its usage limit (resets 2026-09-19), so these Claude lenses stand in for it.

**All eight round-1 findings are fixed, verified against the code:**
- The seller-minted slot is refused. The test goes red when the check is reverted.
- `validate_state` is byte-canonical in all three contracts.
- Seeded merge-law tests are in `cargo test`.
- The summary digest is 32 bytes.
- The remaining items are documented.

**Other results:**
- **Mutations:** none of the testing lens's mutations survived.
- **verify-merge:** every row of the author's table reproduces, and all remaining violations are #81 states.
- **Unlinkability:** the nonce binding keeps blind-signature unlinkability, because the seller only ever sees the blinded token.

### Should Fix (round 2)
1. **Unsigned store details at version 0.**
   - **Where:** `AuthorizedStoreInfoV1::verify` skips everything at `version == 0` (`common/src/store.rs:160`).
   - **Effect:** anyone can inject unsigned details (name, description, `encryption_public_key`) into a store whose details were never published. Two such states never converge; the new store-v0 corpus gives 6 violations.
   - **Worse:** the buyer view does not check the version (`ui/src/state.rs:2081`, `ui/src/components/store_view.rs:267`), so an injected encryption key becomes the one buyers encrypt to.
   - **Fix:** require version-0 info to equal the default, and have the buyer side treat version-0 info as "no details published".
2. **Pin `verify_strict`.** Swapping it for `verify` accepts forged signatures under small-order entry keys, and no test fails.
3. **Order the feedback checks cheapest first.** Run the nonce-derivation check before the RSA verify.
4. **Summary size.** The order summary at `MAX_ORDERS` is about 512 KiB, because 32-byte fields encode as CBOR integer arrays. Byte-string encoding would bring it to about 290 KiB. That is a wire change, so it's free only during this re-key.
5. **Docs:**
   - `docs/design.md:124` claims the contract checks the token's target. It does not.
   - Tidy the comment at `ui/src/migrate.rs:900`.
   - State in the PR body that the random merge-law tests check the algebra, not the cap values, so the exact-cap tests stay load-bearing.

### Follow-ups (not this PR)
- `[profile.test] opt-level = 2` cut the seeded tests from 181 s to 13 s with no loss of coverage. That's a CI change, so it goes in its own PR.
- The entry-key requirements for feedback submission are recorded on #53.

**State:** Needs Changes. After the fix, a focused re-check with the merge-law and skeptical lenses.

[AI-assisted - Claude]

sanity:
## Re-review (round 2): findings and how each was resolved

Head: `f5791feb57ea163ab304a8b733d6c93d7fe03db4`. CI is green at that SHA.

1. **Unsigned version-0 store info (Should).**
   - **Contract:** `AuthorizedStoreInfoV1::verify` now requires version-0 info to equal `Default`. Test: `version_zero_info_must_be_the_default`, covering a name, a key and a signature; red with the old skip restored.
   - **Buyer side:** `on_contract_state` resets a version-0 info to the default before anything reads it, and `BrowsingStore::info` is `None` at version 0. So no injected encryption key, certificate or reputation id is used. Test: `version_zero_store_details_are_not_believed`; red with either guard removed.
   - **fdev:** the store-v0 corpus is folded into my generator and goes from 6 violations to 0.
2. **`verify_strict` pinned.** `an_entry_under_a_small_order_key_is_refused` covers the identity key, its sign-bit variant and its non-canonical `y + p` encoding, each with the `R = identity, s = 0` forgery. It first asserts that the non-strict check accepts the forgery, then that the contract refuses it. Red with `verify_strict` swapped for `verify`. The reputation-rr corpus is folded in and is clean.
3. **Nonce check before RSA.** Moved. `the_cheap_nonce_check_runs_before_the_rsa_check` goes red if it is moved back.
4. **Order summary size.** Contained, so done. A `Bytes32` type encodes the summary's id and digest as CBOR byte strings. It is used only in the summary: `OrderId` sits inside every signed order, so changing its own encoding would have moved every order id and signature preimage. At `MAX_ORDERS` the summary is about 284 KiB instead of about 512 KiB. Test: `the_order_summary_at_the_cap_is_byte_strings`, checking the size bound, the round trip and the byte-string header; red with integer-array encoding.
5. **`docs/design.md`.** It no longer claims the target is checked. It says the target is not enforced because a contract cannot compute its own instance id, and that what actually binds a token is the seller's RSA key in the parameters.
6. **`ui/src/migrate.rs` fold comment.** Rewritten.
7. **PR body.** It now says the seeded random tests check the algebra, not exact cap values, and names the exact-cap tests that stay load-bearing.
8. **#81.** Corrected there: equal-version store info was not seller-only at version 0; that case is fixed here.

**Rebuilt WASM:** store `95f464c4…`, reputation `53efe914…`, mailbox `f78ff5a2…`. The mailbox moves only through `harvest-common`. The delegate is unchanged at `da3c9ca1…`. The legacy/ descriptions for store V15 and reputation V11 are updated, with no new rows.

**verify-merge:** 20 corpora, including store-v0 and the four rr corpora; the PR body's table is updated. The only violations left are the #81 states in store-adv (6) and reputation-adv (8).

[AI-assisted - Claude]

sanity:
## Round 3: findings and how each was resolved

Head: `ad8318e639337100746129a4716da947662b8270`. CI is green at that SHA.

1. **HIGH, a store stranded at version 0 was stuck on "Loading".** `on_contract_state` still resets a version-0 info to the default, but it now keeps `BrowsingStore::info` as `Some(default)`. The seller's paths therefore see the store as loaded and reach the NeverPublished repair prompt. Buyers still find no name, key or reputation link in it, because the default has none.
   - **New test:** `a_version_zero_store_is_resolved_and_prompts_to_publish` ingests `StoreStateV1::default()` through `on_contract_state` and asserts `store_details_are_resolved` and `store_details_gap == NeverPublished`. It fails on `f5791fe`: I checked that with the round-2 `None` restored, which turns it red.
   - **Updated test:** `version_zero_store_details_are_not_believed` now asserts the info is the default with no key. It goes red with either guard removed.
2. **Migration fold carried injected version-0 details forward.** `merge_store_reporting_discard` now resets a version-0 info to the default, next to `listings.normalize()`.
   - **Tests:** `a_fold_drops_unsigned_version_zero_details`, plus `every_fold_order_with_injected_version_zero_details_moves_the_listings`. The second runs the four orders you reproduced through the real fold and asserts that each forward state verifies, carries the listing, and drops the injected key. Both go red with the reset removed.
   - **In the built WASM:** the four forward states from the real fold give 15/15 held under `fdev verify-merge`. Before the fix they were refused.
   - **Registry:** the V15 legacy text no longer says the fold refuses such a predecessor; it now says the fold carries it with the version-0 info reset.
3. **LOW, `check_for_reputation_link`.** It now follows only a version > 0 link. Test: `only_a_signed_reputation_link_is_followed`, red with the gate removed.
4. **LOW, nonce-first comment.** It now says the attacker picks `entry_key` and can always pass this check. The ordering only saves RSA work on junk and honest mistakes, and what bounds RSA work per update is `apply_delta`'s digest skip.
5. **Stale comments.** `ui/src/migrate.rs`, both `ui/src/gateway/store_ops.rs` sites and both `ui/src/state.rs` sites now say version 0 is accepted only as the empty default. The summary entry size is 70 bytes, so about 280 KiB at the cap; the test's bound is now exact.

**Rebuilt WASM:** reputation moved to `58c9632d…`, and only because comment lines shifted a panic location. Store (`95f464c4…`), mailbox (`f78ff5a2…`) and the delegate (`da3c9ca1…`) are unchanged.

**verify-merge:** I re-ran the store corpora:
- store-v0: 0 violations; the injected states are refused.
- store: 1316/1316 held.
- store-adv: 6 violations, all #81.
- store-rr: 108 held, 1 inconclusive (no delta path).

Mutations: every new guard above was removed in turn and observed red.

Noted: the two policy calls (the order cap by age, and the mailbox caps 512/128/64/24) are signed off.

[AI-assisted - Claude]

sanity (commented):
## Final verdict: #82

**HEAD SHA:** `ad8318e639337100746129a4716da947662b8270`. CI is green (5/5), and the branch is based on current main.

**Review history:**
- **Round 1:** 5 lenses (code-first, testing, skeptical, big-picture, merge-law/migration). Found the seller-minted feedback slot and a missing byte-level canonical check.
- **Round 2:** 4 lenses. Found the unsigned version-0 store details.
- **Round 3:** 2 lenses. Found a stranded-seller regression and the migration fold carrying injected version-0 details.
- **Final check** of `f5791fe..ad8318e`: clean. Five of its own mutations were all caught, and the new tests fail on the previous commit.

Codex was over its usage limit throughout (resets 2026-09-19), so these substituted Claude lenses stand in for it, per the review rule.

**Merge laws:** `fdev verify-merge` was run over 20 corpora, including adversarial states at every cap, non-canonical encodings and injected version-0 details. The only remaining violations are the #81 states, which are unsigned fields that need a design decision. Seeded merge-law tests now run in `cargo test`.

**Policy:** sign-off on the order-cap and mailbox-cap choices is recorded above.

**Follow-ups:**
- Feedback entry-key requirements are recorded on #53.
- A separate CI PR will set `[profile.test] opt-level = 2`.

**State:** Ready to Merge.

[AI-assisted - Claude]

Issue #75, opened 2026-09-18
A buyer cannot publish the Paid transition, so an order only settles when the seller opens the app


sanity:
Found in review of #73.

## The design says anyone may publish it

`Paid` is authorized by **evidence**, not by a signature over the status. `common/src/payment.rs` classifies it as `Authority::BitcoinEvidence`, and says of the sibling status that it is "deliberately, unsigned -- evidenced by Bitcoin rather than by authority -- so anyone who can read the public order can submit one". The store contract adds no origin check.

`AppState::settled_orders` documents the same intent directly:

> **Why the buyer, and why anyone may** -- `Paid` is authorized by EVIDENCE, not by a signature over the status [...] so any reader holding the claims can publish the transition -- there is nothing here only the seller could sign. The buyer is simply the party who cares soonest.

## The app does not implement it

`store_ops::owned_store_key` resolves the store's `ContractKey` by looking the store up in `my_stores`, and fails with "this store is not one of yours" otherwise. A `ContractKey` is an instance id plus a code hash, and `store_ops::store_contract_key` says so itself: the instance id **is** the store contract id, and the code hash is the hash of the store contract this build bundles. Both are derivable without owning anything.

So the buyer's publish path has never worked. Before #73 it produced one failed send and one apology per order per session. #73 stops attempting it, because with the new re-reads that apology would have arrived every few minutes.

## Why it matters

Until this is fixed, **a paid, confirmed and attested order reaches `Paid` on the network only if the seller opens Harvest with the address view loaded.** If they never do, the published order says `AwaitingPayment` indefinitely, which is the public record being wrong about something that demonstrably happened.

The buyer's own screen is not affected: the order card reads the live address view and shows "Payment seen on chain" either way. That makes the gap silent -- it is visible only to third parties reading the store contract.

## What would fix it

Reconstruct the key from the store contract id plus the bundled code hash for evidence-authorized transitions, instead of requiring a `my_stores` registration. The signing-key requirement is correct for everything a seller signs (listings, orders); it is `Paid` specifically that needs neither.

Worth checking at the same time whether `settled_orders`' "any reader holding the claims can publish" paragraph should be kept as intent or narrowed, since it has described something the app did not do since it was written.

[AI-assisted - Claude]

PR #83, opened 2026-09-18
fix(payments): stop invoices settling against payments made before them, and stop reissuing addresses

### LINKED ISSUE #77: A fresh install reuses payment addresses from index 0, so a new invoice can be settled by an old payment
Found while re-taking screenshots on a clean peer tonight. This one is money, so filing before anything else.

## What happened

1. This morning, on one node, a store issued an invoice for 10,000 sats. Harvest derived `tb1qug28e25jzwak9uwhffug7t3dlltthnl6fsnf7a` from the seller's `vpub`, the buyer paid it, and the bridge attested it. That order reached `Paid`.
2. Tonight, on a **brand new node and a brand new browser profile**, the same Ghost Key connected, the store was re-created and the **same `vpub`** was pasted in. The payments panel read "0 invoice(s) have taken an address so far".
3. A new invoice was issued for a different listing. Harvest derived `tb1qug28e25jzwak9uwhffug7t3dlltthnl6fsnf7a` again -- the same address as step 1.
4. The new order went **straight to PAID**, settled by bridge `4MZnDAQW`, without anybody paying anything:

```
Your orders
Paid  0.00010000 BTC  PAID
Order iaZi2 . signet
Settled by bridge 4MZnDAQW
```

The address is confirmed identical: the invoice's own readonly field on the new install reads `tb1qug28e25jzwak9uwhffug7t3dlltthnl6fsnf7a`.

## Why it happens

`next_index` lives in `PaymentXpubStatus`, which is device-local delegate state. A device with no history starts at 0. The key is the same, so index 0 derives the same address, and that address already holds a confirmed payment with signed claims -- which is exactly what `assemble_on_chain_proof` is looking for.

Note the copy in the payments panel, which is true only where a count already exists:

> Entering a DIFFERENT key restarts the address count from zero, which is correct [...] Re-entering the key you already use keeps the count

On a reinstall there is no count to keep.

## Why it matters

A seller who reinstalls, moves to a new machine, or clears their browser data, and re-enters the same account key, will issue invoices against addresses they have already been paid on. Those invoices settle themselves. The seller then ships goods for money they received for something else, and the public order record says a payment happened that did not.

It needs no attacker. It is the ordinary consequence of restoring a wallet key, which is the one thing account-level public keys exist to make easy.

## Directions

The index cannot be authoritative on the device, because the device is the thing that gets replaced. Options, roughly in order of how much they fix:

- **Derive from something durable.** Scan forward from index 0 for the first address with no claims and no order naming it, using the address contracts already on the network, and take that as the next index. Slower to issue the first invoice, correct across devices, and needs no new state.
- **Record the high-water mark where it survives**, i.e. in the store contract or another published artifact rather than in delegate state.
- **Refuse to settle an order whose payment predates it.** An order carries an anchor; a claim carries `as_of` and a confirmation height. A payment confirmed before the order was signed should not settle it, whatever address it went to. This is worth doing regardless of the above, because it is the check that makes the failure impossible rather than unlikely.

The third is cheap and narrow and would have caught this exact case, since the payment confirmed hours before the order existed.

[AI-assisted - Claude]

sanity:
## Problem

A fresh install (new node, new browser profile) with the same wallet key restarted the delegate's address counter at 0, because `next_index` lives in device-local delegate state. It re-derived an address that already held a confirmed payment for an earlier invoice, and the new invoice went straight to Paid with nobody paying (#77). A seller who reinstalls or changes machine would ship goods against money received for something else.

## Approach

Three layers, revised after the Full-tier review below.

**1. Contract rule: a payment settles an order only inside that order's window** (`common/src/payment.rs`, `Order::payment_window`, enforced in `verify_on_chain_proof`, which the store contract runs through `AuthorizedOrder::verify`).
- An outpoint counts only if the fold's winning confirmation is strictly after the order's anchor and at most `PAYMENT_WINDOW_BLOCKS` after it.
- `PAYMENT_WINDOW_BLOCKS` = `MAX_ANCHOR_AGE_BLOCKS` (48, how long a buyer will still send) + 2016 (two weeks, Bitcoin Core's default mempool expiry, for a low-fee payment to confirm).
- The lower edge stops a reissued address's old payment settling a new invoice (the #77 case). The upper edge stops a new invoice's payment also settling an older unpaid invoice on the same address. It is per-order, so it stays merge-safe; a cross-order "no outpoint settles two orders" rule would let two individually valid peer states become invalid when merged.
- Judged by the fold's winning confirmation, so one stale-fork claim cannot veto an honest payment.
- An on-chain order with no anchor cannot be paid (`ProofError::NoAnchor`).
- What remains: two orders on one address whose anchors are closer than the window can both be settled by one payment in the overlap (pinned as a known limit); a payment broadcast before the order but unconfirmed when it was signed; one inside a stale anchor's lag; a pre-order transaction reorged and re-mined inside the window.

**2. The address counter follows the store's published orders.** `SetPaymentXpub` and `DeriveOrderAddress` carry `published_scripts`; the delegate moves its counter past the highest index of the stored key whose script appears there (never lowers it, ignores other keys, stops after 100 consecutive misses). The requests are built by `AppState::order_address_request` / `set_payment_xpub_request`, where tests execute them, and a derivation sends only scripts the delegate has not already answered for in this tab. The UI refuses to issue an invoice before the store has been read from the network.

**3. The derived address's own contract is checked before signing.** The UI GETs (no subscription) the bridge's address contract for the derived script. Any claim at all, even only a `ScannedTo` from an unpaid watch, means the address was on an invoice before: it is skipped and another derived (up to 20). The wait is bounded at 15 s and fails open, because a fresh address's usual answer is silence; the reasoning is on `AppState::on_address_reuse_timeout`. This catches what the published-orders record misses (a re-created store's own fresh state, pruned orders, unloaded sibling stores, orders past the scan gap), for any address registered with the bridge under the same bridge set and address-contract build.

**4. Ambiguous settlements go to the seller (UI only).** A provable payment is auto-published as Paid only when no other own, uncancelled order on the same address and network has an overlapping window (a Paid twin counts) and every owned store has loaded. Otherwise the card names the twin (or says a store has not loaded), says one payment cannot pay for both invoices, and offers "Confirm paid", which publishes the same assembled, verified proof; the contract accepts Paid on evidence alone. No payment is attributed between twins. This holds only because only the seller's tab publishes Paid today (#75 would bypass it). "Not found" from the address check signs, like the timeout, and is logged.

**UI copy.** The order card counts only value inside the order's window, shows "Payment seen on chain" only for the full amount ("Partial payment seen" otherwise), and tells the seller when the address holds a payment from before the invoice or after its window. The payments panel no longer claims re-entering the same key "keeps the count".

## Re-key

Store contract `3f47ab79` -> `c4e21292`, harvest delegate `08eac64a` -> `a1118c09`; reputation and mailbox unchanged; `StoreParameters` still 56 bytes. Outgoing hashes recorded in `legacy/` and `ui/src/migrate/tests.rs`. Intermediate builds on this branch (store `903a3529`, `eb18a826`; delegate `51c8b5ff`) were never published or merged and are not recorded. Verification got stricter, so a predecessor store holding an order settled outside its window no longer verifies and the fold discards it; the only known store is the test store, which holds exactly the two self-settled orders. The delegate re-key strands its secrets like every earlier one; re-entering the same key now recovers the count. `contract-rekey-acknowledged` is applied because re-keys were accepted for this fix while there are no users.

## Testing

Every new guard was checked by mutating it and confirming a test goes red; the list is in `docs/untested-invariants.md` (section "Address reuse and the payment window") and in the review-response comment below. Not executed by tests: the wasm-only GET/timer of the address check and its `NotFound` delivery, the card wiring of the warnings, and the wasm cost of the delegate scan. NotFound latency for a never-existing contract was spot-checked once on a throwaway peer: 7 of 8 answered in 4.3-7.1 s, one took 80.8 s.

Local: `cargo test --workspace`, `cargo clippy --workspace --all-targets -D warnings`, wasm32 clippy for the UI, the `address-guard` tests, `cargo check` of `tests/rehearsal`, and `scripts/check-code-hashes.sh` pass. Rebased onto `origin/main` after #80 (clean, no conflicts). Not published.

Fixes #77

[AI-assisted - Claude]


sanity (commented):
## Comprehensive PR Review: #83

### Summary
- **PR Title:** fixes #77 (a fresh install reuses payment addresses)
- **CI Status:** all 5 checks green at `bcce6a1`. The stale "no contract or delegate address moved" FAILURE in the rollup is the run from before the `contract-rekey-acknowledged` label was added; the rerun on the same SHA passed.
- **Review tier:** Full (money path, contract and delegate WASM, re-key)
- **Reviewers run:** code-first, testing, skeptical, big-picture, plus a dedicated money-correctness lens. **External model:** `codex review` was tried twice and failed both times with "You've hit your usage limit" (resets 2026-09-19). Per the review rule, we substituted the extra Claude lens instead of waiting.

### What holds
- The pre-anchor guard sits in `verify_on_chain_proof`. The store contract's `validate_state` goes through it, and so do the delta and merge paths, the UI's `settled_orders` and the buyer's view. All peers agree, because it uses no local clock or tip.
- The `<=` boundary is correct for honest flows. The testing lens mutated it independently and confirmed a test goes red.
- Dropping pre-anchor outpoints before the reversal bookkeeping closes the reversal-poisoning variant.
- The delegate's recovery never lowers the counter, ignores scripts from other keys, and saves before returning.
- The re-key bookkeeping and `docs/untested-invariants.md` are honest.

### Must Fix
1. **One payment can settle two orders.** The guard refuses payments confirmed *before* an order, but not a payment that arrives *after* an older order on the same address. Scenario: old order A is unpaid and recovery misses it (a re-created store's first state is its own empty PUT, a pruned order, a sibling store not yet loaded). A new order B reuses A's address. B's buyer pays, and both A and B verify as Paid. `settled_orders` publishes A too, because the node watches that address. This also makes false the claim at `delegates/harvest-delegate/src/bitcoin.rs:113-117` and in `legacy/` that "the residual is a reused address, not a wrongly-settled order". The fix needs to be per-order so it stays merge-safe. A validate_state rule of "no two Paid orders share an outpoint" would make two individually valid peer states unmergeable. A bound on how long after the anchor a payment may confirm is one merge-safe option.
2. **The seller's card still reads as paid on a reused address.** At `ui/src/components/bitcoin_view.rs:527-529`, an `AwaitingPayment` order with any confirmed balance on its address gets the `btc-pill paid` style and the text "Payment seen on chain". That balance ignores the anchor, so #77's harm ("the seller then ships") survives through this pill. Count only outpoints confirmed after `order.anchor`.
3. **The "store loaded" gate passes in the #77 flow itself.** `ui/src/state.rs:4140-4146` checks only `info.is_some()`, and the node's own freshly PUT empty state from store re-creation satisfies that. This PR's own re-key makes that the path everyone takes: the seller re-enters the xpub, the new store generation has no old orders, and index 0 is issued again. Suggested fix: before publishing an order, read the derived address's own address contract and skip any index that already has claims. That contract is the ground truth for reuse, and it also closes the pruned, sibling-store and beyond-gap residuals for any address that was actually paid.
4. **The wasm call sites that send `published_scripts` are untested.** Those are `spawn_order_address_request` and `save_payment_key`. Passing `Vec::new()` at either compiles, passes every test, and reintroduces #77. Build the request in a native-testable function and test that.

### Should Fix
5. **"Every confirmation" rule: honest payments can be stranded.** At `common/src/payment.rs:1178-1196`, one low-height claim can permanently veto an honest payment. That claim can come from a single trusted bridge on a stale fork, or from a small reorg that re-mines the tx at or below the anchor. Claims are grow-only and `assemble_on_chain_proof` must include them all. Meanwhile the rule protects only against honest submitters: a submitter can leave the old confirmation out (the proof is unsigned). Consider judging by the fold's winning confirmation instead. Also correct the comment "Failing closed there costs a reissued invoice": the buyer has already paid.
6. **`PaymentPredatesOrder` never reaches the seller.** `settled_orders` swallows every assembly error (`ui/src/state.rs:2958`). Surface it on the order card.
7. **The recovery scan runs on every invoice.** Scripts below the start index stay in the set, so every invoice pays for `PUBLISHED_INDEX_GAP` derivations. A fresh install against many orders could do hundreds of thousands, and the request can reach hundreds of KB. Send or keep only scripts that are not yet matched.
8. **Missing handler-level test** for a stale device re-entering the same key via `SetPaymentXpub`. Today only the pure function is tested.

### Consider
- The PR body says "as agreed on the issue", but the issue has no comments. Reword.

### Verdict
**State:** Needs Changes. Re-review is required after the fix, triggered by Must Fix 1 to 4.
**HEAD SHA reviewed:** bcce6a128b39b01f267439f64cd1556b1e0e6569

[AI-assisted - Claude]

sanity:
## Response to the Full-tier review (head `ded4d76`)

Rebased onto `origin/main` after #80 (clean, no conflicts). Each finding and its fix:

**Must Fix 1, one payment settling two orders** (`b1d89d1`). There is now a per-order payment window, `Order::payment_window`: a payment counts only if it confirmed strictly after the order's anchor and at most `PAYMENT_WINDOW_BLOCKS` = 48 + 2016 blocks later.
- **Width:** 48 is `MAX_ANCHOR_AGE_BLOCKS`, how long a buyer will still send. 2016 is two weeks, Bitcoin Core's default mempool expiry, as slack for a low-fee payment to confirm.
- **Merge-safety:** the rule is per-order, so two individually valid peer states cannot become invalid when merged.
- **What remains:** two orders on one address whose anchors are closer than the window can both be settled by one payment in the overlap. This is pinned by `known_limit_overlapping_windows_on_a_reused_address_both_settle`.
- **Doc claims:** fixed in `delegates/harvest-delegate/src/bitcoin.rs`, `legacy/store_contract.toml` and `docs/untested-invariants.md`.
- **Mutations, each red:** no upper edge; upper edge off by one.

**Must Fix 2, the paid pill on a reused address** (`714c36e`). `AddressReading` counts only confirmed value inside the order's window, using the same `payment_window` and the winning confirmation height the verifier uses. `status_pill` renders from that reading.
- **Mutation, red:** counting all confirmed value.

**Must Fix 3, the store-loaded gate passing in the #77 flow** (`efa77a5`). Implemented as suggested. Before signing, the UI reads the derived address's own contract with a GET that does not subscribe.
- **When an address is skipped:** any claim at all, even only a `ScannedTo`, means the address is skipped and another is derived, up to 20 in a row. An undecodable state also counts as used, and an address already watched on this node with claims is skipped without asking the network.
- **The wait:** bounded at 15 s. On timeout it fails open, deliberately: a fresh address's normal answer is silence, so failing closed would refuse nearly every invoice. The reasoning is in the comment on `on_address_reuse_timeout`. The published-orders floor is kept as the first line.
- **Mutations, each red:** inverting used/unused; undecodable treated as unused; no skip limit; bypassing the check; no shortcut for already-watched addresses.
- **Not executed by any test:** the wasm GET, the timer and the `NotFound` delivery.

**Must Fix 4, untested wasm call sites** (`f1561b1`). The requests are now built by `AppState::order_address_request` and `AppState::set_payment_xpub_request`. The wasm glue only calls them.
- **Mutations, each red:** sending `Vec::new()`; dropping the accounted-for filter.

**Should Fix 5, the every-confirmation rule** (`b1d89d1`). The window is now judged by the fold's winning confirmation. I agree with the review: the old rule could be bypassed by leaving the stale claim out, and it stranded honest payments. The comment that said "costs a reissued invoice" is corrected: that buyer has paid, so their payment is stranded. The concession, a pre-order transaction re-mined inside the window, is written up on `Order::payment_window`. Tests: `a_payment_is_judged_by_its_winning_confirmation` and `a_winning_confirmation_before_the_anchor_does_not_settle`.
- **Mutation, red:** restoring the every-confirmation rule.

**Should Fix 6, `PaymentPredatesOrder` never reaching the seller** (`714c36e`). The order card now says when the address holds a payment from before the invoice ("do not ship against it") or one after the window closed. This is computed from the same window as the verifier, rather than from `settled_orders`, which only ever sees successes.

**Should Fix 7, the scan running on every invoice** (`f1561b1`). A derivation sends only scripts the delegate has not already answered for in this tab. Setting a key sends all scripts and replaces the accounted set. Test: `a_script_the_delegate_has_answered_for_is_not_sent_again`.

**Should Fix 8, handler test for a stale device** (`f1561b1`). Added `a_stale_device_re_entering_the_same_key_takes_the_published_count`: through `handle`, the count goes from 1 to 5 and the next address is index 5.

**Consider, "as agreed on the issue".** The PR body is reworded.

**Re-key.** The store contract is now `eb18a826`. `903a3529` existed only on this branch and is not recorded as a generation. The V14 row, `3f47ab79` as shipped on main, describes the final change. The delegate is unchanged at `51c8b5ff`.

Codex was not re-run. Re-review is needed on this head.

[AI-assisted - Claude]

sanity (commented):
## Comprehensive PR Review: #83 (round 2, re-review after fixes)

### Summary
- **HEAD SHA reviewed:** `ded4d7660d62e938e3b45970104497bb2ab5eff1`. CI is green (5/5).
- **Review tier:** Full.
- **Reviewers run:** the same five blind Claude lenses as round 1: code-first, testing, skeptical, big-picture, and money-path correctness. **External model:** Codex is still over its usage limit (resets 2026-09-19), so this round uses the substituted Claude lenses instead, as the review rule requires.

### Round-1 findings
All eight are fixed as claimed. The code-first lens checked each one against the code, not just the commit messages. The contract layer holds:
- **The payment window** is `(anchor, anchor + 2064]`, judged on the fold's winning confirmation. It is identical in `payment_window`, the verifier and `AddressReading`.
- **Retractions:** a retracted pre-order payment cannot reverse a new order.
- **All validators** go through one deterministic function.

**Mutation checks this round:** the testing lens independently mutated the window's upper edge and the rule that any claim at all means an address is used. Both went red. Its third mutation (the skip-limit boundary) survived; see item 6.

### Should Fix
1. **Nothing warns the seller about two orders sharing one address.** The pre-sign address check can't close this: two devices invoicing at the same time both see a contract that doesn't exist yet, and a lookup that times out or wrongly says "not found" signs anyway. When it happens, one payment settles both orders (the pinned known limit) and nothing tells the seller. The robust backstop is local and merge-safe. When two orders in the seller's loaded stores share a `payment_script_pubkey` and their windows overlap:
   - flag both on the order card ("one payment cannot pay for both; do not ship either against a single payment"), and
   - have `settled_orders` decline to auto-publish Paid for either.
2. **Scripts get marked as handled when the delegate never matched them.** At `ui/src/state.rs:5525-5527`, every script sent is marked accounted once the delegate answers `Ok`. A published order beyond the delegate's 100-index scan gap is therefore never sent again. Before Should Fix 7 that miss was temporary; now it is permanent. Fix: the delegate returns the scripts it matched, and only those are marked.
3. **Setting the key doesn't wait for the stores to load.** `set_payment_xpub_request` has no loaded-stores gate, unlike `issue_invoice`. A tab that sets the key before its stores load can reset the delegate's count to 0 while another tab's accounted set hides the scripts that would lift it. Gate it the same way, or reset the accounted set whenever the delegate reports a `next_index` lower than last seen.
4. **"Not found" and late answers are not handled like the timeout.** `on_address_reuse_absent` (`ui/src/state.rs:4509`) signs immediately and logs nothing. In Freenet a GET that dead-ends can report NotFound for a contract that exists, so this is the same fail-open as the timeout, only faster and silent. Also, an answer showing claims that arrives after the timeout has signed is thrown away. Log NotFound as the timeout does, and for a period after a timeout still accept a late "used" answer and turn it into a notification or a flag on the order.
5. **Dust lights the paid pill.** At `ui/src/components/bitcoin_view.rs:617`, `in_window_sats > 0` gives the Paid style. Anyone can send 546 sats to a published address. Use the paid style only when `in_window_sats >= amount_sats`; otherwise show "partial payment seen".
6. **Small fixes:**
   - Pin the skip-limit boundary with a test (the `>` vs `>=` mutation at `state.rs:4547` survives), and check the "20" count in the copy.
   - Remove the `address_skips` / `scripts_in_flight` entries on the derive-failure path.
   - Stop `address_reuse_checks.insert` from silently replacing an existing check.

### Consider
- **Wording:**
  - The after-window note should say the payment is probably this buyer's late payment, to settle with them directly.
  - The "stopped trying" doc at `payment.rs:309-315` is inaccurate: wallets rebroadcast after mempool expiry.
  - The pre-order note shouldn't tell a seller to reissue when a valid in-window payment is also present.
  - Anchorless orders get the "before this invoice" note when the real reason is NoAnchor.
- **Record as concessions:**
  - A Paid order whose winning confirmation is later reorged out of the window can no longer be reversed.
  - `payment_window` saturates at `u32::MAX` (a `checked_add` would be more honest).
  - The pill can disagree with the verifier when the address view's claims are cut at 32.
- **Question for the author:** does a GET for a never-existing contract return NotFound promptly on a live node? If not, every invoice waits 15 s. That should be measured once on a live peer.

### Verdict
**State:** Needs Changes. Items 1-5 touch the money path and the seller's decision to ship, so a focused re-review of the fix commits follows.

[AI-assisted - Claude]

sanity:
## Response to the round-2 review (head `3f583a2`)

No rebase was needed; `origin/main` has not moved since `ded4d76`. Each finding and its fix:

**Should Fix 1, nothing warns about two orders on one address** (`5587ab4`). `AppState::orders_sharing_an_address` flags every pair of the seller's OWN loaded orders that share a payment script and whose windows overlap.
- Both cards say one payment cannot pay for both, and `settled_orders` does not auto-publish Paid for either.
- It is local and writes nothing, so it is merge-safe. It covers the two-device race and every pre-sign lookup miss.
- Own stores only, so a stranger copying a script cannot stall this seller's settlements.
- Tests: `orders_share_an_address_only_when_their_windows_overlap` (both window boundaries and the stranger case) and `a_payment_on_an_address_two_own_orders_share_is_not_auto_published`.
- Mutations, each red: settling regardless; a strict overlap comparison; counting every browsed store.
- Limit: a twin in a store that has not loaded is invisible, and a flagged order is left for the seller to settle by hand.

**Should Fix 2, unmatched scripts marked as handled** (`d358060`). The delegate now returns `matched_scripts` on `OrderAddress` and `PaymentXpubSet`, and the UI marks only those as accounted for.
- A script that was sent but not matched is offered again once the counter has advanced by `PUBLISHED_INDEX_GAP / 2`. That is when a beyond-gap order can have come within the scan's reach, and it keeps foreign and below-counter scripts from being resent on every invoice.
- `PUBLISHED_INDEX_GAP` moved to `harvest-common`.
- Tests: `the_delegate_reports_only_the_scripts_it_matched` and `only_scripts_the_delegate_matched_are_not_sent_again`.
- Mutations, each red: the delegate reporting everything it was sent; the UI marking every sent script.

**Should Fix 3, setting the key before stores load** (`d358060`). Did both. `set_payment_xpub_request` refuses until every owned store has loaded, and the accounted record is cleared whenever the delegate reports a lower count or a different key.
- Tests: `a_payment_key_waits_for_every_owned_store_to_load` plus the falling-count step in the accounting test.
- Mutations, each red: removing the gate; removing the reset.

**Should Fix 4, NotFound and late answers** (`5587ab4`). `on_address_reuse_absent` now logs and signs like the timeout, and both remember the invoice. A later answer showing use flags the order (`AddressWarning::UsedBefore`) and notifies the seller.
- Listening stops once the node watches the address as an order's. Otherwise the bridge's claims for the invoice's own registration would read as reuse.
- Test: `a_late_answer_showing_use_flags_an_invoice_already_signed`.
- Mutations, each red: not remembering on NotFound; removing the watch cutoff.

**Should Fix 5, dust lighting the pill** (`5587ab4`). The paid style now needs `in_window_sats >= amount_sats`. Less than that reads "Partial payment seen".
- Test: `dust_inside_the_window_reads_as_partial_not_paid`.
- Mutation, red: lighting the pill for any in-window value.

**Should Fix 6, small fixes** (`d358060`, `5587ab4`).
- The skip-limit boundary is pinned (`one_short_of_the_skip_limit_still_derives_again`), and the give-up message now reports the number of addresses actually tried.
- The failure paths clear the skip count and in-flight scripts (`abandon_address_request`, with `a_refused_derivation_forgets_everything_held_for_it`; red with the removal dropped).
- A second invoice handed an address that is already under check is skipped instead of replacing that check (`a_second_invoice_on_an_address_already_being_checked_is_skipped`; red without the check).

**Consider, wording.**
- The after-window note calls the payment the buyer's probable late payment, to settle with them directly.
- The mempool-expiry doc now says wallets rebroadcast and that the bound is chosen, not natural.
- The pre-order note does not tell the seller to reissue when a full in-window payment is also present.
- An anchorless order gets its own note.
- Pinned by `each_note_names_its_own_reason`; the reissue branch mutation goes red.

**Consider, concessions.** Rows are added to `docs/untested-invariants.md` for three things: a Paid order reorged out of its window cannot be reversed; `payment_window` saturates at `u32::MAX`; the pill can disagree with the verifier past the 32-claim cap.

**Question, NotFound latency.** Spot-checked once on a throwaway peer (WS 7661, connected to the nova gateways, killed by PID afterwards). I ran `fdev execute get` on 8 random never-existing ids: 7 returned NotFound in 4.3-7.1 s, and 1 took 80.8 s. So the 15 s wait usually ends on NotFound, and occasionally a fresh invoice waits the full 15 s. This is one peer on one sitting, not a distribution. It is recorded in `docs/untested-invariants.md`.

**Re-key.**
- Store contract `eb18a826` -> `c4e21292`. The only changes are harvest-common doc and delegate wire types; there is no verification change.
- Delegate `51c8b5ff` -> `a1118c09`, for `matched_scripts`.
- Both outgoing builds existed only on this branch and are not recorded as generations. The V14 and V13 rows describe the final change.

CI is green at `3f583a2` (5/5). Codex was not re-run; a focused re-review of the fix commits is still needed.

[AI-assisted - Claude]

sanity (commented):
## PR Review: #83, round 3 (re-review of `d358060` and `5587ab4`)

**HEAD SHA reviewed:** `3f583a27a206fd495dd20e26fb29327b49ba5940`. CI is green. Locally, `cargo test --workspace --locked` passes 809/809 and the rehearsal crate compiles.
**Lenses:** skeptical, money-path correctness and testing (Claude). The contract verification logic did not change this round, so the lenses focused on the fix commits and read the full PR for context. **External model:** Codex is still over its usage limit (resets 2026-09-19), so the substitute Claude lenses are used instead.

**Tests:** the testing lens made 11 independent mutations and all 11 were caught. Every round-2 guard is covered by a test. What follows is about design, not coverage.

### Should Fix: simplify rather than add

The contract layer (the payment window) has been sound since round 2. Each UI mitigation added since then has brought new problems, so this round removes machinery that can't do its job and narrows what's left.

1. **The overlap hold-back can never lift, which strands honest payments.**
   - **Where:** `orders_sharing_an_address` (`ui/src/state.rs:3119-3153`) groups orders regardless of their status, and nothing ever clears the result.
   - **Why that matters:** `settled_orders` is the only way an order becomes Paid, and there is no cancel or manual-Paid action.
   - **Scenario:** the card says "reissue one of them", but the dead twin stays in the store. The buyer of the other order pays in full, and their payment never reaches the public record.
   - **Fix:**
     - Hold back only while the confirmed total inside the window can't cover the combined amounts of the overlapping twins that are still `AwaitingPayment`.
     - Ignore twins on a different `order.network`: signet and testnet4 share script bytes.
     - Word the card text by status. For an order already Paid, say it may have been marked paid on the other order's payment. Don't say "by itself", which implies a manual route that doesn't exist.
     - `status_pill` should never use the paid style while the warning is showing.
2. **Remove the late-answer / `UsedBefore` machinery.**
   - It can't fire after NotFound, because nothing sends the GET again. NotFound is the usual answer (7 of 8 measured).
   - Its timeout window is only a few seconds.
   - The flag lives only in memory.
   - It can be a false positive from the invoice's own registration.
   - Its test delivers a state that can't happen in production.
   
   Keep the NotFound logging. Fix the docs and the comment at `response_handler.rs:123-124`, which overstate what it catches.
3. **Drop the loaded-stores gate on setting the payment key** (`state.rs:4481-4492`). A single owned store that never loads blocks the seller from setting or fixing their key, permanently and without naming the store. And the gate buys almost nothing: with the same key the counter is kept, and scripts from a store that wasn't loaded are sent on the next derive anyway.
4. **Regression: the after-window note.** It now says a payment is "probably this buyer's late payment" (`bitcoin_view.rs:755-757`) even when the payment falls inside a later own order's window on the same address. Say "probably" only when no other own order's window contains that height.
5. **Key the reuse-check timer to its own check.** At `state.rs:590-601`, invoice 1's timer can time out invoice 2's newer check for the same contract id.
6. **Compute the overlap map once per state change**, not on every card render and every address-state arrival (currently O(N²)).
7. **Docs:**
   - The overlap hold-back exists only in the seller's UI. #75 (letting buyers publish Paid) would remove this backstop.
   - A twin order in a store that hasn't loaded yet is not seen.

### Consider
- Add tests for the two untested branches of `forget_accounted_scripts_if_counter_fell` (a key change, and the key being cleared).

### Verdict
**State:** Needs Changes. Items 1-4 get a focused re-check of the fix commits.

[AI-assisted - Claude]

sanity:
## Response to the round-3 review (head `43935e1`)

All seven items are in one commit, `43935e1`. The delegate and the contracts did not change this round, so no hashes moved. `origin/main` has not moved, so there was no rebase.

1. **Hold-back that can lift.** `refresh_shared_address_warnings` replaces `orders_sharing_an_address`.
   - An order awaiting payment is held (`AddressWarning::Held`) only while the confirmed value inside its own window can't cover its amount plus every overlapping twin still `AwaitingPayment`. Once it can, the hold is released.
   - Twins on another network are ignored.
   - A paid or reversed order gets `MayBeTwinsPayment` only: "this one may have been marked paid on the other invoice's payment". The copy no longer says "by itself".
   - `status_pill` never uses the paid style while a warning shows.
   - Tests: `shared_address_warnings_hold_until_the_address_covers_the_twins` (both window boundaries, hold then release, a paid twin, another network, a stranger's store), and the buy-flow test, which is now held and then released through `settled_orders`.
2. **Late-answer machinery removed.** The listener, the `UsedBefore` flag, `signed_unchecked` and their test are gone. NotFound still logs. The `response_handler.rs` comment now says absence is not proof. The doc row is rewritten, and the latency measurement stays in it.
3. **Key gate removed.** `set_payment_xpub_request` is back to a plain request builder. The reset of the accounted set is kept. Its two untested branches are now covered by `a_new_or_cleared_key_forgets_what_was_accounted`: a different key at a count that did not fall, and the key being cleared.
4. **After-window note.** It says "probably this buyer's late payment" only when no other own order's window on the same address and network contains that height. Otherwise it says the payment may be the other invoice's. Tests: `another_own_orders_window_is_found_by_height` and the note test.
5. **Timer keyed to its own check.** `PendingReuseCheck` carries the derivation's `request_id`, and `on_address_reuse_timeout(contract_id, request_id)` ignores a timer from an earlier check. Test: `a_stale_timer_does_not_end_a_newer_check_on_the_same_address`.
6. **Overlap map computed once per state change.** It is stored in `shared_address_warnings` and recomputed on store states, address states and `StoreList` answers. The card and `settled_orders` read it.
7. **Docs.** A row in `docs/untested-invariants.md` says the hold-back is UI-only and that #75 would bypass it, and that a twin in a store that hasn't loaded isn't seen.

**Mutations this round, 11 in total, all red after one test was strengthened:**
- **Hold-back:** never releasing; counting paid twins in the amount needed; ignoring the network; `settled_orders` ignoring the hold; strict overlap.
- **Card:** the pill still paid while a warning shows; the late note ignoring other windows.
- **Other-window check:** ignoring the script. This one survived at first. I added a different-address order to the test, and then it went red.
- **Timer and reset:** a stale timer ending a newer check; a key change not forgetting the accounted set; a cleared key not forgetting it.

**Net diff this round:** +440 / −317 lines (net +123) over 5 files. This goes against the hope that the net count would fall. Most of the growth is tests: the new hold/release cases plus the timer and reset tests. The hold/release calculation also replaced a simpler overlap check. I did not trim tests to get the number down.

CI is green at `43935e1` (5/5). Codex has not been re-run. A focused re-check of items 1-4 is still pending.

[AI-assisted - Claude]

sanity (commented):
## PR Review: #83, round 4 (re-check of `43935e1`)

**HEAD SHA reviewed:** `43935e1e14df75fbda41846a4df1b7806f846f23`. CI is green.
**Lenses:** skeptical and money-path (Claude). Codex is still over its usage limit (resets 2026-09-19), so these Claude lenses stand in for it.

**Clean:** removing the late-answer machinery, keying the timer to its own check, network separation, and ignoring stores the seller doesn't own.

### Blocking
1. **One payment can settle two orders once a twin is Paid.**
   - **What's wrong:** at `ui/src/state.rs:3163`, `needed_sats` leaves out twins that are no longer AwaitingPayment. But `in_window` still counts the payment that settled the Paid twin.
   - **Common path:** twins sit in two stores. After a reload, the first store to load publishes A with no twin in view. When the second store loads, B is released on the same payment.
   - **The test enshrines it:** `a_payment_on_an_address_two_own_orders_share_is_not_auto_published` asserts this double settlement as correct. My round-3 wording ("still AwaitingPayment") led here.

### Should Fix
2. **Deadlock with partly overlapping windows.** Two buyers each pay in full, but in parts of the windows the other order doesn't share. Both orders stay Held forever.
3. **An abandoned twin strands the buyer who paid.** Nothing expires the abandoned order, and the seller has no cancel or manual-Paid action.
4. **The paid-twin warning never clears.** `MayBeTwinsPayment` stays up for legitimately paid pairs and for Cancelled twins.
5. **A twin in a store that hasn't loaded fails open.** No twin in view means no hold.

### Direction for round 5
Stop trying to work out from chain data which order a payment belongs to. When windows overlap, that can't be decided, which is why each round of arithmetic has broken. Instead:
- **Auto-publish Paid only when nothing is ambiguous.** That means no own twin that isn't Cancelled, on the same script and network, with an overlapping window, **and** every owned store has loaded.
- **Otherwise, a seller-confirmed publish.** When a valid proof assembles, show a "Confirm paid" action next to a warning that names the twin. It publishes the same verified proof. The contract accepts Paid from anyone holding valid evidence, so this needs no new trust. It gives every stranded case a way out, and the seller makes the call with the warning in front of them.

[AI-assisted - Claude]

sanity:
## Response to the round-4 review (head `ebd366a`)

Everything is in one commit, `ebd366a`. There was no rebase (`origin/main` has not moved) and no hash moved (UI and docs only).

**The UI no longer tries to work out which order a payment belongs to** (blocking 1, Should Fix 2-5):
- **Removed:** the `needed_sats` / `Held` / release arithmetic and `AddressWarning`.
- **When an order is ambiguous:** `refresh_same_address_orders` records the seller's other own orders that are not Cancelled and sit on the same script and network. `settlement_hold` calls an order ambiguous when any of them has an overlapping window, whatever its status; Paid twins count.
- **When Paid is published automatically:** `publish_settled_orders` does it only when the order is not ambiguous and every owned store has loaded. That gate is on publishing only; setting the key has no gate.
- **The withheld case:** the assembled, verified proof is kept, and the card shows the twin's id (or that a store hasn't loaded). It says "one payment cannot pay for both invoices" and offers **Confirm paid**. `confirm_paid` publishes that same proof. The contract accepts Paid on evidence alone, so this asks for no new trust.
- **Paid orders:** an informational note, with no pill change, shows only when the order's own in-window payment height lies inside another own order's window. Cancelled twins are ignored everywhere, including `another_own_order_window_holds`, which now reads the cached map.

**Tests:**
- The test that asserted the double settlement is gone. It is replaced by `a_twin_withholds_the_settlement_and_the_seller_can_confirm_it`: a Paid twin and an abandoned awaiting twin each withhold, Confirm paid publishes exactly the assembled proof and it verifies, and a Cancelled twin does not count.
- `an_owned_store_not_loaded_withholds_the_settlement`.
- `an_unambiguous_payment_is_auto_published`.
- `an_order_is_ambiguous_with_any_overlapping_uncancelled_own_twin`: partial overlap, both window boundaries, paid, cancelled, another network, a stranger's store, plus the by-height lookup and the Paid-note condition.
- The partial-overlap and abandoned-twin cases now end with the seller's confirmation rather than being stuck.

**Mutations, 11, all red:**
- **Withholding and confirming:** publishing despite a hold; confirm publishing nothing; dropping the loaded-stores gate.
- **Which twins count:** excluding Paid twins; counting Cancelled ones; ignoring the network; strict overlap; counting a stranger's stores.
- **Paid note:** the note ignoring payment heights; heights not recorded.
- **Pill:** the paid pill shown while confirmation is awaited.

**Net diff this round:** +386 / −352 (net +34) over 3 files. Most of the removed arithmetic went, but the Confirm paid path and its tests added code.

**Limits, recorded in `docs/untested-invariants.md`:**
- UI-only. #75 (letting buyers publish Paid) would bypass it.
- A twin in a store that hasn't loaded is not seen, which is why an unloaded store withholds every settlement.
- The card wiring (the button and the notes) is a component and is untested.

CI is green at `ebd366a` (5/5). Codex has not been re-run.

[AI-assisted - Claude]

sanity (commented):
## PR Review: #83, round 5 (re-check of `ebd366a`)

**HEAD SHA reviewed:** `ebd366a6a5be42a1531908dc10eca18234a37e5f`. CI is green.
**Lenses:** skeptical and money-path (Claude). Codex is still out of quota, so these Claude lenses stand in for it.

**Blocking:** none. The round-4 double settlement is fixed. A twin that's already Paid now counts. No path auto-publishes a twin this tab can see. The ordinary one-store sale still settles automatically, as promptly as before.

### Should Fix (small, then a focused re-check before merge)
1. **Confirm paid copy.** `SettlementHold::explain` (`ui/src/state.rs:1030-1047`) should show:
   - this invoice's amount,
   - each twin's amount and status, with Paid marked,
   - the confirmed value inside the window.
   
   Without the status, a seller can confirm an abandoned twin on a payment that already settled its sibling, and ship twice.
2. **The loaded-stores gate only covers identities this tab has heard from** (`owned_stores_loaded`, `state.rs:3177`). Hold while any Ghost Key identity the vault lists has no `StoreList` answer yet, or its request failed. Otherwise the docs claim more than the code does.
3. **A store that gave up.** A store in `store_state_unavailable` holds back every settlement for the whole session, and the note says to wait. Either name the store(s) in the note, or stop counting a store that gave up as "not loaded" (the Confirm button still covers it). Decide which, and say why.
4. **Nothing re-publishes when a hold clears.**
   - **Where:** the StoreList arm and the store-state arm publish only for the store that arrived (`state.rs:1974`, `:5113`).
   - **Effect:** other stores' withheld entries show a paid-style pill with no button, until the next address re-read (up to 5 minutes).
   - **Fix:** re-run the publish for every owned store that has a withheld entry.
5. **The Confirm button comes back after a successful confirm.** The next publish run re-inserts the entry (`state.rs:3319-3324`). Skip ids in `settlements_submitted`, or show "submitted".

### Consider
- Clear withheld entries for store ids that were migrated away.
- The Paid-order note: use warning styling and name the twin.
- A Cancelled twin is excluded even if someone paid it anyway. Only a foreign client can cancel today; add a docs row.

[AI-assisted - Claude]

sanity:
## Response to the round-5 review (head `1927fa0`)

All of it is in one commit, `1927fa0`. There was no rebase and no hash moved (UI and docs only).

1. **Confirm paid copy.** `SettlementHold::explain(amount_sats, in_window_sats)` now gives this invoice's amount, the confirmed value inside its window, and each twin as "id for N sats (status)". A Paid twin shows as "ALREADY PAID". To carry that, the cached map holds `SameAddressOrder { id, window, amount_sats, status }`. Test: `the_confirm_copy_names_the_amounts_and_a_paid_twin`.
2. **Identities.** `store_lists_answered` records every `StoreList` answer. `unloaded_stores` names any vault-listed Ghost Key that has no answer, which also covers a request that failed, and every owned store whose state hasn't arrived. `settlement_hold` returns `StoresNotLoaded(names)`. Test: `an_unanswered_store_list_holds_until_it_arrives_then_publishes`.
3. **A store that gave up.** I kept counting it and name it as "did not answer", rather than dropping it. It could hold a twin, and auto-publishing past a possible twin can't be taken back, whereas the Confirm button already gives the seller a way out. The reasoning is in the comment on `unloaded_stores`. Test: `a_store_that_did_not_answer_is_named`.
4. **Republish when a hold clears.** `republish_withheld` re-runs the publish for every store that has a withheld entry. It is called on each `StoreList` answer and each store state. The same test as item 2 checks that the settlement goes out as soon as the answer arrives.
5. **Not offered again.** The publish run skips ids already in `settlements_submitted`. Test: `a_confirmed_settlement_is_not_offered_again_and_migration_drops_entries`.

**Consider items.**
- `adopt_migrated_contract_id` drops withheld entries for the predecessor store id (same test as item 5).
- The Paid-order note now uses warning styling and names the twin(s), via `orders_whose_window_holds`, which replaces the boolean helpers.
- A concession row in `docs/untested-invariants.md` records that a Cancelled order is excluded even if someone paid it.

**Mutations, 8, all red:** dropping the Paid mark; no identity gate; the answer not recorded; no republish on a store list; an unanswered store not named; a confirmed settlement offered again; migration keeping entries; the window lookup ignoring heights.

CI is green at `1927fa0` (5/5). Codex has not been re-run.

[AI-assisted - Claude]

sanity:
## Final re-check Lows (head `cdcb499`)

Everything is in one commit, `cdcb499`: +73 / −2 lines across `ui/src/state.rs` and `docs/untested-invariants.md`. No hash moved.

1. **A twin confirmed from this tab.** `settlement_hold` now sets `confirmed_by_you` on twins whose id is in `settlements_submitted`. `explain` labels them "confirmed by you, publishing" instead of "awaiting payment". Test: `a_twin_confirmed_here_is_labelled_as_such`. Red when the flag is not set.
2. **A failed store-list request.** The failure is recorded in `store_list_failed`. The hold then names that Ghost Key's store list as "whose request failed; reloading the page retries it". There is no retry loop. Test: `a_failed_store_list_request_says_so`. Red when the failure is not named.
3. **Docs.** Added a concession row: a submit that returns Ok but whose Paid never arrives stays in `settlements_submitted` for the session, so the hold text and the button stay hidden until a reload.

CI is green at `cdcb499` (5/5).

[AI-assisted - Claude]

sanity (commented):
## Final verdict: #83

**HEAD SHA:** `cdcb499a5838fa6d84c571b30eac0c2e0c04d456`. CI is green: 5/5 checks.

This is a light check of the last commit (`1927fa0..cdcb499`, +73/−2, `ui/src/state.rs` and docs only). It closes the three Low findings from the final re-check:
- a twin confirmed from this tab is labelled "confirmed by you, publishing";
- a store-list request that failed to send is named as failed, with how to retry;
- the submitted-but-never-arrived case is recorded as a concession in `docs/untested-invariants.md`.

**Review history:** six rounds.
- Round 1: five Claude lenses.
- Round 2: five lenses again.
- Rounds 3-6: focused two- and three-lens re-checks of each fix.

Codex was over its usage limit throughout (resets 2026-09-19), so these substituted Claude lenses ran instead, per the review rule. Every finding is fixed or recorded as an accepted concession in `docs/untested-invariants.md`.

**State:** Ready to Merge.

[AI-assisted - Claude]

PR #86, opened 2026-09-18
fix(ui): self-host fonts and stop wasm-opt aborting silently

### LINKED ISSUE #28: ui: harvest.css imports Google Fonts, which CSP blocks — the app never gets its typefaces
Found by publishing Harvest and loading it for the first time. Not visible in any test.

## What happens

`ui/assets/harvest.css:4`:

```css
@import url('https://fonts.googleapis.com/css2?family=Libre+Baskerville:ital,wght@0,400;0,700;1,400&family=Source+Sans+3:wght@300;400;500;600&display=swap');
```

The published app is served with `default-src http://127.0.0.1:7509 'unsafe-inline' 'unsafe-eval' blob: data:`, so the browser refuses it:

> Loading the stylesheet 'https://fonts.googleapis.com/css2?…' violates the following Content Security Policy directive: "default-src http://127.0.0.1:7509 …". The action has been blocked.

So Harvest renders in fallback fonts, always, for everyone. The design specifies Libre Baskerville and Source Sans 3 and gets neither.

## Why fixing it by loosening CSP would be wrong

The obvious fix is to allow `fonts.googleapis.com` in the CSP. **Don't.** Every visitor's browser would then make a request to Google on every load of a decentralised app — announcing to a third party who is using Harvest and when. That is a worse outcome than the wrong typeface, and it is the kind of thing a privacy-oriented project should not ship by default.

**Bundle the fonts as assets instead**, served from the app's own origin, which the CSP already permits.

## Why no test caught it

Nothing in the suite loads the app in a browser against a real CSP. Every existing check is a unit or integration test in Rust, and the wasm-gated UI layer is type-checked and linted but never executed. This is the first defect found by *running* the published app rather than by reasoning about it, and it took about ninety seconds.

Worth considering a smoke check that loads the published app headless and fails on any CSP violation in the console — cheap, and it catches a class nothing else can see.

[AI-assisted - Claude]

### LINKED ISSUE #34: build: wasm-opt aborts on every UI build and dx reports success anyway, so the shipped wasm is unoptimised
## What

`wasm-opt` fails on every UI build, and `dx` continues anyway and reports success. The published app therefore ships an **unoptimised** UI WASM.

From a clean `cargo make compress-webapp` run today:

```
100.893s ERROR wasm-opt failed with status code signal: 6 (SIGABRT) (core dumped)
stderr:
stdout: compile unit size was incorrect (this may be an unsupported version of DWARF)

100.926s INFO Client build completed successfully! 🚀
```

Note the two lines are 33 ms apart: the optimiser aborts with SIGABRT and the build is then declared successful. Nothing downstream reports a problem.

## Evidence it has been failing all along, not just today

`assets/harvest-ui_bg-<hash>.wasm` is **7,283,353 bytes** in today's build — byte-for-byte the same size as the build published this morning, whose contents differ (different content hash, since the Bitcoin constants changed). Two builds with different content and an identical byte count is what an un-run optimiser looks like; an optimiser that ran would be very unlikely to land on the same total twice.

## Why it matters

7.3 MB of unoptimised WASM is fetched by every user who loads Harvest, from a network where bandwidth is a tracked concern (`freenet-core#5153` measures the receive path in exactly these terms). `wasm-opt -Oz` on a Rust/wasm-bindgen artifact of this shape typically removes a substantial fraction; whatever the real figure, it is currently zero and nobody is being told.

The secondary cost is the shape of the failure rather than its size: a toolchain step aborts with a core dump and the build says "completed successfully". That is the same class of thing as the other guards fixed in this repo today — a success message not conditioned on success — except here it belongs to `dx` rather than to us.

## Cause

The message names it: `compile unit size was incorrect (this may be an unsupported version of DWARF)`. `wasm-opt` is parsing DWARF debug info in the release artifact and choking on the version rustc emitted. So the fix is on our side rather than requiring a new `wasm-opt`: release builds of the UI do not need debug info at all.

Likely one of:

- `[profile.release] debug = false` / `strip = "debuginfo"` (or `strip = true`) for the UI build;
- have `dx` pass `--strip-debug` / disable DWARF preservation to `wasm-opt`;
- a `binaryen` version bump, if the DWARF version is genuinely newer than it supports.

The first is cheapest and most likely correct, since debug info in a shipped webapp WASM is not wanted anyway — it is both the cause of the abort and a chunk of the 7.3 MB.

## Suggested acceptance

Not "wasm-opt no longer errors" alone — that could be achieved by silencing it. Require both:

1. no `wasm-opt failed` line in the build log, **and**
2. the emitted `assets/*_bg-*.wasm` is materially smaller than 7,283,353 bytes,

and record the new size, so a future regression to "optimiser silently not running" is visible as the number going back up. A check that only greps for the error string would pass the moment the message changes.

## Not urgent

Nothing is broken for users beyond download size, and it predates today's work. Filed so it is not lost — found while verifying the bundle gate in #33, which is what made the build log worth reading closely.

[AI-assisted - Claude]

sanity:
## Problem

Two independent build/UI defects.

**#28**: `harvest.css` imported Libre Baskerville and Source Sans 3 from `fonts.googleapis.com`. The published app's CSP (`default-src http://127.0.0.1:7509 ...`) blocks that request outright, so the app has always rendered in fallback fonts.

**#34**: `wasm-opt` aborted (SIGABRT) on every UI release build with `compile unit size was incorrect (this may be an unsupported version of DWARF)`, and `dx` swallowed the failure: it logged an ERROR line, fell back to shipping the unoptimized wasm-bindgen output, and still reported the build successful. The shipped wasm has therefore never been optimized.

## Approach

**#28**: bundle the two font families (7 static weights/styles actually used, OFL-licensed, latin subset) under `ui/assets/fonts/` and reference them from `harvest.css` via `@font-face`, served from the app's own origin. Widening the CSP to allow `fonts.googleapis.com` was the other option, but that would make every visitor's browser fetch from Google on every load, leaking who is using Harvest and when — a bad trade for a privacy-oriented app. Self-hosting needs no CSP change at all, since the app's own origin is already permitted by `default-src`.

`check-webapp-bundle.sh`'s `REQUIRED` list and the `build-ui` Makefile task are updated so the 7 fixed-name font files are copied into the release bundle and accepted by the publish gate.

**#34**: root cause is `dx`'s own `--debug-symbols` flag, which defaults to `true` regardless of `--release`. With it on, `wasm-bindgen` keeps the DWARF debug info rustc emitted, and `dx` asks `wasm-opt` to *rewrite* it (`--debuginfo`) rather than drop it. This rustc's DWARF output crashes that rewrite path in the binaryen version `dx` bundles. `build-ui` now passes `--debug-symbols false`, so `wasm-opt` is asked to `--strip-debug` instead — nothing left to crash on, and the optimizer actually runs.

Separately, this build's wasm never carries the wasm-bindgen "name" section either way, with or without `--debug-symbols` — verified by forcing it back to `true` and inspecting the (unoptimized, since wasm-opt still aborted) output. `dx` only retains names for `--wasm-split` or hotpatch ("fat") builds, neither of which this is, so panic traces in production already show `wasm-function[N]` rather than Rust symbol names regardless of this flag. That's accepted for production; `--debug-symbols` only controls the DWARF debuginfo, which is what was actually breaking `wasm-opt`.

Because `dx build` exits 0 whether or not `wasm-opt` actually ran, `cargo make` could not previously tell a real optimization pass from the silent fallback. `build-ui` now captures `dx`'s own output and fails the task if the `wasm-opt failed` signature ever reappears, so this can't silently regress to shipping unoptimized wasm again.

### Review follow-up: the grep guard depends on dx's exact wording, and CI never runs `dx`

Two reviewers flagged that `grep -q "wasm-opt failed"` depends on an internal log string from dioxus-cli 0.7.3 (`src/wasm_opt.rs`), not a documented interface, and nothing pinned the `dx` CLI version or ran it in CI — so a `dx` upgrade that rewords the message would silently disarm the guard. That's exactly the false-green shape #34 itself warned about ("a check that only greps for the error string would pass the moment the message changes").

Fixed by adding a version check at the top of `build-ui`: before building, it runs `dx --version` and fails loudly (with a message pointing back at this PR and #34, and instructions to re-verify) if it isn't exactly `dioxus 0.7.3`. Verified this fires: temporarily changed the expected version, confirmed `cargo make build-ui` fails fast (before the expensive `dx build` even starts) with a clear error, then restored it.

**How #34's size criterion is covered, given there's no ongoing size gate:** #34 asked for two things — no `wasm-opt failed` in the log, and the emitted wasm materially smaller than the broken baseline (7,283,353 bytes at the time #34 was filed). The version pin plus the log grep cover the first directly. For the second, there's no per-build size assertion in this PR, only the one-time measurement recorded above (7,685,419 → 7,038,487 bytes on this tree, confirmed via `wasm-objdump` that the `.debug_*` sections are gone). Codifying "no `.debug_*` custom sections" as an ongoing check would be redundant with what the version pin + grep already guarantee: given dx 0.7.3 and a `wasm-opt failed`-free log, we've already established that `wasm-opt` ran with `--strip-debug` and completed, which is the only way debug sections could be absent. Recording an absolute byte threshold instead would be brittle in the other direction — legitimate future code changes shift the number constantly, and CI doesn't build the UI wasm as a `dx` release bundle at all today (the `wasm` CI job builds contracts, the delegate, and clippys the UI for wasm32, but doesn't run `cargo make compress-webapp`), so there's no natural place to assert it without adding a new CI job outside this PR's scope. Left as a manual/PR-time check rather than a standing gate.

## Testing

- `cargo test --workspace`: all pass (harvest-ui 496, harvest-delegate 107, harvest-common 141, contracts 10+2+3).
- `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- `cargo clippy -p harvest-common --features address-guard --all-targets -- -D warnings` and its test run: clean (mirrors the CI job).
- `cargo clippy -p harvest-ui --target wasm32-unknown-unknown -- -D warnings -A clippy::for_kv_map`: clean (mirrors the CI job).
- `cargo check --manifest-path tests/rehearsal/Cargo.toml`: clean (mirrors the CI job; pre-existing unrelated warnings only).
- `cargo fmt --all -- --check`: clean.
- `cargo make compress-webapp`: builds clean, no `wasm-opt failed` in the log, and the publish bundle gate (`check-webapp-bundle.sh`) passes with the new fonts included (`15 required, 2 reachable`).
- Measured wasm size on the current tree: `harvest-ui_bg.wasm` goes from 7,685,419 bytes (broken: unoptimized, ~347 KB of DWARF sections) to 7,038,487 bytes (fixed: `-Oz` applied, zero `.debug_*` sections) — an 8.4% reduction, of which roughly 4% is wasm-opt's own size optimization beyond just removing debug info. The workspace's `[profile.release]` already uses `opt-level = "z"`, `lto = true`, `codegen-units = 1`, which limits how much further wasm-opt alone can shrink it.
- Verified the wasm-opt-failure check actually fires: temporarily forced `--debug-symbols true` back on, confirmed `cargo make build-ui` now exits non-zero (105) with the `wasm-opt failed` / SIGABRT signature printed and the new `ERROR: dx build failed, or wasm-opt aborted...` message, then reverted and re-verified the clean build.
- Verified the dx-version-pin check actually fires: temporarily changed the expected version string, confirmed `cargo make build-ui` fails fast (exit 105, before running `dx build`) with a clear message, then restored it and re-verified the clean build.

Fixes #28, fixes #34

[AI-assisted - Claude]


sanity (commented):
## Review: #86

**HEAD SHA:** `eb04216c2ad45930bb532511eb0b982562bc2c57`. CI is green (5/5).

**Tier:** Full, because this changes build and bundle configuration. Lenses run:
- **Skeptical.** Traced dioxus-cli 0.7.3's source for the grep's string. No pipe or exit-code loss.
- **Code-first.** Found that the grep alone does not meet #34's own acceptance criterion.
- **Publish path.** Built main and this branch for real:
  - main reproduces #34: a wasm-opt SIGABRT, yet exit 0 and unoptimised wasm with 5 `.debug_*` sections.
  - This branch: optimised, 650 KB smaller, fonts at the paths the CSS uses.
  - The bundle gate fails when a font is deleted.

Codex is over its usage limit (resets 2026-09-19), so these Claude lenses ran instead.

**Findings addressed in `eb04216`:**
- The wasm-opt grep depends on an internal dx log string, and nothing pinned the dx version. `build-ui` now refuses any dx other than 0.7.3, with instructions to re-verify the grep before moving the pin.
- Removed the redundant cleanup.
- Wrote down the debug-symbols trade-off.

The rebase onto main (`420182a`) had no conflicts, so no re-review of the earlier content was needed.

**State:** Ready to Merge.

[AI-assisted - Claude]

PR #89, opened 2026-09-18
ci: optimise test builds so the crypto-heavy merge-law tests run in seconds

sanity:
## Problem

The seeded random merge-law tests (`common/src/store.rs`, `reputation.rs`, `mailbox.rs`, added in #82) exercise RSA-2048/Ed25519/BLAKE3 through hundreds of random states. `cargo test` inherits `[profile.dev]` by default, so that crypto compiles and runs completely unoptimised, dominating the test suite's wall clock.

## Approach

Add `[profile.test] opt-level = 2` to the workspace root `Cargo.toml`. `debug-assertions` and `overflow-checks` are `[profile.dev]` keys this profile still inherits, so they stay on — only codegen quality changes.

`tests/rehearsal` is a deliberately separate cargo workspace (own `[workspace]` block, so this setting can't reach it even if it wanted to). CI never runs `cargo test` there anyway, only `cargo check --manifest-path tests/rehearsal/Cargo.toml` (no `--tests`), which builds under the `dev` profile, not `test`. So it needs no equivalent change.

## Testing

Measured locally on the same machine, warm target dir + warm sccache both times:

| | debug default | `opt-level = 2` |
|---|---|---|
| the 5 seeded merge-law tests (`cargo test -p harvest-common --lib merge_laws`) | 54.8s | 6.0s |
| `cargo test --locked --workspace` | 1m12.7s | 10.6s |

Verified nothing else changes behaviour:
- `debug-assertions=on` confirmed present in the resolved `rustc` invocation (`cargo test -v --no-run`).
- `overflow-checks` confirmed still on at runtime with a throwaway `#[should_panic(expected = "attempt to add with overflow")]` test under the new profile (removed before committing — not part of this diff).
- `cargo clippy --locked --workspace --all-targets -- -D warnings` and `cargo check --locked --manifest-path tests/rehearsal/Cargo.toml` both still pass.
- `~/.cargo/config.toml` on this machine sets `incremental = false` and dev-profile debug-info trimming globally (disk-housekeeping, unrelated to this repo); nothing there conflicts with `[profile.test]` opt-level. `.github/workflows/ci.yml` sets no profile overrides of its own.

CI job times on `main`, PR #82's merge commit (`5110283`), same job that this PR's diff speeds up:
- `fmt, clippy, test` job total: 5m27s (20:38:16–20:43:43)
- its `cargo test` step alone: 2m21s (20:38:48–20:41:09)

Will report this PR's own CI timing for the same job/step once it runs, for a direct before/after.

[AI-assisted - Claude]


sanity:
CI results for this PR (run 35394100401, head 5d3f63c):

- `fmt, clippy, test` job: **4m59s** (baseline on #82's merge commit: 5m27s)
- `cargo test` step: 3m9s (baseline: 2m21s) — **this one run is a wash, not a win**, and worth explaining rather than hiding: it's the first CI run under the new `[profile.test]`, so `Swatinem/rust-cache` restores a target dir built under the OLD profile, cargo detects the fingerprint mismatch and recompiles all 220 crates from scratch (2m56s of the step), which eats the savings this run. Once this merges and the cache updates, that one-time cost goes away.
- The actual payoff is visible where it should be, underneath that recompile cost — the two crypto-heavy test binaries:
  - `harvest-common` unit tests (177 tests, includes the seeded merge-law tests): **107.45s -> 8.94s**
  - `harvest-common --features address-guard` tests (181 tests): **111.91s -> 9.05s**

So the crypto-heavy tests themselves are ~12x faster, matching the local measurement in the PR description (54.8s -> 6.0s for the 5 seeded tests alone, on this machine). The job-level number will show the full win starting from the next PR, once the cache is warm under the new profile.

[AI-assisted - Claude]

sanity:
Light-tier review (one skeptical Claude lens; Codex is over its usage limit until 2026-09-19, so a Claude lens stood in). Findings:

- **Shipped artifacts:** none can be affected. The contract and delegate WASM are built with `--release`, the UI with dx's own profiles.
- **Runtime checks:** `debug-assertions` and `overflow-checks` stay on. That was checked empirically.
- **Tests:** none depend on unoptimised timing or stack depth.
- **The one finding:** the comment and the commit message cited timings from different runs. That is reconciled in `b2f28ca`, a comment-only change. CI is green on that head.

[AI-assisted - Claude]

Issue #79, opened 2026-09-18
First-run usability: what a non-technical person cannot work out from Harvest's own screens


sanity:
A reviewer with no Bitcoin and no Freenet background was asked to read a walkthrough of Harvest's screens and say where they got lost. The full report is long; these are the findings that are about the **app** rather than about the walkthrough, roughly in order of how much they would cost a real seller or buyer.

## The share link cannot be shared

My Store presents this under "Share this link so buyers can open your store":

```
http://127.0.0.1:7651/v1/contract/web/6FzSeAUKcqJrveKyU8RJgGKc5jRB1Z2juvxXtwTA4Em9/#store=6dGok...
```

The reviewer spotted, unprompted, that `127.0.0.1` means "this computer" and asked how they were supposed to send that to a buyer. They are right, and it is the single most load-bearing action a seller takes. A buyer needs their own node plus the `#store=` part; the string offered is not usable as given.

It is also visibly truncated in its box, and the label says nothing about what the buyer needs at their end. There is no copy button, either -- worth noting because the field is deliberately styled as a value to copy.

## A buyer has no visible way to buy

The buyer's view shows listings, prices and a "Contact Seller" button, and nothing that reads as "buy this". The purchase actually goes through messaging the seller, who then issues an invoice, but nothing on screen says so. The reviewer's conclusion was "eleven pictures and not one shows a buyer paying".

Whatever the flow is, the buyer's first screen should say what the next step is.

## "Settled by bridge X" appears under AWAITING PAYMENT

Every order card carries `Settled by bridge 4MZnDAQW` regardless of status, including directly beneath an `AWAITING PAYMENT` badge. "Settled" means paid in ordinary English, so the card says both things at once. It appears to mean "this is the bridge that would settle it".

Related, same panel: "Watched addresses -- Not watching any addresses yet." sits below orders that Harvest is plainly tracking, which reads as broken.

## The bridge stops watching, and the fix is asking the seller to log in twice a day

The panel says, accurately:

> the bridge stops watching about a day after the last renewal. A payment made while it is not watching is not picked up later, so a seller waiting to be paid should open Harvest more than once a day.

Read as a seller, that is "if you go away for the weekend a customer's payment can be missed permanently". The reviewer flagged this as a reason not to use it, and I think that judgement is correct rather than squeamish. The renewal needs to survive the tab being closed, or the consequence needs to be something other than a lost payment.

## Prices are Bitcoin-only

[...]

PR #91, opened 2026-09-18
feat: short store codes, owner-bound store state, portable share links

### LINKED ISSUE #52: store links: a 12-character store code, derived from the seller key as the contract's parameters
## Summary

A shareable Harvest store link should be short enough to paste into a message, and opening one should be all it takes to have that store from then on. Today the link carries the full 44-character contract id, and nothing remembers the store after the visit.

Adopt Delta's mechanism: make the store contract's `Parameters` a **12-character base58 prefix of the seller's Ed25519 verifying key**, so the short code *is* the parameter the address is derived from.

**This is cheap now and expensive later — see "Timing".**

## Why not just truncate the contract id

Because it is wrong, and `store_link.rs` already says so. `parse_store_id` explicitly refuses a short id:

> that decodes into a fixed 32-byte buffer and discards the written length, so a truncated id decodes to a *different*, zero-padded contract id instead of failing. A typo in a shared link should open nothing, not something else.

So a truncated address does not fail — it resolves to a real but wrong store. Shortening the *parameter* avoids this entirely.

## The mechanism (from Delta)

Delta's `SiteParameters` is a 10-character base58 prefix of the site owner's Ed25519 public key (`common/src/state.rs:16,47-50` in `freenet/delta`), and the contract address is `BLAKE3(BLAKE3(wasm) || cbor(prefix))`. Any client rederives the full `ContractKey` locally via `ContractKey::from_params_and_code`, with the site WASM embedded by `include_bytes!` — **zero network round trips, no registry, no naming service** (`ui/src/state.rs:421-431`).

Their comment puts it well: *"Anyone can do this — the WASM is public and the prefix is the only parameter."*

**Harvest's store contract is already the right shape.** `StoreParameters` has exactly one field, `seller_verifying_key` (`common/src/store.rs:58`). Delta holds a prefix of that key where Harvest holds all 32 bytes. That is the whole difference.

## Length: 12, not Delta's 10

10 base58 characters pin 58^10 ~= 2^58.6. Grinding a keypair colliding with a *chosen* seller is then on the order of 10^7 GPU-hours — very roughly $1-6M of rented compute, and falling every year. Ed25519 vanity grinding is mature and off-the-shelf (Solana uses the same curve and base58).

12 characters give 58^12 ~= 2^70.3 — about 4,000x more work, out of reach for the lifetime of the design. It costs two characters, and a link stays under a third of today's 44.

These are order-of-magnitude figures from published vanity-grinder throughput, not a measurement.

## What the length does not fix

Collisions are **first-writer-wins**. Delta's contract binds the address to the full owner key once claimed (`SiteState::merge`, `common/src/state.rs:508-536`), so an *established* store cannot be taken over — a second owner at the same address is rejected outright.

The residual risk is therefore **pre-emption, not usurpation**: squat the address before the real seller claims it, and their store can never be created at the only address their identity derives. A ghostkey circulates (reputation, other apps) well before its owner opens a shop, so the attacker's input is available.

Two things follow, and both need deciding:

1. **The refusal must be loud.** Delta surfaces nothing on a rejected merge. That is what turns a rare attack into an unexplainable one — the seller sees a store that simply never appears. Harvest should say "this address is already claimed by a different key".
2. **A blocked seller needs a second address** without discarding the ghostkey that carries their reputation. A salt or nonce in `StoreParameters` would provide one, at the cost of the prefix no longer being derivable from the key alone. Worth deciding deliberately rather than discovering later.

## Timing — the window is open now

Changing `StoreParameters` changes its field set, which changes every store's address: a re-key of everything published. **Nobody has created a real store yet, so today this costs nothing.** Once real stores exist it becomes a migration.

`StoreParameters::new` already carries a scar from exactly this hazard, in its own doc comment: it gained two Bitcoin fields and lost them again, and the migration probe went on deriving V1 addresses that were never published — reporting a clean "nothing to migrate" over a seller's entire store.

## Scope

- [ ] `StoreParameters` holds a 12-char base58 prefix of the seller verifying key instead of the full key; keep the private-field constructor discipline that is already there.
- [ ] Store contract validates that the state's owner key prefixes to the parameter, and binds to the full owner key on first claim (Delta's `merge` is the model).
- [ ] Surface the "already claimed by a different key" refusal to the seller rather than failing silently.
- [ ] `store_link.rs`: emit and parse `#store=<12 chars>`; derive the `ContractKey` locally from prefix + embedded WASM. Keep the length check, adjusted — a wrong-length code must still open nothing.
- [ ] Stores are remembered after a link visit (currently `open_store_from_url` browses but nothing persists), with removal writing a hidden marker rather than deleting conversations — the store list and the buyer's conversation history are currently the same data.
- [ ] Decide the salt/nonce escape hatch, or record explicitly that we are accepting no recovery from pre-emption.

## Note on the existing fragment rationale

While reading this I checked `store_link.rs`'s stated reason for using the fragment. One of its two justifications does not hold:

> A fragment is never sent to the server. The gateway therefore never learns which store a visitor opened, and the id never lands in an access log.

A Freenet webapp is served by the node's own HTTP interface on loopback (`http://127.0.0.1:50509/v1/contract/web/<id>/`), so in the normal deployment there is no third party in that path — the "access log" is your own machine's. It also conflates two different things called a gateway: a Freenet gateway is a UDP bootstrap peer and never sees an HTTP URL. Fragment-vs-query changes nothing at the network layer; the node issues the same contract GET either way.

The claim does hold for a *remote* HTTP endpoint (a hosted proxy such as try.freenet.org, or freenet-core#4381), which is a real but narrow case stated as a general property.

**The fragment is still right**, for the other reasons the same file gives: it survives Freenet's trailing-slash redirect where a query string survives only if the server re-attaches it, and it keeps the store id out of the query string that already carries `__sandbox=1`. Worth rewriting the comment to say that, so the rationale survives the next careful reader.

---

Length decided by @sanity (12 over 10, "let's play safe").

https://claude.ai/code/session_01W5BNFv1ZcroT2YWbkQ22A5

[AI-assisted - Claude]

sanity:
## Problem

A shared store link carried the 44-character contract id and, worse, the seller's own page URL: `http://127.0.0.1:7651/...` in the screenshots (part of #79). A buyer on another node opening that reaches their own loopback on the seller's port, so a seller could not hand their store to anyone. Nothing remembered a store after a link visit either.

## Approach

**Store codes.** `StoreParameters` now holds `store_code`, the first 16 base58 characters of the seller's verifying key (private field, `new` / `from_code` constructors). A link carries `#store=<16 chars>` and any client derives the `ContractKey` from the code and the bundled store WASM. `from_code` refuses any other length or alphabet, so a wrong-length code opens nothing. Per the decision on the issue there is no salt.

**Binding to the owner, convergently.** A code names many keys, so the owner moved into the state: `StoreStateV1::owner`. The contract requires the owner to begin with the code, verifies every record against it, and refuses a state that names an owner but holds nothing that owner signed (a key alone proves nothing; the cost of a code is the cost of a keypair, which only a signature shows).

**Why 16 characters (Ian's decision on #52, after the review).** 12 characters is about 2^70 against one chosen seller, but a grinder checks each key against every live code at once, so hitting some store is about 2^60 at 1,000 stores and 2^57 at 10,000, and a hit takes the store. 16 characters is about 2^94 against one seller and about 2^80 against 10,000 at once. Smaller-key-wins is kept.

Delta's `SiteState::merge` binds to the first claimant and refuses others. That is not convergent: if two keys sharing a code both publish before converging, each peer keeps whichever arrived first, forever. So the merge here uses a total order computable from the states alone: the smaller key by bytes wins the address and the other key's records are dropped; with one owner the per-record merge is unchanged. That is the lexicographic product of a chain of owners with each owner's own semilattice, so it is associative, commutative and idempotent. The cost, stated in the code and in `docs/untested-invariants.md`: a smaller key sharing a seller's code takes the address even after they publish, which is the same keypair search as pre-emption (figures above). The laws hold for honestly signed content; the unsigned-field ties of #81 remain as before. The top-level `ComposableState` is hand-written (it was `#[composable]`), because a change of owner has to drop the other key's records. Deltas name their owner; across owners a delta is everything or nothing.

**Loud refusal.** An outranked update is ignored rather than erroring (an error would make the merge order-dependent), so the UI compares the owner in the store state that comes back with the seller's key and says "already claimed by a different key", as a one-time notification and on the My Store card. Certificate verification now also requires the certified key to BE the state's owner, since the address only pins the code.

**Portable share link.** `http://127.0.0.1:7509/v1/contract/web/<Harvest's contract id>/#store=<code>`: the default Freenet node address, which is what `fdev` prints and what freenet.org's River invite button uses, plus Harvest's fixed web container id. Nothing from the seller's page (port, `__sandbox=1`). The card also shows the bare code, and the Browse tab now has an input that accepts a code or a pasted link, for a buyer whose node is on another port.

**Remembered stores.** New delegate requests `RememberStore`, `SetStoreArchived`, `ListRememberedStores` and secret family `harvest:known_store:{code}` (under `harvest:`, added to `all_secret_key_shapes`, code-validated so the 1024 cap bounds bytes, origin-gated). Archive, never delete (pinned structurally: the module is generic over `SecretStore` alone, which has no removal); visiting an archived store does not un-archive it, and a visit does not guess about an unreadable record. A store is remembered only once its state has loaded, so a mistyped or unreachable code never joins the list. The list is per node, not per Ghost Key. A pre-#52 link (a whole contract id) shows an "old-format link, ask the seller for their store code" notice. The Browse tab lists remembered stores with Archive / Unarchive and "Show N archived", with copy saying archiving keeps conversations and does not close a store you own.

## Re-key and migration

All four artifacts move (V16 store, V12 reputation and mailbox, V15 delegate recorded with provenance from `5110283`). This is a parameter change, so every store generation up to V16 lives under a whole-key encoding: `ui/src/migrate.rs` gains `store_param_shape` / `LAST_WHOLE_KEY_STORE_PARAM_GENERATION = 16` and a frozen copy of the whole-key encoding, alongside the existing V2..=V5 band. Whole-key states have no owner, so the fold names the seller (whose key derived the probed address) as owner of any such state holding records. **This delegate re-key drops the delegate's secrets again** (store registry, RSA and X25519 keys, conversations, Bitcoin config), as previous re-keys did: nothing drives the export handshake yet. Accepted pre-launch.

## Testing

- `claim_tests` in `common/src/store.rs` (two keys ground to a shared 2-character code; the contract accepts any code length, so this exercises the production merge), contract-level tests through `update_state` / `get_state_delta`, the ghostkey-cert owner test, migration tests (end to end from a V16 whole-key address; and a local snapshot owned by a different key is not reassigned by the owner fill-in), delegate `known_stores` tests including a structural never-deletes test and a gating test, `store_link` tests (including the old-format link), app-state tests (remembered only after load, once per session; the foreign-owner notice).
- Mutation-verified with `mut52/mut.py`, which now fails on any mutation that does not apply exactly once: 33 mutations of the new guards, all killed after fixes. Survivors found and fixed along the way: the ground-key fixture ordered its pair with the rule under test; a verify refusal was masked by the parts' own check; the once-per-session remember guard was hidden by the queue's own dedupe. The old-link 43..=44 length check bounds decode cost and changes no answer, so it is recorded rather than mutation-tested.
- `fdev verify-merge` on the built store WASM (16-character build): results in the review follow-up comment. The corpora use short codes by construction, since the contract accepts any code length.
- `cargo test --workspace --locked`, clippy (workspace, address-guard, UI wasm32), `cargo check` of `tests/rehearsal`. The rehearsal was not run against a live node.

Not done here: the UI component wiring (store list, the card) is wasm-only and untested, recorded in `docs/untested-invariants.md`. A buyer whose node is not on port 7509 uses the code rather than the link.

Fixes #52

[AI-assisted - Claude]


sanity (commented):
## Comprehensive PR Review: #91

**HEAD SHA reviewed:** `e96f9465406ce0eb2666fa14a9a99e1691947dcd`. CI is green.

**Tier:** Full (contract parameters, a hand-written merge, migration, new delegate requests).

**Lenses:** code-first, testing (in its own worktree), skeptical, big-picture, and merge-law/migration, which re-ran `fdev verify-merge` with new three-owner and two-owners-at-cap corpora.

**External model:** Codex is over its usage limit (resets 2026-09-19), so the Claude lenses stand in for it.

### What holds
- **The hand-written merge is sound.** It is commutative, associative and idempotent, including with three owners sharing a code. Proof by reasoning: a lexicographic product of the owner chain with each owner's own join.
  - **New corpora:** store-triad holds 4489 cases with 0 violations across all 14 properties; store-triadcap has 0 violations. Native pair and triple checks are all clean.
  - **The check can fail:** two mutants fail it. A mis-scoped cross-owner delta gives 47 violations; first-writer-wins gives 196.
- **`validate_state` accepts exactly merge's fixed points.** The #82 rules are intact.
- **Migration:** `LAST_WHOLE_KEY_STORE_PARAM_GENERATION = 16` is correct. The owner fill-in cannot assign the wrong owner, and nothing signed is lost. Reputation and mailbox are unchanged in shape.
- **Registry:** all legacy/ `code_hash` values match main's artifacts.
- **No third party without a colliding key can disrupt a store.** Checked: owner-less states, unsigned owners, owner-mismatched deltas, replayed signatures (`verify_scoped_signature` pins the requestor to Harvest webapp ids), and version-0 info.
- **Every #52 checklist item is done.** The rebase onto main will not conflict.

### Must Fix (decided by Ian)
1. **Lengthen the store code from 12 to 16 base58 characters.** Keep the smaller-key-wins rule.
   - **Why:** the 2^70 figure is the single-target cost. Grinding against every live code is N-fold cheaper: about 2^60 at 1,000 stores and 2^57 at 10,000. A hit displaces the store, and every replica drops the seller's listings and Paid orders.
   - **At 16 characters:** about 2^94 per target and 2^80 against 10,000 stores.
   - **Also update:** the threat-model text (state the multi-target cost) and the design note (code, docs, the #52 comment).

### Should Fix
2. **Remember a store only once it has loaded.** Right now `open_store` (`ui/src/store_link.rs:172-211`) calls `remember_store` before the GET succeeds, so a mistyped or unreachable code stays in the list permanently.
3. **Test the migration owner fill-in against a local snapshot owned by a different key.** Dropping the `is_none()` guard at `ui/src/migrate.rs:557` survives every test.
4. **Pin "archive never deletes"** with a structural test, because there is no guard to mutate. Also align `upsert()` with `list()` on undecodable records: don't silently un-archive.
5. **Old 44-character links** (`store_link.rs:158-160`) open nothing, silently. Show "this is an old-format link; ask the seller for their store code".
6. **The mutation harness skipped a guard silently.** `mut52/mut.py` reports `NOT APPLIED (count 0)` for "verify allows bare owner" and carries on. Make it fail loudly on any count other than 1. (The guard does hold; the testing lens killed it manually.)
7. **Docs:**
   - The `StoreStateV1` doc says merges agree "whatever the order or grouping". Add "for honestly signed content; #81 remains".
   - State that the remembered-store list is device-wide, not per Ghost Key.
   - Fix the garbled bullet at `common/src/store.rs:1058`.
   - A partial delta during an owner switch leaves a replica with a subset of the new owner's store until the next summary sync. Note it (it needs a colliding key).

### Consider
- A delegate re-key drops the delegate's secrets again (store registry, RSA and x25519 keys). That is accepted pre-launch; say it in the PR body.
- The link only opens as-is for buyers whose node is on port 7509. The bare code covers everyone else.

### Verdict
**State:** Needs Changes. Must Fix 1 changes contract parameters, so a focused re-review (merge-law and skeptical) follows.

[AI-assisted - Claude]

sanity:
Follow-up to the Full-tier review. Rebased onto `5df7ac8` without conflicts. Head is now `2c6d36ac3cd7e37aaf9f15b964f8aae8c52ef058`, and CI is green on it: all 5 checks, including "committed contract wasm matches a fresh build" and "no contract or delegate address moved" (the label is still applied).

| Review item | What was done |
|---|---|
| Must Fix 1: 16-character code | `STORE_CODE_LEN = 16`, and the parameters are now 29 B of CBOR. `from_code` only accepts exactly 16 characters, and link parsing and emitting, the UI copy and the tests all follow. The threat model on `STORE_CODE_LEN` and `StoreStateV1`, and in `docs/untested-invariants.md`, gives both costs: about 2^94 to take one chosen seller's code, and about 2^80 to hit any one of 10,000 stores at once (2^84 at 1,000). It also says why 12 characters (about 2^57 at 10,000 stores) was not enough. Smaller-key-wins is kept. The V16 entry in `legacy/store_contract.toml` is updated to match. No new rows were added: the outgoing generation is still `main`'s, unpublished since then. |
| SF2: remember only once loaded | `open_store` no longer remembers the store. `AppState::remember_loaded_store` does that when the store's state actually arrives, once per session. It returns whether it asked, so the guard can be observed. |
| SF3: the fill-in and a store another key owns | New test `the_owner_fill_in_does_not_reassign_a_store_another_key_owns`. It merges a local snapshot owned by a different key with the seller's whole-key predecessor. It fails when the `is_none()` guard is dropped. |
| SF4: archive never deletes; unreadable records | `nothing_here_can_delete_a_record` is structural: the module takes `SecretStore` alone, which has no removal, and the test also checks behaviourally that the key count never falls. `upsert` no longer reads an unreadable record as "not archived" on a visit: it leaves the record as it is and reports that, the same way `list` hides it. An explicit archive or unarchive replaces it. |
| SF5: old 44-character links | `store_link::is_old_format_link` now recognises a whole contract id. The Browse tab says "This is an old-format store link... Ask the seller for their store code, or a new link." |
| SF6: the mutation harness | `mut52/mut.py` now aborts (exit 2) when a mutation matches anything other than exactly once. The skipped "verify allows bare owner" entry applies now and is killed. Final run: 32 mutations, all killed. Survivors found and fixed along the way: the once-per-session remember guard was hidden by the queue's own dedupe, and one entry had a stale 11-character mutation. The old-link 43..=44 length check only bounds decode cost and changes no answer, so it is recorded as not mutation-testable rather than pinned by a timing test. |
| SF7: docs | The `StoreStateV1` doc now says the laws hold "for honestly signed content", and that #81 remains. The remembered-store list is documented as per node, not per Ghost Key. The garbled `delta` bullet is rewritten. There is a doc paragraph and a docs row for the partial-delta subset that can appear during an owner switch. |
| Consider: delegate re-key | The PR body now states that this delegate re-key drops the delegate's secrets again, and that this is accepted pre-launch. |

**`fdev verify-merge` on the rebuilt 16-character store WASM** (code hash `7c56e044...`). All corpora were regenerated against the new types: mine, plus the review's `store-triad` and `store-triadcap` built with the review's generator. They use short codes by construction, since the contract accepts any code length.

| Corpus | Result |
|---|---|
| `store` | 0 violations (1316 CLI cases, 2000 bundle cases) |
| `store-empty` | 0 violations |
| `store-cap` | 0 violations |
| `store-claim` | 0 violations (630 bundle cases, every property run) |
| `store-triad` | 0 violations (900 CLI cases; 2000 bundle cases, 97 inconclusive at the case cap; 0 in every per-property run) |
| `store-triadcap` | 0 violations |
| `store-adv` | 6 findings (3 `state_commutativity`, 3 `reconciliation_cycle`), the same as the #82 baseline, which are the known #81 unsigned-field ties |

Since this changes contract parameters again, it needs the focused re-review (merge-law and skeptical) the review called for. Not merged.

[AI-assisted - Claude]

sanity (commented):
## Final verdict: #91

**HEAD SHA:** `e8bfe8e20445a23d77b32c93ce74d0548109cc70`. CI is green (5/5).

**Review history:**
- **Full tier (5 Claude lenses):** code-first; testing, in its own worktree; skeptical; big-picture; merge-law/migration.
  - The merge-law lens proved the hand-written merge sound. It added three-owner and two-owners-at-cap corpora, both at 0 violations, and showed that two mutants fail them (47 and 196 violations).
  - Ian decided the code length: 16 characters, keeping smaller-key-wins.
- **Focused re-check after the fixes (skeptical and merge-law):**
  - verify-merge shows 0 violations across 11 corpora, except the #81 states.
  - The WASM rebuilds byte-identical.
  - Migration was cross-checked against main: all 16 old addresses are derived exactly.
- **Final cleanup (`e8bfe8e`):** doc fixes, the old-format notice for pasted links, the remembered list always answered, and a literal known-answer pin for the old whole-key encoding. No contract logic changed and no WASM hash moved. I read this diff myself.

Codex was over its usage limit throughout (resets 2026-09-19), so these Claude lenses were substituted, per the review rule.

**State:** Ready to Merge.

[AI-assisted - Claude]

PR #95, opened 2026-09-19
fix(store-contract): drop the diagnostic-only related-contract fetch

sanity:
## What

The store contract's `validate_state` returned `ValidateResult::RequestRelated` for any state holding Paid or PaymentReversed orders. It asked every validating peer to fetch up to `MAX_RELATED_CONTRACTS_PER_REQUEST` (10) Bitcoin address contracts, then cross-checked them only to log a line when one held no claims. Its own comments said it never affected validity, and it could not have without letting two peers holding identical bytes disagree.

So every peer paid up to 10 fetches per validation for a log line nobody reads. Contracts should be minimal and carry no diagnostics-only work, so this PR removes:

- the `RequestRelated` fetch and the cross-check loop in `validate_state`
- `MAX_RELATED_CONTRACTS_PER_REQUEST`
- the store contract's `bitcoin_address_instance_id` wrapper
- the store contract's direct `freenet-bitcoin-common` dependency, which nothing else in the contract used (it still arrives through `harvest-common`, with the same features)

Payment validity still rests on the proof embedded in each order, which `verify` checks as before.

## What stays, and why

`Order::bitcoin_address_code_hash` and `Order::bitcoin_address_instance_id` (in `harvest-common`) stay. The UI uses them to find the address contract to watch for each order (`state.rs` `address_contracts_to_watch`, settlement, and the buyer's live address view in `bitcoin_view.rs`). The Order wire format is unchanged. Only their doc comments changed, since they described the contract-side cross-check. Those edits keep line counts the same, and the rebuilt reputation and mailbox contracts and the delegate are byte-identical to `main`.

## Re-key (store contract only)

| | code hash |
|---|---|
| V17, outgoing (built at `bc57dac`, PR #91, committed on `main` through `03c082e`) | `7c56e044b2f69825f209500c54c433038e1be5b3e5cc59b78a92bf4173175f68` |
| new current | `1e4eba431c84532ecc684d1d3d664f8d8cbde9156f7a4fcab2af0ca20fcbc79c` |

- `legacy/store_contract.toml` gets a V17 row. V17 is the first code-addressed generation (29 B parameters), so it falls on the code side of `LAST_WHOLE_KEY_STORE_PARAM_GENERATION` (still 16) by itself.
- `ui/src/migrate/tests.rs`: the pinned-hash list and the `PUBLISHED_UNDER` shape table now include V17 (`Code`). New test `a_code_addressed_store_is_carried_into_the_next_generation` checks that a V17 store, which already names its owner, is found at the address the code parameters derive and verifies against the current contract. I ran a mutation to confirm the test can fail: with `LAST_WHOLE_KEY_STORE_PARAM_GENERATION = 17`, the probe recovers nothing and the test goes red.
- `tests/rehearsal`: `ENCODING_BY_GENERATION` was a `bool` (legacy or not) and assumed no recorded generation used the code shape. It now takes a three-way `Shape`, and V17 is `Code`. Checked with `cargo check`. Running the rehearsal needs a live node, and I have not run it.
- Tests and docs about the cross-check were removed or rewritten: three contract tests were replaced by one (`a_paid_order_validates_without_asking_for_related_contracts`, which checks a paid order with and without a code hash). The rows and ranked entry in `docs/untested-invariants.md` are now marked closed by removal, and `docs/bitcoin-integration-status.md` was updated.

## Verification

- `cargo test --workspace --locked`: all pass
- `cargo clippy --locked --workspace --all-targets -D warnings`, the `address-guard` clippy, and the UI wasm32 clippy: clean
- `cargo check --locked --manifest-path tests/rehearsal/Cargo.toml`: OK
- `cargo fmt --check`: clean
- `fdev verify-merge` on the rebuilt store WASM: the output is line-for-line identical to the PR #91 baseline across all 11 store corpora (same generator, byte-identical corpora, bundles regenerated against the new WASM). There are 0 violations except the 6 in `store-adv`, which are the known #81 unsigned-field states (3 `state_commutativity`, 3 `reconciliation_cycle`).

[AI-assisted - Claude]


sanity (commented):
## Final verdict: #95

**HEAD SHA:** `f2aafa7d42340f8756fa75a227f3ea117da28fbc`. CI is green (5/5).

**Tier:** Full.

**Lenses:**
- **Migration and re-key:** verified by BLAKE3 and by rebuilding every artifact from source.
- **Code-first/big-picture:** nothing functional was removed and no stale references remain.
- **Testing, in its own worktree:** every requested mutation was caught.

**Follow-up applied (`f2aafa7`):** the rehearsal span guard now also requires a code-shaped generation.

Codex was over its usage limit throughout, so these Claude lenses were substituted. The merge was held overnight per the no-late-night-merges convention.

**State:** Ready to Merge.

[AI-assisted - Claude]

Issue #93, opened 2026-09-19
Revision 2: store identity, Ghost Key backing, custody, and the seller-first UI


sanity:
Tracking issue for implementing revision 2 of Harvest's entity model. Ian approved it on 2026-09-18. The design record is currently `~/code/tmp/harvest-entity-model.html` on nova, and it will move into `docs/` in the first PR.

## Decisions (summary)
- **A store has its own Ed25519 store key.** The 16-character store code (#52, #91) derives from it.
- **Ghost Keys back a store.** A backing is a statement signed by the Ghost Key and accepted by the store key. There is one current backing at a time (several at once is deferred; the contract shape allows it later), and upgrading means retire plus add. A Ghost Key backs one store at a time, and readers enforce that.
- **The record follows the store.** A new Ghost Key never resets it. A clean slate means a new store plus a new Ghost Key.
- **Store key custody.** The store secret is wrapped to each backing Ghost Key in store state, using a key derived from the vault's deterministic signature over a domain-separated message. The inbox and record keys derive from the store key. A spike validates this first.
- **No rotation after a leak.** Instead, a one-way "closed" flag signed by the store key; buyers' software refuses to pay a closed store.
- **Complaints are stored once**, in the store's record. Each carries a buyer-signed Bitcoin block reference (height and hash), and the client applies a future-dating check before showing it. There is no aging.
- **The Ghost Key record is a minimal index** of the stores the key has backed, used to discover stores on a new device.
- **The payout wallet is per store**, kept on each device. The address counter is recovered from that store's published orders.
- **Removals:** `Order.buyer_fingerprint`, the untied "Issue an invoice" form, and `seller_fingerprint` in store details. `ListingKind::Request` is hidden.
- **No related contracts.** Every cross-contract fact is checked by readers. Contracts stay minimal.
- **UI:** navigation becomes Stores / My purchases / My store. My store has Overview, Listings, Orders (the inbox) and Settings. The Reputation and Payments tabs go.

## Plan (one PR at a time; each phase 1 PR re-keys contracts, which is cheap pre-launch)
- [ ] Remove the diagnostic-only related-contract fetch from the store contract (in progress)
- [x] Spike: validate store-key custody. All four checks pass (branch `spike/store-key-custody`, efa6930); report findings are folded into the design doc
- [ ] #94: stop logging key material in the UI (prerequisite for custody; in progress)
- [ ] Phase 1a: store key, backing statements, closed flag, store content signed by the store key
[...]

PR #96, opened 2026-09-19
fix(privacy): stop logging delegate and vault responses verbatim (#94)

### LINKED ISSUE #94: UI logs delegate responses verbatim, including conversation keys and backup secrets
`ui/src/gateway/response_handler.rs:343-351` logs every delegate response in full at `info!`:

```rust
info!("Harvest delegate response: {:?}", r);
info!("Ghostkey response: {:?}", r);
```

`dioxus::logger::initialize_default()` keeps `info` in release builds, so these reach the browser console in production. Several responses carry secrets, and their `Debug` is derived:

- `HarvestDelegateResponse::ConversationKeys` prints each `ConversationKey`'s `buyer_to_seller` / `seller_to_buyer` AES-256 keys (`common/src/delegate.rs:553`).
- `HarvestDelegateResponse::BuyerConversationExported` prints the backup string, which its own doc says "holds a secret" (`common/src/delegate.rs:408`).

`ConversationSecret` was given a redacting `Debug` for exactly this reason (`delegate.rs:706-728`), so the intent is already that secrets do not reach the console; these two paths slipped past it.

It also blocks store-key custody (revision 2, #93): the custody design derives the store key's wrapping key from a vault `SignResult` over a fixed message, so that signature is a secret. Through line 351 it would be printed in full, and `GhostkeyResponse`'s `Debug` lives in `ghostkey-common` where Harvest cannot redact it.

Suggested fix: log the response variant name only (plus non-secret identifiers such as `request_id` or the fingerprint), never `{:?}` of the whole response, and add a test that formats every secret-bearing response and asserts no key bytes appear. Harvest is pre-launch, so nothing has leaked from real users.

[AI-assisted - Claude]

sanity:
Fixes #94.

The UI logged every delegate response with `{:?}` at `info!`, which release builds keep. This replaces those log lines with secret-free summaries, gives the secret-bearing protocol types a redacting `Debug`, and adds a test that covers every request and response variant.

## Leaks found and fixed

Log sites that formatted a whole response:

- `ui/src/gateway/response_handler.rs:343` `info!("Harvest delegate response: {:?}", r)`: printed `ConversationKeys` (both AES-256 direction keys per buyer), `BuyerConversationList` (both direction keys per recalled conversation) and `BuyerConversationExported` (the backup string, which contains the X25519 secret).
- `ui/src/gateway/response_handler.rs:351` `info!("Ghostkey response: {:?}", r)`: printed `SignResult` signatures (a secret once store-key custody derives a wrapping key from one, #93), and would print `ExportResult`/`ExportAllResult` signing key PEMs.
- `ui/src/gateway/response_handler.rs:347` `info!("Bitcoin delegate response: {:?}", r)`: no private key material, but it printed the seller's payment xpub, which names every address they will be paid at.
- `ui/src/state.rs:5503`, `:6080`, `:6245`: the "Unhandled delegate / ghostkey / bitcoin delegate response" fallbacks, same `{:?}` of the whole response.
- `ui/src/gateway/response_handler.rs:31` `info!("Unhandled host response: {:?}", other)`: a `StreamChunk` is raw bytes of whatever is being streamed.

`Debug` output that would leak through any future `{:?}`:

- `ConversationKey` (derived `Debug`, both direction keys).
- `RecalledConversation` (derived `Debug`, both direction keys).
- The backup string in `BuyerConversationExported` and `ImportBuyerConversation` (a plain `String`).
- `PaymentXpubStatus` and `BitcoinDelegateRequest::SetPaymentXpub` (the xpub; `AppState` derives `Debug` over the status).
- `FeedbackToken` and `TransactionRecord` (nonce, entry key and blinded token together).

Found in review:

- `ui/src/gateway/response_handler.rs` decode errors put serde's message into an `error!`, and it can quote a string value.
- `ui/src/gateway/response_handler.rs` logged a delegate's user-input prompt with `{:?}`.

The sweep covered every `info!`/`warn!`/`error!`/`debug!`/`trace!`/`println!`/`eprintln!` call in `ui/src`, `delegates/harvest-delegate/src` and `common/src`, and every `{:?}`/`{:#?}` outside tests. The remaining non-test `{:?}` arguments are contract/delegate keys, truncated contract ids, and small enums (network, artifact). The harvest delegate itself logs nothing. UI-side types that hold keys (`messaging::ConversationKeys`, `state::ConversationBackup`, `BuyerConversation`'s secret) already redact.

## Approach

- **Log lines** use the summaries in `ui/src/gateway/log_summary.rs`. A summary prints the variant name and the request id or Ghost Key fingerprint, and nothing else. This covers Harvest, Bitcoin, vault and host responses, and delegate user-input prompts. The enums are `#[non_exhaustive]` from the UI, so each summary has a wildcard arm, and it prints no fields.
- **The call sites are guarded** by `ui/src/gateway/log_scan.rs` (test-only). It scans every formatting macro in the UI, including `format!`, `write!`, `panic!` and assertion messages, and allows `{:?}` only for per-file vetted arguments that state their type. It refuses any shape it will not reason about (`target:`, tracing fields, raw-string formats, `dbg!`, `event!`, `web_sys` console calls), and output macros may not interpolate an expression named like a secret. `DelegateResponse` no longer derives `Debug`. The structural alternative is #97.
- **The websocket URL** is logged without its query string, which carries `authToken=`.
- **Decode errors** report the sender, the CBOR variant name and the payload length. They no longer include serde's message, which quotes a string wherever it expected an enum, such as an xpub or signing key PEM on version skew.
- **Redacting `Debug`** in `harvest-common`:
  - `ConversationKey` and `RecalledConversation` redact their direction keys.
  - `FeedbackToken` redacts its nonce and entry key, and `TransactionRecord` redacts its blinded token and blind signature. Printed together, these undo blind-signing unlinkability.
  - `PaymentXpubStatus` and `BitcoinDelegateRequest::SetPaymentXpub` redact the xpub.
  - The backup string is `BackupString`, a `#[serde(transparent)]` newtype. The wire is unchanged, which `a_backup_string_encodes_as_its_text` pins.

Behaviour is otherwise unchanged: `BackupString` is unwrapped at the UI and delegate boundaries.

## Tests

- `common/src/delegate.rs::no_delegate_message_prints_a_secret` samples every variant of the Harvest request and response enums and of both Bitcoin enums. Each field that must not print is filled with a sentinel. It asserts that neither `{:?}` nor `{:#?}` shows the sentinel, and the check ignores whitespace. Exhaustive `classify_*` functions force every new variant to be classified, and a coverage check then fails until it has a sample. `carries_secret` confirms each secret-bearing sample really holds the sentinel.
- In `ui/src/gateway/log_summary.rs`, exact-match summary tests cover the Harvest, Bitcoin and vault summaries. The vault test first asserts that the upstream `{:?}` does leak.
- `payload_shape` has its own test, and `response_handler.rs::a_decode_error_does_not_quote_the_payload` covers decode errors.
- The call-site scan is `no_call_site_formats_an_unvetted_value`, with `the_scanner_sees_every_shape` pinning each shape. The delegate's undecodable-request error is covered by `an_undecodable_request_is_refused_without_quoting_it`.

Each fix was verified by mutation; see the mapping comment on this PR.

## Re-key

The harvest delegate re-keys: it compiles the new `Debug` impls. The outgoing generation `4fae20e1...` (from `bc57dac`) is recorded as V16 in `legacy/harvest_delegate.toml` and in the lineage test. The new delegate hash is `4d36b2a4...`; the intermediate builds on this branch (`d093dd43...`, `4183ba5a...`) were never merged or published. The store, reputation and mailbox contracts are byte-identical to `main`. Parameters are unchanged. Harvest is pre-launch, so no user data is stranded.

Checked locally: `cargo test --workspace --locked`, `cargo clippy --locked --workspace --all-targets -- -D warnings`, the `address-guard` clippy, the wasm32 UI clippy and `cargo fmt --all -- --check` all pass.

[AI-assisted - Claude]


sanity (commented):
## Comprehensive PR Review: #96

**HEAD SHA reviewed:** `8b5de6e735625f3211ff1581b14896bbaf16671e`. CI is green.

**Tier:** Full (security: key material in logs; delegate re-key).

**Lenses:** a dedicated secret-leak hunter, code-first/big-picture, and testing (in its own worktree). **External model:** Codex is still over its usage limit (resets 2026-09-19), so the Claude lenses stand in for it.

### What holds
- **No high- or medium-severity leak remains.** Every path that previously printed secrets in release now logs a summary. The leak lens also checked every macro and console call, panic, Display/error string, the delegate crate, and freenet-stdlib's client.
- **Wire format:** `BackupString` is serde-transparent (the test pins the bytes).
- **Delegate re-key:** V16's outgoing hash equals main's committed delegate WASM (BLAKE3). The new hash matches CI's fresh build, and the contracts are byte-identical to main.
- **The core test is load-bearing:** `no_delegate_message_prints_a_secret` catches restored derived Debug and misclassification.

### Should Fix
1. **The Bitcoin log summary is untested.** Mutating the `PaymentXpub` arm of `BitcoinDelegateResponse::log_summary` to print the whole status (which includes the xpub) survives all 907 tests. Add a sentinel test for it. Also give `PaymentXpubStatus` and `BitcoinDelegateRequest::SetPaymentXpub` a redacting Debug: they still derive it, and `BitcoinState`/`AppState` derive Debug over them.
2. **The call sites are unguarded.** Reverting the "Unhandled host response" fallback (`ui/src/gateway/response_handler.rs:19-37`) to `{:?}` survives all tests. So would the three "Unhandled ... response" fallbacks in `ui/src/state.rs` and `apply_delegate_response`. #94 was exactly a call-site `{:?}`. Guard the call sites, not only the summary functions.
3. **Simplify: move the Harvest and Bitcoin summaries into the UI crate** (as `ghostkey_response_summary` already is) and remove the `log-summary` feature of harvest-common. The summaries are only used by the UI, so there is nothing to gate if they never compile into `common`.
4. **Decode errors can echo a secret string.** `response_handler.rs:387`, `:394` and `:523` put serde's message into an `error!`, and on version skew serde quotes the offending string value (the backup, the xpub, `signing_key_pem`). Log the sender and payload length instead.
5. **The `{:#?}` half of the test cannot fail for byte secrets** (`common/src/delegate.rs:989-993`). Pretty-printed arrays put each element on its own line. Also match with whitespace stripped.
6. **Small ones:**
   - Comment that any new secret field must use the `SECRET` sentinel (classification is per variant, not per field).
   - Redact `FeedbackToken`/`TransactionRecord` Debug. Printing `entry_key`, the nonce and the blinded token together undoes blind-signature unlinkability.
   - Summarise delegate user-input prompts at `response_handler.rs:527` instead of `{:?}`.

### Verdict
**State:** Needs Changes. A focused re-check of the fix commits follows.

[AI-assisted - Claude]

sanity:
Review fixes are in `3c35b9d`. Mapping to the Should Fix items, with the mutation that proves each check goes red:

1. **The Bitcoin summary is tested, and the xpub is redacted.** `PaymentXpubStatus` and `BitcoinDelegateRequest` (for `SetPaymentXpub`) now have a hand-written `Debug`. Every arm of the request enum names every field, so a new field has to be decided on before it compiles. Both Bitcoin enums are now in `no_delegate_message_prints_a_secret`, with exhaustive classify functions. `a_bitcoin_response_summary_prints_only_its_name_and_request` checks the summary by exact match.
   - Mutation: `PaymentXpub` arm prints `{status:?}`. Result: red.
   - Mutation: the status prints `xpub`. Result: red (bitcoin response variant 6).
   - Mutation: `SetPaymentXpub` prints `xpub`. Result: red (bitcoin request variant 6).
2. **The call sites are guarded.** `no_log_line_debug_formats_an_unvetted_value` scans every UI source file and allows `{:?}`/`{:#?}` in a log macro only for arguments listed in `VETTED_DEBUG_ARGS`, each with its reason. It scans real code only: comments, `#[cfg(test)] mod` blocks and `tests.rs` files are stripped. The comment there notes that a scrape can match itself. `the_log_scan_finds_what_it_must_and_nothing_else` pins what the scanner sees: positional, inline and named args, `{:#?}`, a `//` inside a string, and it ignores comments and test modules. The structural half is that `DelegateResponse` no longer derives `Debug`, and every response is logged only through `gateway::log_summary`. `GhostkeyResponse`, `HostResponse` and the common enums derive `Debug` in crates the UI cannot change, which is why the scan is needed as well.
   - Each of these, reverted to `{:?}`, turns the guard red, naming the file and line: the host fallback, all three `state.rs` fallbacks (one as `{:#?}`), the ghostkey arm of `apply_delegate_response` (inline `{r:?}`), and the user-input prompt.
3. **Simplified.** The Harvest and Bitcoin summaries moved to `ui/src/gateway/log_summary.rs`, next to the ghostkey and host summaries. The `log-summary` feature, its cfg gates and the Cargo wiring are gone, so `common/Cargo.toml` and `ui/Cargo.toml` are identical to main. The contracts are still byte-identical to main.
4. **Decode errors.** `decode_delegate_message` reports `payload_shape`: the CBOR variant name (only if it is a plain identifier) and the length. It no longer includes serde's message. `a_decode_error_does_not_quote_the_payload` first shows that serde does quote an xpub or requestor string that sits where an enum is expected, then asserts that our error does not. The log line names the `DelegateSender`.
   - Mutation: the ghostkey arm returns serde's message. Result: red.
5. **Whitespace-insensitive.** `leaked` strips whitespace before matching.
   - Mutation: `ConversationKey` prints its key only under `{:#?}`. With the new check, red (variant 9). With the old needle, the same mutation passes.
6. **Small ones.**
   - The comment by `SECRET_BYTE` tells authors to use the sentinel for any new field that must not print, because classification is per variant.
   - `FeedbackToken` redacts its nonce and entry key, and `TransactionRecord` redacts its blinded token and blind signature. `BeginTransaction` and `TransactionList` are now classified as secret-bearing. Mutation: either one printing its field. Result: red (variant 14).
   - User-input prompts are logged as `request N (k response option(s))`.

**Hashes.** The delegate moved again, to `4183ba5a…`. The intermediate `d093dd43…` was never merged or published. V16 (`4fae20e1…`, main's) still records the outgoing generation, and its description is updated.

One trap worth knowing: putting `FeedbackToken`'s `Debug` impl before `impl FeedbackToken` renamed `nonce_for`'s mangled symbol, because rustc numbers impl blocks in source order. That alone re-keyed the reputation contract. The impl now sits after the inherent impl, with a comment saying why, and the reputation contract matches main.

Checked locally: `cargo test --workspace --locked`, workspace clippy, `address-guard` clippy, wasm32 UI clippy and `cargo fmt --check` all pass.

[AI-assisted - Claude]

sanity:
Round 2 fixes are in `aee9849`. Each item is mapped below with the mutation that proves it goes red.

1. **The call-site scan fails closed and covers more.** It moved to its own test-only module, `ui/src/gateway/log_scan.rs`. The module docs list exactly what is covered and what is not.
   - Refused outright:
     - a format macro whose format argument is not a plain `"..."` literal (`target:`, `x = ?r`, `?r`, `%r`, raw strings, a variable);
     - `info!{..}` and `info![..]`;
     - any `?`/`%` field sigil after the format string;
     - `dbg!`, `event!` and the `span!` family;
     - any `web_sys` `console::…(` call.
   - Also scanned now: `format!`, `format_args!`, `write!`, `writeln!`, `panic!`, `unreachable!`, `todo!`, `unimplemented!`, and the `assert`/`debug_assert` message forms. A `{:?}` laundered through a string or a helper is therefore caught where it is formatted.
   - `the_scanner_sees_every_shape` has one fixture per shape, each asserting its exact findings.
   - Mutations on real code, each red with file and line:
     - two-step `let described = format!("{:?}", response); info!("{described}")` in the `state.rs` fallback;
     - `info!(target: "harvest", ...)` in `response_handler.rs`.
   - The three real `format!` hits are now vetted entries with reasons: `state.rs` `status` (`OrderStatus`), `migrate.rs` `SIZE_CLASS_CAPS` (a constant), and `connection.rs` `e` (the browser's `WebSocket::new` error).
   - An entry nothing uses any more fails the test.
2. **No auth token in logs.** `connection.rs` logs `node_address`, the base URL with its query string cut off. Output macros may not interpolate an expression whose text names a secret: `websocket_url`, `token`, `secret`, `password`, `signing_key`, `xpub`, `backup`. This applies to Display too; `format!` is exempt because it builds the URL rather than emitting it.
   - Mutation: restore `info!("…{}", websocket_url)`. Result: red.
3. **Non-ASCII.** The lexer steps by char boundary and recognises char literals of any width. `scan_file` turns any internal panic into `log scanner could not parse <file>: …`, and `a_scanner_failure_names_the_file` pins that.
   - The fixture includes `'é'`, `'—'`, `'\u{e9}'` and lifetimes.
   - Mutation: revert both the boundary step and the char-literal width. Result: the fixture panics with "not a char boundary" inside `'—'`. Either fix alone prevents the panic, because the only non-ASCII outside strings and comments is in char literals and identifiers.
4. **By-text vetting is documented** on `Vetted`. Entries are now scoped to one file and must state the type they assume (`ty`).
5. **`FeedbackToken` and `TransactionRecord` destructure** in `Debug`, so a new field does not compile until it is named there.
6. **Delegate decode error.** The delegate's `payload is neither …` error names the variant and length (`payload_shape`) instead of serde's message. `an_undecodable_request_is_refused_without_quoting_it` first shows serde quoting an xpub-shaped string, then asserts that the delegate's error does not.
   - Mutation: return serde's message. Result: red.
7. **Follow-up** filed as #97: redacting newtypes over the raw response enums, so the UI cannot `Debug`-format them at all.

**Hashes.** The delegate moved again, to `4d36b2a4…`; the earlier builds on this branch (`d093dd43…`, `4183ba5a…`) were never published. V16 (main's `4fae20e1…`) still records the outgoing generation, and its description now mentions the decode-error change. The contracts are byte-identical to main.

Checked locally: `cargo test --workspace --locked`, workspace, `address-guard` and wasm32 UI clippy, and `cargo fmt --check` all pass.

[AI-assisted - Claude]

sanity:
Final-check fixes are in `281c148`:

1. **Nested macros are scanned.** After checking a macro's format string, the scan now continues inside its arguments (`i = open + 1`) instead of jumping past them. The outer macro's findings still come only from its own format string. New fixtures `info!("{}", format!("{:?}", r))` and `info!("{}", format_args!("{:?}", r))` each expect `D("r")`.
   - Mutation: in the `state.rs:6085` ghostkey fallback, replace the summary with `format!("{:?}", response)`. Result: red (`state.rs:6085: {:?} of response`). With the old jump-past restored, the same mutation passes, and the new fixtures fail.
2. **The websocket error text is dropped.** `WebSocket::new`'s error no longer appears in the connection error. Chrome's SyntaxError quotes the whole URL, token included. The error is now "could not open the websocket to <node address without query>". The inaccurate vetted entry for `e` is removed; the stale-entry check would have flagged it otherwise.
3. The residual low-severity bypasses are listed on #97.

UI-only change: the delegate hash (`4d36b2a4…`) and the contracts are unchanged, and CI's "committed contract wasm matches a fresh build" passes. CI is green on `281c148`.

[AI-assisted - Claude]

sanity (commented):
## Final verdict: #96

**HEAD SHA:** `281c148786c2136051435190ab944d55ad472ddd`. CI is green (5/5).

**Review history:**
- **Full tier:** a secret-leak lens, code-first/big-picture, and testing in its own worktree.
- **Two fix rounds**, each re-checked. Round 1 re-check: the leak lens and testing lens. Round 2: a final adversarial pass that inserted realistic bypasses into real code.
- **The last commit** (`aee9849..281c148`, UI only): the scan now continues inside a log macro's arguments, so `info!("{}", format!("{:?}", r))` is caught (fixtures added, a real mutation goes red). The browser's WebSocket error text, which Chrome can fill with the token URL, is no longer shown. I read this diff myself.

Codex was over its usage limit throughout (resets 2026-09-19), so the substituted Claude lenses stand in for it.

**Residuals:** low-severity bypasses are recorded on #97, which proposes the structural fix: UI-owned redacting newtypes over the raw responses.

**State:** Ready to Merge. It will merge after #95, in the morning (US Central), per the no-late-night-merges convention.

[AI-assisted - Claude]

sanity:
Rebased onto main after #95 merged (`87f7784`). The rebase applied with no conflicts, so no re-review was needed. CI is green on the rebased head `fb79efb`, including the check that the committed WASM matches a fresh build. Merging.

[AI-assisted - Claude]

PR #98, opened 2026-09-19
feat: store keys, Ghost Key backings and the closed flag (#93 phase 1a)

sanity:
Part of #93 (phase 1a). Rebased onto `main` after #95 and #96 merged and were published. 1a, 1b (#99) and 1c merge and publish together. The fix rounds are in the comments below.

## Problem

A store has been owned by its seller's Ghost Key, so it could not outlive that key: raising the stake or replacing a leaked key meant a new store, a new link and no history. Revision 2 of the entity model (moved into the repo here as `docs/design/entity-model.md`) gives a store its own key and has Ghost Keys back it.

## Approach

**A store key owns the store.** `StoreStateV1::owner` is now an Ed25519 store key; the store code derives from it and it signs the details, listings and orders. It is minted by the Harvest delegate (`CreateStoreKey`) and kept there, per device; the UI asks the delegate to sign (`SignStoreUpdate`) and never holds the secret. The delegate signs only a store's own records. Store-key signatures use the same `ScopedPayload` envelope as the vault, so one verifier checks both. `store_keys::load` and `keep` are the only reader and writer of the secret family: that is the seam phase 1b's custody requests plug into.

**Backings, retirements, the closed flag.** A backing is a Ghost Key's signed statement (store key, backer, its certificate, a network-qualified block reference), countersigned by the store key. Backings and retirements (store-key signed, per Ghost Key, permanent) are grow-only sets; the closed flag is a one-slot set, so it merges by OR. A slot clash resolves to the smaller encoding, as orders do. Nothing removes an entry. Bounds: 64 Ghost Keys with a backing or a retirement, and a 4 KiB certificate. Past the bound the merge keeps the 64 smallest keys, each key's backing and retirement together, so the merge is total and cannot un-retire a key in any arrival order (this replaced a refusal after the review, then a retain-with-backing rule after the merge-law re-check; see the comments).

**Reader rules** (never contract rules): the current backing is the unretired one with the highest block height (backing key breaks a tie), leaving out one dated above a known tip. A Ghost Key that is the current backing of two loaded stores counts for neither. Orders are checked against the store key; a closed store gets `PaymentBlocker::StoreClosed` before any other check and shows a banner.

**Migration.** Stores up to V18 were owned by a Ghost Key. The current contract accepts such a state as it is, so the probe carries it forward unchanged; it arrives unbacked, so buyers cannot pay it. My Store offers "Move this store": the creation flow with its name, description and listings re-signed by a new store key (listing ids unchanged), backed by the old Ghost Key. New code and link; open orders not carried. Store-key-owned stores get their own migration probe keyed by the store key.

**Not in this phase:** custody (1b), so a store key lives on one device and a delegate re-key loses it; the mailbox and record are still addressed by the Ghost Key (1b, 1d), so there is no change-Ghost-Key or retire UI yet.

Re-keys everything: store V18, reputation V13, mailbox V13, delegate V16 recorded in `legacy/`; store parameters unchanged (29 B). Needs `contract-rekey-acknowledged`.

## Testing

- `cargo test --workspace --locked`, clippy (workspace, address-guard, UI wasm32), rehearsal check: green locally.
- Seeded merge-law test for the new sets (61 states, 400 cases, byte-level, plus inflation).
- 41 mutations over the new guards (common, delegate, UI), all red; one survivor (the `apply_delta` cap, masked by a `verify` in the test helper) led to a tighter test. Rows in `docs/untested-invariants.md`.
- `fdev verify-merge` on the built WASM: 26 corpora. New: `store-backing` (honest: store-key owner, backings, retirements, closure, with details and listings; 14 states, 4 transitions, 7 delta steps) 0 violations over 557 cases, 1115 with the delta bundle; `store-backing-adv` (same-slot clash between two genuinely signed backings, a retirement of a never-backing key) 0 violations over 712; `store-backing-bad` (backing attached by the wrong Ghost Key, countersigned by a stranger, no countersignature; retirement and closure by a stranger) 0 violations, the invalid states refused (inconclusive). Every pre-existing corpus is identical to the #95 baseline: 0 violations except the known #81 unsigned-field cases in `store-adv` (6) and `reputation-adv` (8).

[AI-assisted - Claude]


sanity (commented):
## Comprehensive PR Review: #98 (phase 1a of #93)

**HEAD SHA reviewed:** `8fc40ddaea83f3a2900f8038533a553b9956f821`. Stacked on #95.

**Tier:** Full.

**Lenses:** code-first against the approved design, skeptical/security, testing (in its own worktree), and merge-law/migration (re-ran `fdev verify-merge` with new race and cap corpora). **External model:** Codex is still over its usage limit, so these Claude lenses stand in for it.

### What holds
- **Cryptography:**
  - No cross-type signature replay. `verify_scoped_signature` compares exact CBOR and pins the requestor.
  - No third party can attach, retire or close anything.
  - `SignStoreUpdate` and `CreateStoreKey` are origin-gated, and the allow-list is enforced at both layers.
  - A retired key can't come back.
- **Merge laws below the caps:** every merge that succeeds obeys them. The new race corpus (backing and retirement in either order, closure racing listings) passes 1115/1115 with all 14 properties.
- **Tests:** 940 pass. The testing lens killed 13 of 15 independent mutations.
- **Design conformance:** the implementation matches the PR description line for line. The legacy rows match the committed WASM (b3sum).

### Must Fix
1. **Merge is not total at the caps.**
   - **What happens:** `SignedSetV1::apply_delta` (`common/src/backing.rs:428-470`) refuses when a union passes 64 records. `apply_parts` is all-or-nothing, so any listing or closure in the same delta is refused too, and the replicas stay split for good. That was confirmed natively and in WASM (20 of 100 ordered pairs).
   - **Why the green runs missed it:** fdev files these cases as "inconclusive (contract error)", not violations.
   - **Who can reach it:** the store-key holder, for free, since certificates aren't checked. A thief holding a leaked key can use this to keep a closure from propagating.
   - **Fix:** merge must be total and deterministic. The closed flag must be applied independently of the capped sets and never blocked. No trimming may un-retire a backing, and a retirement must never be dropped while its backing is kept. Cover the delta laws at the cap as well.
2. **The closed flag isn't enforced on every payment surface.**
   - **Where it's missing:** the store page's `StoreInvoices` (`ui/src/components/store_view.rs:297`) still shows copyable payment addresses under the "closed" banner. The Bitcoin tab's `my_orders`/`OrdersSection` (`ui/src/components/bitcoin_view.rs:89-108`, `:592`) shows awaiting orders from closed stores.
   - **Why it matters:** a thief with the leaked key can keep publishing orders to their own address, which is exactly the case section 6.4 exists for.
   - **Fix:** no payable surface anywhere for a closed store. Apply the same treatment to unbacked stores.
3. **"Move this store" can run twice.**
   - **Where:** `start_store_creation_if_ready` takes `pending_store_creation` (`ui/src/state.rs:5534`) before the PUTs finish, so the button comes back (`ui/src/components/my_store.rs:333-335`).
   - **Effect:** a second store backed by the same Ghost Key. Both then count for nothing, and there's no retire path.
   - **Fix:** a persistent in-flight marker, plus the seller-facing section 6.2 check ("This Ghost Key already backs <store>") at create and move time.

### Should Fix
4. **The store key's seed can leave the delegate.** `export_scope()` (`delegates/harvest-delegate/src/migration.rs:87-89`) is `harvest:`, which includes `harvest:store_sk:*`. Exclude store keys, and fix the "never leaves" docs (`store_keys.rs:9`, `delegate.rs:346`).
5. **Tip lag strands a new store.** A backing is dated at the seller's tip, `current_backing` drops heights above the reader's tip, and the view is only recomputed when store state arrives. Date backings a few blocks back, and recompute verdicts on tip updates.
6. **Regression: legacy stores can't publish Paid.** `store_owner_key` (`ui/src/state.rs:4668`) now returns `None` for a Ghost-Key-owned store, which blocks settlement of open legacy invoices. Fall back to the loaded state's owner for settlement.
7. **Tests:**
   - Add a certificate cap test at exactly `MAX_CERTIFICATE_PEM_BYTES`.
   - Add a `current_backing` tie-break test with backings in non-map order; today the tie-break passes only because of `BTreeMap` ordering.
8. **Low items:**
   - Mint a store key only after the tip check; orphan keys burn slots toward `MAX_STORE_KEYS`.
   - A failed creation leaves "Creating contracts..." with no retry.
   - A stale doc comment and a moved `cfg` in `ui/src/gateway/store_ops.rs` (~422).
   - A misplaced doc comment in `ui/src/gateway/migrate_ops.rs:240-256`.
   - The docs overstate the cap cost (certificates aren't checked) and claim the block hash is checked (only the height is). Either check the hash within the tip window, or fix the claim.
   - Note that on a slot clash the smaller encoding wins, not the newer record.
9. **Record as phase 2 deferrals** in `docs/design/entity-model.md` and `docs/untested-invariants.md`: the seller UI for closing a store, and the Settings message for a Ghost Key that already backs another store.

### Sequencing
- **Store keys and registrations live in the delegate.** Nothing carries delegate secrets across a delegate rebuild yet, so the next delegate change strands every store-key store: it can't be signed for, can't be found, and its backing can't be retired.
- **Phases 1b (custody) and 1c (the Ghost Key index) fix this.** So 1a, 1b and 1c will be built as one ordered stack and merged and published together. Nothing ships until the stack is complete.
- **#96 merges before this PR.** Both record a delegate V16, so this PR's delegate row is renumbered on rebase.

### Verdict
**State:** Needs Changes. Must Fix 1-3 trigger a re-review.

[AI-assisted - Claude]

sanity:
## Fix round for the review above

Head: `b1630ff11306f38e9c45d65978bfd2fb658280df`. CI is green on that SHA (fmt/clippy/test, address guard, wasm build, committed-wasm staleness, contract drift with the `contract-rekey-acknowledged` label).

### Must Fix 1: merge totality at the caps

The bound is now a normalisation step, not a refusal. `apply_parts` applies every part of an update first, then calls `normalize_backings`:

- **Backings:** the 64 backings with the smallest Ghost Key bytes are kept. Top-N over a slot-only ranking is associative, and a slot cut from a union ranks below 64 kept slots, so later merges can never bring it back.
- **Retirements:** a retirement is kept exactly when its backing is kept (`verify` now requires every retirement to name a held backing). A retirement is never dropped while its backing survives, so nothing is un-retired.
- **Closed flag:** its own set. The bound never touches it, nor details, listings or orders in the same delta, so a closure or a listing riding with overflowing backings always lands.

The merge no longer errors on caps, so fdev can no longer file a cap case as "inconclusive".

Evidence:
- Seeded merge-law tests over a pool of 72 backings, plus tests for the cut-backing un-retire case, the orphan retirement, the certificate at exactly the max size, and a tie-break that does not rely on map order.
- Native check over the reviewer's corpora: 0 merge errors.
- `fdev verify-merge`, rebuilt WASM:
  - `store-r98cap`: 724/724 hold, 0 violations, 0 inconclusive. Unions exceed the cap and the delta steps are included.
  - `store-r98race`: 809/809 hold.
  - `store-backing` and `store-backing-adv`: 0 violations, 0 inconclusive.
  - Across the other 22 corpora, the only violations are the known #81 cases (`store-adv` 6, `reputation-adv` 8). Everything else inconclusive comes from states the contract correctly refuses (`*-bad`, `*-noncanon`, `*-adv`, `store-v0`), not from caps.

### Must Fix 2 through Should Fix 9

- **Must Fix 2:** `BrowsingStore::payable()` (not closed, and has a store key) gates `StoreInvoices`, `buyable`, and the Bitcoin tab's `my_orders`. `PaymentBlocker::StoreClosed` is checked first.
- **Must Fix 3:** creation and move are single-flight behind a marker that is released only on success or failure. The section 6.2 check ("already backs <store>") runs before anything is signed.
- **Should Fix:**
  - `store_sk` is excluded from delegate export (`WithoutStoreKeys`).
  - Backings are dated six blocks behind the tip, and verdicts are recomputed on every tip update.
  - The legacy settlement path falls back to the loaded owner key.
  - The Low items are fixed. The phase 2 deferrals (close UI, Settings message) are recorded in `docs/design/entity-model.md`.

Per the new sequencing, phase 1b stacks on this branch as its own PR. 1a, 1b and 1c are meant to merge and publish together.

[AI-assisted - Claude]

sanity:
## Rebase, merge-law re-check (D) and skeptical re-check (M1, L1, L3-L6)

Head is now `6dc8f1503304b6018ea97c07bd1f0904fdf7c231`, rebased onto `main` (#95, #96, #100). CI is green on that SHA. The rebase and conflict resolution make the earlier reviews stale.

### Rebase: how each conflict was resolved

- **`legacy/harvest_delegate.toml` (in the re-key commit).** `main` had already recorded V16 = `4fae20e1` (#96's outgoing delegate), and this branch had its own V16 row for the same hash with a phase 1a description.
  - Kept `main`'s V16 row verbatim.
  - Added **V17 = `4d36b2a4`**, #96's own delegate, which is committed on `main` at `0173527` and published after #100. It carries the phase 1a description and `delegate_key` `9a6f7b44…`, derived as BLAKE3(code_hash) with the empty parameters, and checked against V16's recorded key the same way.
  - `b3sum` of `main`'s committed WASM: store `1e4eba43`, reputation `bf1f3c47`, mailbox `21b1bbed`, delegate `4d36b2a4`. The store V18, reputation V13 and mailbox V13 rows already name those hashes, so they are unchanged.
- **`ui/src/migrate/tests.rs` (pinned delegate hashes).** Kept `main`'s V16 comment for `4fae20e1` and appended V17 `4d36b2a4` (`git show 0173527:…`).
- **`ui/public/contracts/harvest_delegate.wasm` (binary).** Took the branch's side for that commit. Every artifact was rebuilt and synced at the head.
- **Follow-ups the rebase needed** (commit `e66ad06`):
  - #96's exhaustive secret classification in `common/src/delegate.rs` now names `CreateStoreKey`, `SignStoreUpdate`, `StoreKeyCreated` and `StoreUpdateSigned`, none of them secret-bearing, with samples.
  - The dropped-signature warning names its signer in words instead of `{:?}`, which #96's call-site guard refuses.
- **The re-key commit's subject still says "delegate V16".** After the rebase it records V17. The message was not rewritten.
- **`legacy/README.md`** now states the practice: a row records WASM that reached `main`, not intermediate builds that existed only on an open PR branch.

### D: arrival order could un-retire a key (`2144ddf`)

`normalize_backings` now ranks one set of slots: every Ghost Key with a backing or a retirement.
- It keeps the 64 smallest by bytes, and keeps a backing or a retirement exactly when its slot survives.
- The "retirement must name a held backing" rule is gone from `verify`, so an orphan retirement is valid.
- This is top-N over a slot-only ranking, so it is associative in any arrival order. A key's backing and retirement share a slot, so they are kept or cut together, and a cut slot never ranks back in.

New tests, each red with the old rule:
- `a_retirement_arriving_before_its_backing_still_retires_it` (by merge and by delta).
- `retired_keys_count_toward_the_bound`.
- The seeded past-the-bound test gains 150 stale-summary delta-order cases.

`fdev verify-merge` on the rebuilt store WASM (`cab5fb04…`, unchanged by the later M1 commit), with delta steps:
- **`store-r98race`**: the retirement-before-backing (R1) cases are restored (`R1`, `R1_closed`, both orders with `B1`, `B1later` and `L3_B1`). 1005/1005 hold; the bundle run holds 1461/1461; `delta_permutation_invariance` holds 38/38. 0 violations, 0 inconclusive.
- **`store-r98retire`**: your generator, native 59,319 triples clean. 1548/1548 hold; the bundle run holds 1999/2000; `delta_permutation_invariance` holds 504/504. 0 violations.
  - 1 inconclusive: the `ret1_noowner` delta (no owner named) applied to an empty store, refused because nothing can be verified without an owner. That refusal is correct.
- **`store-r98cap`**: 0 violations, 0 inconclusive; 2000/2000 in the bundle run. **`store-backing`, `store-backing-adv`**: 0 violations, 0 inconclusive.
- **`store-backing-bad`**: 0 violations, 416 inconclusive. All are invalid signatures refused: a closure, retirement or backing not signed by the store key, or a backing not signed by its Ghost Key. The orphan-retirement states in that corpus are now valid and hold.
- **The other 23 corpora** are identical to the previous run. The only violations are the known #81 cases (`store-adv` 6, `reputation-adv` 8). The remaining inconclusives are refused invalid states in `*-adv`, `*-noncanon`, `*-rr` and `store-v0`, plus the single known cases in `store-empty`, `store-rr`, `reputation`, `reputation-empty` and `mailbox-empty`, the same counts as before.

### B: skeptical re-check fixes (`6dc8f15`)

- **M1: a retry is idempotent.**
  - `CreateStoreKey` gains `ghostkey_fingerprint` (serde default).
  - The delegate remembers the key it minted for that Ghost Key's creation (`harvest:store_creation:{fp}`) until `RegisterStore` names it, and answers the same key to every retry from any tab or after a reload. The contract id derives from the store key, so a retry PUTs the same contract id.
  - Within a session the backing both keys signed is kept (`resumable_backing`) and reused, so a retry asks for no signature.
  - The section 6.2 check moved to the point where the key is known. A store owned by that key does not count against it, so an honest retry is never refused by the store it is re-creating.
  - Tests: `a_retried_creation_gets_the_same_store_key_until_registered`, `registering_a_store_ends_its_resumable_creation`, `a_retry_after_a_failed_publish_reuses_the_key_and_the_backing`, `a_retry_is_not_refused_by_the_store_it_is_re_creating`, plus the updated `a_ghost_key_that_already_backs_a_store_cannot_back_another`.
  - Limit: after a reload the signed backing is gone, so a retry signs a new one. It has the same key and contract id, and the two backings share a slot.
- **L1:**
  - A `HarvestDelegateResponse::Error` while the creation waits for its key releases it.
  - My Store shows a Cancel control beside "Creating contracts". Cancel withdraws the creation's signatures and keeps the key and backing for a resume.
  - Test: `a_stalled_creation_can_be_released`.
- **L3:** while a pre-revision-2 registration has not loaded, My Store shows "Loading your existing store…" instead of Create Store (`legacy_store_loading`, test `an_unloaded_legacy_store_holds_back_creation`).
- **L5:** the `BACKING_BLOCK_DEPTH` doc said the block "only has to prove the backing was written after it". It also orders backings, and now says so, with why dating six blocks back is harmless.
- **L6:** the `MAX_BACKINGS` doc now says that retired keys keep slots, so after more than 60 rotations the cut can fall on the current backing.
- **L4 (optional)** is not done: the Payments tab and My Store still describe unbacked stores separately.

Nine mutations over the M1, L1 and L3 guards were all killed.

[AI-assisted - Claude]

sanity:
## Re-check fix round (`35f8059`)

Head is `35f8059ea585925a49c50269634aeba7baee4ef2`. CI is green. The store, reputation and mailbox WASM are unchanged; only the delegate moved.

- **Cancel during the PUTs:** Cancel is now hidden and refused once the contracts are being published (`store_publishing`), so single-flight holds until the creation succeeds or fails.
  - Test: an added case in `a_stalled_creation_can_be_released`.
- **Section 6.2 across tabs:** `CreateStoreKey` gains `another_store` (serde default). With a fingerprint, the delegate refuses a NEW key to a Ghost Key that already has a store-key store registered on the device, unless `another_store` is set. An unfinished creation is still resumed. No UI sets the flag yet.
  - Test: `registering_a_store_ends_its_resumable_creation`.
- **`finish_creation` on the duplicate-registration branch:** test `a_repeated_registration_naming_the_key_ends_the_creation`.
- **Docs:**
  - The V17 row names `harvest:store_creation:`.
  - The `SignedSetV1` docs no longer mention the removed "retirement needs its backing" rule. They were at the lines you flagged in `backing.rs` (~339 and ~468).
- **Mutations:** 3 new, all killed.

The 1b items from the re-check are on #99.

[AI-assisted - Claude]

sanity:
## Section 6.2 now has both ways out (`99c4743`)

Head `99c474384e095a19e2c9d155eeee9afaa88d28b9`, CI green. The store, reputation and mailbox WASM are unchanged (`cab5fb04…`, `074b5758…`, `252d8355…`), so the merge-law results stand; only the delegate moved.

The review's point was right: the refusal told sellers to "retire it there first", and nothing in the product could produce a `Retirement` or set `another_store`. After 1c a duplicate backing would have made both stores unpayable with no way back.

- **Retire.** My Store offers "Retire this Ghost Key's backing" for a store this device holds the key for. The store key signs a `Retirement`, it is published to the store contract, and the delegate's new `RetireStore` request forgets the registration, so `CreateStoreKey` stops refusing and My Store stops listing it. The store key is KEPT, so a retirement made by mistake can be followed by backing the store again.
- **Escape.** A creation refused under 6.2 is offered back: "This Ghost Key already backs <store>. Open a second store under it anyway?" Confirming re-runs the creation with `another_store`, which turns off this tab's check and the delegate's. The prompt says both stores then read as unbacked until one backing is retired.
- **Docs.** The design doc records both, and that a duplicate backing is recoverable by retiring one; `docs/untested-invariants.md` gains the rows.
- **Tests:** `a_seller_can_retire_a_backing`, `retiring_needs_our_store_key_and_a_current_backing`, `a_retirement_that_does_not_verify_is_not_published`, `retiring_a_store_frees_the_ghost_key_and_keeps_the_key`, `a_refused_second_store_can_be_confirmed` (including that the second creation actually proceeds). Ten mutations, all killed.

[AI-assisted - Claude]

sanity (commented):

### Clean

### Should Fix

### Verdict

[AI-assisted - Claude]

sanity:

### Must fix

### Should fix

### Consider

### What is good

### Merge laws

### Verdict

[AI-assisted - Claude]

sanity:

### Must fix

### Should fix

### Consider

### What is good

### Merge laws

### Verdict

[AI-assisted - Claude]

PR #99, opened 2026-09-19
feat: store key custody, wrapped to each backing Ghost Key (#93 phase 1b)

sanity:
Part of #93, stacked on #98. Phase 1b: store key custody.

#96 (the fix for #94) is merged and published, and this stack is rebased onto it: a vault `SignResult` is logged as its bare name, so the wrap signature this PR routes through it does not reach the browser console. 1a, 1b and 1c merge and publish together.

## Problem

After phase 1a the store key lives on one device only, in the Harvest delegate that minted it. A second device cannot sign for the store. A delegate re-key loses delegate secrets (nothing drives the export handshake), so it leaves the store readable but unsignable. The store's inbox and record keys were also random per device, so two devices would publish different keys.

## Approach

This is the design the spike validated (`spike/store-key-custody`; `docs/design/entity-model.md`, "Store key custody" and section 6.3). What the build adds is recorded in the new "Phase 1b: what was built" section.

- **Wrapped copies in store state.** The store gains `copies`: its key's seed, sealed to each backing Ghost Key under a key derived (HKDF-SHA256, AES-256-GCM) from the vault's deterministic signature over `"harvest/store-key-wrap/v1\0" || store_vk`.
  - Each copy is signed by the store key, so nobody else can plant or replace one.
  - A copy is kept unless its backer is retired, so the retirement is the custody tombstone, in any arrival order. A copy may arrive before its backing and is kept; its backer takes a slot in the store-wide bound (after the #98 merge-law re-check).
  - At most four scopes are kept per backer, the smallest ones. The bound is a normalisation, like the backings bound, so the merge stays total.
- **The UI never holds the store key.** A wrap `SignResult` is recognised first and routed to `WrapStoreKeyFor` or `UnwrapStoreKey`, never to `pending_signatures` or any publish path. The signature travels in a redacting `WrapSignature`. The delegate checks the signature is this store's wrap signature, from this backer, under Harvest's scope. It checks a recovered seed is the store's key before keeping it.
- **When custody runs.** On every store state the UI decides:
  - wrap, if it holds the key and the current backer has no copy under the current scope;
  - recover, if it does not hold the key and a copy exists;
  - otherwise nothing.

  The decision is tried once per (store, backer) per session. A recovered store is registered again and reappears in My Store.
- **Derived keys.** The inbox (X25519) and record (RSA-2048) keys derive from the store key.
  - Conversation keys are derived with the store key, so every device reads the same mailbox.
  - Creation and edits publish the derived keys. `StoreInfoV1` gains `record_public_key`.
  - A device that derives a different record key from the published one says so, rather than trusting RSA keygen to be stable across `rsa` versions.

**Not in this PR.** The mailbox is still addressed by the backing Ghost Key (phase 1d changes that), so there is still no change-backing UI. The spike's `spike/store-key-custody` branch is superseded by this code.

**Re-key.** Store, reputation, mailbox and delegate WASM all move again. No legacy rows are added here: 1a's artifacts were never published, and the stack publishes together, so the only superseded generation is the one #98 already records.

## Testing

- **Common.** The spike's known-answer vectors, plus the checks on `from_sign_result` and wrap/unwrap. Copies tests:
  - the tombstone;
  - the scope bound;
  - a copy for an unbacked key is refused;
  - seeded merge laws over stores with copies.
- **Delegate.**
  - wrap and recover round trip;
  - a foreign signature or a copy that does not open keeps nothing;
  - the 64-key cap;
  - subkeys;
  - conversation keys derived from the store's inbox key.
- **UI.**
  - custody decisions (wrap, recover, nothing, backer not connected, asked once);
  - wrap-signature routing never touches the signing queue, and an unrequested one is dropped;
  - recovery re-registers, and a failure registers nothing;
  - a wrapped copy is published;
  - edits publish and wait for the derived keys;
  - the conversation request names the store key;
  - the record-key mismatch warning.
- **Mutations:** 28 scripted mutations over the new guards, all killed. The first run left five survivors: the unbacked-copy filter, the copy's store check, the recovered-seed check, the recovery cap, and the in-flight guard. Each now has a test, and the rerun killed all five. Rows are in `docs/untested-invariants.md`.
- **`fdev verify-merge`** on the rebuilt WASM, with new corpora `store-copies` (tombstone, same-slot clash, six scopes across two sides, copies racing retirements, closure and listing alongside) and `store-copies-bad`. Results so far:
  - `store-copies`: 565/565 hold. The bundle run with delta steps: 852 hold, 0 violations, 7 inconclusive. All 7 are the claim rule refusing a copies-only delta applied to a state with no owner (the copy's backing is not there), not a cap and not a law failure.
  - `store-copies-bad`: 0 violations. Its inconclusives are the invalid states being refused.
  - `store-r98cap`: 0 violations, 0 inconclusive (1044 bundle cases).
  - `store-r98race`, `store-backing`, `store-backing-adv`: 0 violations, 0 inconclusive.
  - The remaining 23 corpora are still running; I will post them as a comment.
- `cargo test --workspace --locked`, clippy (workspace, `address-guard`, UI on wasm32), the rehearsal `cargo check`, and `cargo fmt` are clean.

[AI-assisted - Claude]


sanity:
## `fdev verify-merge`: full run on this PR's WASM

Contract WASM: store `75962728…`, reputation `5a13e4f7…`, mailbox `689221a1…`. These are the same contract hashes as head `1a98fd0`; only the delegate moved in the last commit.

- **New corpora:**
  - `store-copies`: 565/565 hold. The bundle run with delta steps: 852 hold, 0 violations, 7 inconclusive. All 7 come from the claim rule, which refuses a copies-only delta applied to a state with no owner (the copy's backing is not there).
  - `store-copies-bad`: 0 violations. The inconclusives are the invalid states being refused.
- **Merge-totality corpora from the #98 review:**
  - `store-r98cap`: 0 violations, 0 inconclusive (1044 bundle cases).
  - `store-r98race`: 0 violations, 0 inconclusive.
- **The other 25 corpora** give the same numbers as the #98 fix-round run. The only violations are the known #81 cases (`store-adv` 6, `reputation-adv` 8). The remaining inconclusives are all states the contracts correctly refuse (`*-bad`, `*-noncanon`, `*-adv`, `*-rr`, `store-v0`).

[AI-assisted - Claude]

sanity:
## Rebase onto the reworked #98, and fixes for the security, design and testing lenses

Head is now `d907fee0ff3f8701151821cd15ad1c18572a7eb2`, rebased onto #98's new head `6dc8f15`, which is itself on `main` with #96 underneath. CI is green on that SHA. The rebase resolved conflicts, so the earlier reviews of this PR are stale.

### Rebase: how each conflict was resolved

- **`common/src/store.rs`, in the custody commit (`66819a8`), two hunks.**
  - The first was `normalize_backings` against the custody commit's `normalize_copies`. Took #98's single slot ranking and extended it to custody, per D: `backing_slots` also counts every copy's backer, so a slot cut drops that key's backing, retirement and copies together. `normalize_copies` then keeps a copy unless its backer is retired, plus the four-scopes-per-backer bound; it no longer requires a held backing.
  - The second was `verify`. Dropped the custody commit's re-added orphan-retirement check. Kept the copy checks except "must name a held backing": a copy for a retired key is still refused, and so are more than four scopes per backer.
  - So that commit already carries the new copy rule. Its old tests are updated two commits later (`7b7fade`), so the intermediate commits of this branch do not all pass on their own.
- **`ui/src/backing_flow.rs`, `on_store_key_created` (two commits).** Kept #98's M1 version (drop a stale resumable backing, the section 6.2 check with the retried store excepted, then set the key) and appended this PR's `request_store_subkeys(key)`. The second conflict was formatting only; kept HEAD.
- **`ui/src/components/my_store.rs`.** Took this PR's request list (`CreateStoreKey` only; the record and inbox keys derive from the store key) with #98's new `ghostkey_fingerprint: Some(..)`.
- **The committed WASM (binary).** Took the base's side in each intermediate commit, then rebuilt and synced at the head.
- **Follow-up the rebase needed.** #96's exhaustive secret classification now names the custody variants. `WrapStoreKeyFor` and `UnwrapStoreKey` are secret-bearing, since they carry the wrap signature and `WrapSignature`'s `Debug` redacts it; `GetStoreSubkeys` and the three responses are not.

### D applied to custody (`7b7fade`)

The first build dropped a copy whose backing was not held, so a copy that arrived before its backing was lost for good. That is the same arrival-order fault the #98 re-check found in retirements, and the old rule also fails the new stale-summary delta-order check.
- **New rule:** a copy is kept unless its backer is retired or its slot is cut. The retirement is still the tombstone in either order.
- **Tests:** `a_copy_arriving_before_its_backing_is_kept_and_a_retired_one_never_is` (by delta, both orders). The seeded copies test gains stores with no backing and 300 delta-order cases. Both are red with the old rule.

### Review fixes (`d907fee`)

- **#96 underneath.** The `custody_flow.rs` docs and the untested-invariants row now say what is true on this branch: a vault `SignResult` is logged as its bare name (`ghostkey_response_summary`, pinned by `a_vault_response_summary_prints_no_secret`). I confirmed that on this branch; the `SignResult` arm logs only "Received signature from ghostkey delegate".
- **Outstanding vault work.** `user_signature_under_way` counts pending custody, and custody starts only while nothing else waits on the vault. A deferred request starts when the vault answers.
  - So a custody refusal is not read as a watch refusal, and watch requests hold back behind custody.
  - Limit, recorded: a refusal names no request. If the seller starts signing while a custody prompt is open, one refusal drops both.
  - Tests: `custody_is_the_only_vault_prompt_while_it_waits`.
- **Blind signing: recorded as a gap, not built.** `handle_blind_sign` still uses the per-Ghost-Key RSA key while a new store's record uses the store-derived key. Feedback submission is unwired (#53). This is noted on `handle_blind_sign`, in the design doc, and in an untested-invariants row.
- **A record-key mismatch now blocks publishing.** It records the store in `record_key_mismatch`, and `start_store_edit_if_ready` refuses to publish that store's details from this device. A recovered key requests its subkeys, so the check runs before anything is published.
  - Tests: `a_record_key_that_disagrees_with_the_published_one_is_reported` (asserts the block), `a_recovered_store_is_registered_again`.
- **Recovery through any backer.** `custody_needed` recovers through any unretired backer that has a copy, the current one first, and wraps for the current backer only. Test: `recovery_can_use_any_unretired_backer`.
- **Re-evaluation on `GhostKeyList` and `StoreList`.** Tests: `a_ghost_key_list_starts_custody`, `a_store_list_starts_custody`.
- **Missing Harvest delegate.** `pending_custody` is cleared and the seller told. The wasm vault-request failure path notifies too. Test: `with_no_harvest_delegate_custody_is_dropped_and_said`.
- **Softened docs.** The "no request returns it" wording is gone from `store_keys.rs` and `CreateStoreKey`. The docs now say the Harvest origin is trusted with the seed in principle, because it supplies the wrap signature.
- **Delegate test for recovering against the wrong store key:** `recovering_against_the_wrong_store_key_keeps_nothing` (A's copy with B's signature, and A's signature for B).

Each new guard was mutated red: 8 over the review items, 2 over the custody arrival rule.

### `fdev verify-merge` on this head's WASM (store `4a32a70b…`)

- **`store-copies`**, which gains copy-before-backing states and both orders against `B1` and `B1_R1`: 729/729 hold. The bundle run with delta steps holds 1161/1161; `delta_permutation_invariance` holds 32/32. 0 violations, 0 inconclusive.
  - The 7 claim-path inconclusives from the previous run are gone: a copy-only delta to an empty store now lands.
- **`store-copies-bad`**: 0 violations, 55 inconclusive. All are invalid copies refused, either not signed by the store key or malformed.
- **Merge-totality corpora**, same as #98:
  - `store-r98race` with the R1 cases: 1461/1461.
  - `store-r98retire`: 1999/2000. The 1 inconclusive is the owner-less delta to an empty store, correctly refused.
  - `store-r98cap`: 2000/2000.
  - `store-backing` and `store-backing-adv`: clean. `store-backing-bad`: its inconclusives are refused signatures.
- **The other 22 corpora** match #98's run and the previous baseline. The only violations are the known #81 cases (`store-adv` 6, `reputation-adv` 8).

[AI-assisted - Claude]

sanity:
## Re-check fix round (`fa00167`)

Head is `fa0016797ad3b418db2d9bcb55258e57fe897e49`, rebased onto #98's `35f8059`. The only conflict was `my_store.rs`: I kept this PR's single `CreateStoreKey` and added `another_store: false`; the delegate WASM took the base side and was rebuilt at the head. CI is green. The store, reputation and mailbox WASM are unchanged, so the merge-law results stand.

- **HIGH: an in-session retry hung on "Creating contracts...".** `on_store_key_created` now fills the pending creation from the subkeys the session already holds (`fill_creation_from_subkeys`), and asks the delegate only when it has none.
  - The retry test's `started()` no longer hand-fills `rsa_public_key_der`; the test answers `GetStoreSubkeys` through the real path. It failed ("published again" never happened) before the fix.
- **MEDIUM: custody could stay pending for the session.** A custody request is now given up after `CUSTODY_TIMEOUT_MS` (= `SIGNATURE_TIMEOUT_MS`), and at once if sending it fails (`spawn_custody_request`). Either way the seller is told and the vault is free again.
  - Test: `an_unanswered_custody_request_times_out`. The send-failure path and the timer are wasm-only.
- **Record-key mismatch:** the check runs again when store details arrive (`recheck_record_key`), the warning is shown once, and the block lifts when a later check agrees.
  - Tests: `a_record_key_block_follows_the_published_details`, and `an_edit_is_refused_while_the_record_key_disagrees`, which drives a pending edit through `start_store_edit_if_ready`.
- **`recovered_registration` mailbox:** recorded, not fixed. The mailbox comes from the recovering backer rather than the founding Ghost Key, which only differs once a store has a second backer. That belongs to 1d, which re-addresses the mailbox; it is noted on the function and in the invariants doc.
- **Docs:** `StoreStateV1::copies` now states the copy rule (a copy is kept unless its backer is retired or its slot is cut). The V17 row names the 1b requests.
- **Copy-only backers past the bound:** test `copy_only_backers_past_the_bound_keep_the_smallest_slots` (72 backers, merge laws, at most 64 kept).
- **Mutations:** 6 new, all killed.

[AI-assisted - Claude]

sanity:
## Re-check fixes (`2d1ce3f`)

Head `2d1ce3ffcd9a66d3c54e44817619d473843ec5c8`, rebased onto #98's `99c4743`, CI green. The only conflicts were `common/src/delegate.rs` (the new `StoreRetired`/`RetireStore` variants next to the custody ones: kept both, and renumbered the custody variants in #96's secret classification) and `my_store.rs`'s `CreateStoreKey` (kept this PR's single request, with 1a's `another_store`). Contract WASM is unchanged, so the merge-law results stand.

- **A failed `GetStoreSubkeys` send is retried.** `request_store_subkeys` now goes through a path that clears `store_subkeys_requested` when the send fails, so the delegate is asked again instead of a creation waiting on a request nobody made.
- **Custody timestamps.** `take_custody` removes the pending request and its timestamp together, everywhere; a request with no timestamp is no longer treated as instantly expired; and the timeout re-checks while the request is pending rather than firing once, so a timer that goes off a millisecond early cannot strand it.
- **`fill_creation_from_subkeys`** answers whether it FILLED a creation, which is what its caller acts on.
- **Tests:** `a_finished_custody_request_leaves_no_timestamp_behind`, `filling_a_creation_answers_whether_it_filled_one`, plus the existing retry and timeout tests. Mutations killed for the timestamp handling and the fill semantics; the send-failure clearing is wasm-only and is recorded as such in the invariants doc.

[AI-assisted - Claude]

sanity (commented):

### Clean

### Should Fix

### Verdict

[AI-assisted - Claude]

sanity:

### Must fix

### Should fix

### Consider

### What is good

### Merge laws

### Verdict

[AI-assisted - Claude]

sanity:

### Must fix

### Should fix

### Consider

### What is good

### Merge laws

### Verdict

[AI-assisted - Claude]

sanity:
Fix round pushed and the index merge-law coverage gap closed. Full write-up on #101: https://github.com/freenet/harvest/pull/101#issuecomment-5751159589

This branch carries B4 (a failed subkeys *send* now releases the store creation, which previously needed a reload) and S3 (custody scheduling restarts after each response, so the second store no longer waits for an unrelated event). Both guards checked red by mutation; wasm32 build checked, since the sibling wasm-only paths are not reachable from the native tests. Head `d95c026`, CI green.

B1/B2/B3 on #98 remain open pending Ian's decision on the retire control. The stack merges as one unit.

[AI-assisted - Claude]

PR #101, opened 2026-09-19
feat: the Ghost Key index, a store list addressed by the Ghost Key (#93 phase 1c)

sanity:
Part of #93, stacked on #99. Phase 1c: the Ghost Key record, built as the minimal index that decision 6.8 settled on.

## Problem

A device that knows only a Ghost Key has no way to find the stores behind it. The store registry is per device, in the Harvest delegate. A reader applying "one current store per Ghost Key" (decision 6.2) sees only the stores its tab happened to load.

## Approach

**A new contract, `index_contract`, addressed by the Ghost Key alone** (`common/src/ghostkey_index.rs`, `contracts/index-contract`).
- An entry is the `BackingStatement` the Ghost Key signed through the vault when it backed a store, with that signature: the backer's half of the store's own backing. Publishing one needs no second vault prompt.
- The contract checks three things: the entry names this Ghost Key, its signature verifies, and the certificate is within the bound.
- It is grow-only, with one entry per store key and the smaller encoding winning a clash. It holds at most 64 entries, keeping the smallest store keys. That is top-N over the slot, so the merge is total and order-independent.
- There are no complaints in it (decision 6.8) and no related contracts (decision 6.7).
- **Retirements are not copied in.** The store key signs them and they live in the store. The index lists places to look; the store says whether a backing stands.
- **An entry is not a backing.** Only the Ghost Key signs an entry, so it can name any store key. Readers follow it to the store, where a backing needs the store key's acceptance.

**How the UI uses it** (`ui/src/index_flow.rs`):
- For each connected Ghost Key, it reads the index and loads every store listed. Custody then recovers the store key and registers the store again. The delegate registry stays as a cache.
- When a store loads, it reads the current backer's index and loads what it lists, so decision 6.2 covers the key's other stores.
- It publishes our store's backing into its backer's index unless the index already lists it, once per session. The same rule migrates existing stores onto the index.
- An index is routed by its contract id and used only if every entry verifies.

**Re-key bookkeeping:**
- `legacy/index_contract.toml` exists with no rows. `ui/build.rs` allows that one registry to be empty until its first superseded generation, and the migration probe is wired (`Artifact::Index`).
- The build script, the hash check, `harvest-addresses`, the address guard, CI and the drift guard all list the new artifact.
- The drift guard builds the merge base with this branch's script. That script now skips a crate the base lacks (`HARVEST_ALLOW_MISSING_CRATES=1`, set only there) and reports the artifact as NEW. On this PR the base cannot build the address tool, since `IndexParameters` does not exist there, so the address comparison falls back to its label-acknowledged path. It needs `contract-rekey-acknowledged`, which is set.
- No legacy rows are added for the other artifacts, because 1a, 1b and 1c publish together.

**Not in this PR:**
- "Still being checked" for a store whose backer's index has not arrived yet.
- Listing index-found stores in My Store before custody recovers them.
- Rehearsal coverage of the index contract.

## Testing

- **Common:** an entry is the key's own signed statement (wrong backer, wrong signer, wrong slot, certificate bound). Clashes resolve to the smaller encoding in either order. The bound keeps the smallest keys. Seeded merge laws and stale-summary delta order past the bound (200 + 200 cases). A delta brings the receiver up to the sender.
- **Contract:** validation, and state merge agreeing with delta merge, including the empty state.
- **UI:**
  - the index address matches the node's derivation;
  - a Ghost Key's index leads to every store it lists;
  - an index that does not verify is ignored;
  - a loaded store leads to its backer's index;
  - our store is published once and not when listed;
  - a store we do not hold is never published.
- **Mutations:** 19 scripted, all killed. The first run left one survivor, the backer check; a test for it was added and it is now killed.
- **`fdev verify-merge`:** new `index`, `index-cap` (unions past the bound, with delta steps) and `index-bad`, plus the full store, reputation and mailbox set on this WASM. Results will follow as a comment.
- **Other checks:** `cargo test --workspace --locked` (1038 passed), clippy three ways, the rehearsal `cargo check`, and fmt.

[AI-assisted - Claude]


sanity:
## `fdev verify-merge`: the index contract

Index WASM `f88f123c…` (address `BU1TmCGp5qNWkckhfja628koNcWcVMYwkHKwEU2nWhE4`, 45 bytes of parameters).

- **`index`** (honest indexes, a same-slot clash, several deltas off one base): 252/252 hold in the states run, 387/387 in the bundle run with delta steps, `delta_permutation_invariance` 27/27. 0 violations, 0 inconclusive.
- **`index-cap`** (unions past `MAX_INDEX_ENTRIES`, with delta steps): 249/249 and 357/357, `delta_permutation_invariance` 18/18. 0 violations, 0 inconclusive.
- **`index-bad`** (an entry by another Ghost Key, one signed by another key, one in the wrong slot): 0 violations. Its 70 inconclusives are the invalid states being refused.
- Native laws over every corpus's states and their merges: 0 commutativity, 0 associativity, 0 idempotence failures.

The store, reputation and mailbox corpora are being re-run on this branch's WASM (the store contract moved with `harvest-common`); I will post those numbers when the sweep finishes.

[AI-assisted - Claude]

sanity:
## The regression sweep on this branch's WASM

Store `c76f7a76…`, reputation `57af8e42…`, mailbox `fad8339c…` (every artifact moved with `harvest-common`).

All 28 store, reputation and mailbox corpora match the numbers on #99, case for case. The only violations are the known #81 unsigned-field cases: `store-adv` 6 and `reputation-adv` 8. Everything else inconclusive is a state the contracts correctly refuse (`*-bad`, `*-noncanon`, `*-adv`, `*-rr`, `store-v0`), plus the single known cases in `store-empty`, `store-rr`, `reputation`, `reputation-empty` and `mailbox-empty`.

Notably unchanged by phase 1c: `store-copies` 729/729, `store-r98race` 1005/1005, `store-r98retire` 1548/1548, `store-r98cap` 1444/1444, all with 0 violations and no new inconclusives.

[AI-assisted - Claude]

sanity (commented):

### Clean

### Should Fix

### Verdict

[AI-assisted - Claude]

sanity:

### Must fix

### Should fix

### Consider

### What is good

### Merge laws

### Verdict

[AI-assisted - Claude]

sanity:

### Must fix

### Should fix

### Consider

### What is good

### Merge laws

### Verdict

[AI-assisted - Claude]

sanity:
## Fix round 1 of 2 pushed, and the index merge-law coverage gap is closed with evidence

Heads: #99 `d95c026`, #101 `63bf641` (rebased onto the new #99). #98 is unchanged at `99c4743`. CI is green on all three.

### Fixed

- **B4** — a `GetStoreSubkeys` request that could not be **sent** cleared its marker and nothing else, so `store_creation_in_flight` stayed set and every later creation attempt was refused for the session; only a reload recovered. The delegate's *error* response already released the creation, and a send failure is the same situation, so it now takes the same path. The state change moved into `on_subkeys_request_failed`, off-target and tested, leaving only the send in wasm. The `docs/untested-invariants.md` row claiming this was untestable because wasm-only is corrected rather than left standing.
- **S3** — `start_custody_for` refuses while *any* custody request is pending, because `user_signature_under_way()` counts a non-empty `pending_custody`. So `start_custody_where_needed` starts exactly one store, and neither response handler restarted the scheduler; the wrap-signature branch in `state.rs` also returned before the restart that follows it. With several stores needing wrapping or recovery, all but the first waited for an unrelated event or a reload. `on_store_key_recovered` restarts from a wrapper around its whole body so no early exit can forget it.
- **S1** — `index_entries_published` did two jobs. The `never_fits` branch inserted into it purely to record "told the seller once", and the same set gates "published once", so a slot freed later in the session meant the entry was never published and nothing was said. The notice now has its own set.
- **Codex P2** — the same set from the other side: a failed index PUT left the store marked published for the session, so nothing retried and other devices could not discover it until a reload. `on_index_publish_failed` clears the mark.
- **S2** — the full-index message told the seller to "retire a backing you no longer use, or back this store with a different Ghost Key". A Ghost Key index is grow-only and never reads retirements, so retiring frees nothing; and there is no path to back an existing store with a different key. The message now says what is actually lost (discovery through that Ghost Key) and what still works (the store's own link).
- **S4** — `follow_indexed_store` derived only the current generation's instance id. A registered store gets `start_store_key_migration`, which probes the whole lineage; a store found through the index has no registration yet, so one still living under an older store-contract generation returned no state and the device never reached custody recovery — the case the index exists to serve.

All four new guards were checked **red by mutation**, not merely watched passing. The wasm32 build was checked on both branches, since the S4 and P2 edits are wasm-only code that the native test run cannot reach.

### The merge-law coverage gap, closed structurally

The sweep lived entirely in an agent's job scratch, which is deleted after 7 days, so the list of corpora it swept was effectively whoever-ran-it-last's shell history. That is how `index-clash` and `index-adv` came to be missing from it for a whole round — the round's riskiest change was the delta bound and the rewritten clash comparison, so the two corpora built to attack exactly that were the two not running, and the sweep reported clean.

`tests/merge-laws/` now holds the generator and `run.sh` with the full 35-corpus list, both index corpora included. The corpora themselves are not committed (about 550 MB, reproducible in minutes) and are gitignored. `gen/` is its own cargo workspace, like `tests/rehearsal`, so running the test tooling can never move a dependency version the contracts compile against and re-key an artifact as a side effect. The README records both ways this sweep has passed while covering less than it claimed, and the standing #81 violations so they are not re-reported as a regression.

Verified end to end from the committed location — generator builds, all five index corpora regenerate, sweep runs them:

| corpus | state pass | bundle pass (delta laws) |
|---|---|---|
| `index` | 252 / 0 violations | 387 / 0 |
| `index-cap` | 249 / 0 | 357 / 0 |
| `index-bad` | 85, 15 holds, 0 violations, 70 inconclusive | same — identical to the prior sweep; a corpus of deliberately invalid states |
| `index-clash` | 304 / 0 | **484 / 0** |
| `index-adv` | 568 / 0 | **904 / 0** |

Zero violations and zero inconclusive throughout apart from `index-bad`'s expected refusals. The delta laws are genuinely running on the two corpora that were silently skipped — `index-clash` gives `delta_determinism` 80, `delta_idempotence` 70, `delta_permutation_invariance` 40, `reconciliation_cycle` 45; `index-adv` gives 126 / 140 / 84 / 91.

One correction on process: an earlier run of mine reported an empty result for `index-cap`, and the cause was me — I rebuilt the contract WASM in the same worktree while the sweep was reading it. Those numbers were discarded and the sweep re-run with nothing else touching the tree. The figures above are from the clean run.

The rest of the sweep is carried forward rather than re-run: `git diff` from the reviewed head touches only `ui/` and `docs/`, plus comment-only edits in `common/src/ghostkey_index.rs`, and `scripts/build-contract-wasm.sh` gives `index_contract 0df754b5` before and after. `no contract or delegate address moved` is green on every head. So the contract bytes are unchanged and the earlier full sweep still describes this code, standing #81 violations included.

### Also

Three wrong comments in `common/src/ghostkey_index.rs` corrected (comments only, no re-key). One was a false statement about Rust semantics — `Ok` sorts before `Err`, not the reverse — which the implementer's own report had repeated. The fourth item, `apply_delta` being able to mutate and then return `Err`, is **not** fixed here: the fix changes `common/`, which every contract compiles against, so it re-keys all four contracts plus the delegate and invalidates the whole corpus set, for a case that is unreachable. Filed as #102 to ride with the next change that re-keys them anyway; the doc comment now states the exception honestly and points there.

### Still open

**B1, B2 and B3 are not fixed**, and will not be until Ian settles what the retire control should be — the question is with him now. The short of what I found tracing it: the design's story for retirement is raising your stake, which is "retire the old backing and add the new one", and the add-new half does not exist in the code, so a retirement can never be followed by a re-backing by any key. What the button actually does is permanently close the store. I also could not find any reader-side code that penalises a Ghost Key for backing two stores — the check exists only where a second store's creation is refused in the seller's own tab — which weakens the case that the escape hatch creates a situation needing this remedy.

The three PRs merge and publish as one unit, so nothing ships until that is settled and the new heads are re-reviewed.

[AI-assisted - Claude]

sanity:
## Re-review round 2 complete — all findings addressed, merge-law sweep clean

Heads: #98 `316b575`, #99 `a7a1fa5`, #101 `bbc4287`. The retire control is **removed** rather than fixed (Ian's decision, 2026-09-20 — see #104), so B1, B2 and B3 from the previous round are moot.

### Lenses

Two blind Claude reviewers plus `codex review --base 2d1ce3f`, then a fourth, targeted pass: a **systematic sweep for one defect class** after the same shape was found three times by three different reviewers, each by chance.

### The log guard, verified rather than asserted

The rebase required renumbering #96's fail-closed log-classification tables after removing the `RetireStore` arms — the one edit in the cascade where a wrong boolean is silently dangerous. A reviewer derived both tables from the enums rather than from the author's description (27 request arms, 28 response arms, indices unique and contiguous, both counts correct, every secret-bearing boolean checked against what the variant actually carries), then **proved the guard can fail**: five mutations, each red, each named by test — `WrapStoreKeyFor` flipped to `false`, `UnwrapStoreKey` flipped, a duplicated index, and each `*_VARIANTS` off by one. All caught by `no_delegate_message_prints_a_secret`.

It also reconstructed the pre-rebase heads from the branch reflogs and interdiffed: the renumbering was the **only** content change in #99, and #101 had none beyond line drift. The committed wasm reproduces byte-for-byte from a clean build, and `grep -a` on the artifact finds zero `RetireStore`/`StoreRetired` — so it is genuinely post-removal, not a stale blob.

Two guard weaknesses were found and fixed in passing: `StoreKeyWrapped`'s sample used the `Err` arm, so no `AuthorizedCopy` was ever built and the Debug scan had nothing to scan; and `carries_secret` scanned only for the array-of-integers encoding, leaving it **structurally blind to every `#[serde(with = "serde_bytes")]` field** — a variant classified secret-bearing on such a field would have passed vacuously.

### The optimistic-marker class

Three instances of one shape were found in this stack by three different reviewers, by chance. Treating that as a lower bound, a deliberate sweep found **nine more**.

The shape: *a set-membership or `Option` "already did this / in flight" marker is claimed BEFORE the operation it guards has succeeded, and the failure path does not release it* — so one transient failure disables a capability for the session, with no timeout and no recovery but a reload.

Fixed here (this stack's blast radius): a second `pending_store_edit` leak via `request_certificate`, which blocked all custody and all Bitcoin watch requests for a session; a custody request that failed to send and was never retried; a `browsing_stores` **placeholder** gap in one of this round's own fixes; and a retry storm this round introduced, now capped.

The remaining six are **#107**, the PR immediately after this stack. Three of them block a purchase.

The class itself is now written up in `docs/untested-invariants.md`: the shape, the tells to grep for, why native tests are structurally blind to it (the failure arms are `#[cfg(target_arch = "wasm32")]`), the caution that clearing a marker can be its own defect, and `ui/src/gateway/migrate_ops.rs` named as the model — it registers before every send too, but gives each one a failure deliverer **and** a timeout, and it is the only module here that does.

### Other findings addressed

From Codex: custody answers were matched by store key alone, so a late answer to a timed-out attempt could cancel a live one or register the wrong backer (which decides the mailbox); index and indexed-store GETs that failed to send left their markers claimed. Both fixed. **#105** filed rather than fixed, with reasoning: a store-key store's reputation contract is never probed for predecessors. It cannot seal a false verdict (markers are per-instance), cannot bite until that contract re-keys, and phase 1d re-addresses the record anyway — so it carries a hard gate instead. **#106** filed: `tests/rehearsal` has no working test target, pre-existing.

Every new guard was checked **red by mutation**. One pass caught only two of four; the other two had tests that passed while testing nothing relevant, and now have coverage.

### Merge laws

Full 35-corpus sweep at the final head, run clean with nothing else touching the worktree:

- **Zero violations and zero inconclusive everywhere** except the two documented baselines: `store-adv` 6 and `reputation-adv` 8, both on `state_commutativity` and `reconciliation_cycle`. Those are **#81**, pre-existing, with identical counts in sweeps weeks older.
- The previously-missing adversarial index corpora ran with the delta laws genuinely generating cases: `index-clash` 484, `index-adv` 904, both clean.
- No corpus was skipped.

The sweep tooling now lives in the repo at `tests/merge-laws/` rather than in agent scratch, with the full corpus list so the coverage gap cannot silently return. It also **snapshots the WASM it tests** and prints the BLAKE3 hashes, after a concurrent rebuild corrupted two runs during this review — the symptom being one corpus producing no output, indistinguishable from a real refusal.

### Verdict

Ready to merge, pending green CI on each head. The three merge and publish as one unit.

[AI-assisted - Claude]

sanity:
Reopening to retrigger CI — no workflow run exists at all for the current head (`55ca27ad`), which appears to be a missed pull_request event rather than a failure. [AI-assisted - Claude]

Issue #107, opened 2026-09-20
Six optimistic markers claimed before their operation succeeds, never released on failure (three block a purchase)


sanity:
Six instances of one defect class, found by a systematic sweep during the #93 phase-1 re-review. The four instances inside that stack's blast radius are fixed there; these six are fixed here, in the PR that immediately follows it.

All `file:line` below are at **`dcd8295`** (the #101 head at the time of the sweep).

## The class

**An optimistic marker: a set-membership or `Option` "already did this / in flight" marker is claimed BEFORE the operation it guards has succeeded, and the failure path does not release it.** One transient failure then disables a capability for the rest of the session, with no timeout and no recovery but a page reload.

The recurring tells: `.insert(x)` used as a gate via its `bool` return; any `spawn_local` whose failure arm only calls `warn!`/`error!` without clearing what the success path clears; `self.x = Some(..)` set as an in-flight guard before an async send.

**Why this is worth a sweep rather than a fix each time it is noticed.** Three instances were found in the #93 stack, each by a *different reviewer*, each by chance. A systematic sweep for the shape then found **nine more**. No reviewer had found the worst one.

## Blocks a purchase

These three stop a stranger completing a purchase, so they lead.

### 1. Tip contract subscription — `state.rs:7254` (`register_tip_contract_with_id`)

`bitcoin.subscribed` is claimed before the subscribe; the failure arm at `7256-7259` only `error!`s, and nothing releases it. No `TipView` means `tip_height(network)` is `None`, so `TipView::current_anchor` is `None` and **no invoice can be issued at all** — `begin_store_creation` also refuses with `NO_BLOCK_FOR_BACKING`, and payment depth cannot be measured. The only repair is a *change* of tip generation; re-resolving to the same generation is blocked by the marker.

### 2. Mailbox subscription — `state.rs:2471-2473` (`register_store_mailbox`)

`mailbox_to_store` doubles as the "already asked" record (its own comment says so, `state.rs:2456-2461`). It is claimed, then `subscribe_in_background("mailbox", …)` is called at `2474`; that helper (`state.rs:1867-1877`) only `error!`s on failure. Every later call early-returns "Already ours", so a failed mailbox GET means **the mailbox is never subscribed again this session**: the seller's Inbox stays empty and a buyer never fetches the reply they hold keys for. Purchases go through messaging, so this is on the critical path. No repair event exists.

[...]

PR #111, opened 2026-09-21
fix(purchase): three optimistic markers block a purchase

sanity:
## Problem

harvest#107 found six instances of one defect class in the #93 phase-1 sweep: a set-membership or `Option` "already did this / in flight" marker is claimed BEFORE the operation it guards succeeds, and the failure path never releases it. One transient failure then disables a capability for the rest of the session, with no timeout and no recovery but a reload.

This PR fixes the three that block a purchase outright, in priority order:

1. **Tip contract subscribe** (`register_tip_contract_with_id`, `state.rs`): `bitcoin.subscribed` claimed before the send; failure only logged. No `TipView` means no invoice can be issued at all, and the only repair (a change of tip generation) was blocked by the marker itself.
2. **Mailbox subscribe** (`register_store_mailbox`, `state.rs`): `mailbox_to_store` doubles as the "already asked" record, claimed before the send, failure only logged. A failed mailbox GET meant the mailbox was never subscribed again this session: seller's Inbox stays empty, buyer's reply never fetched.
3. **`store_state_unavailable` recorded from a send that never left the tab** (`subscribe_to_own_store`, `state.rs`) — the inverse of the class, and the one no reviewer found. A SEND failure was conflated with a TIMEOUT; `publish_store_details` reads that as `published_version = 0`, publishes at version 1, and the contract's last-writer-wins merge silently drops it as stale. Silent loss of a seller's edit. The timeout arm is unchanged and correct — a GET that truly went out and got no answer really is evidence.

## Approach

Each fix follows the model already in this codebase (`on_custody_send_failed`, `on_index_publish_failed`): split the state mutation out of the spawned send so it's reachable and mutation-checkable off-target (native tests are structurally blind to `#[cfg(target_arch = "wasm32")]` failure arms), bound the retry count with a small cap and a single notification at the cap (a send failure is not a refusal — nothing was asked — so unconditional clearing would retry at whatever event re-drives the caller, with no backoff), and leave the marker set past the cap so it stops retrying.

`subscribe_in_background`, previously generic but with exactly one caller (mailbox), is replaced with a dedicated `subscribe_to_mailbox` so the mailbox fix has its own failure path.

## Testing

Five new tests, one per fix plus a stale-callback guard test for the mailbox fix:
- `a_tip_subscribe_failure_retries_a_bounded_number_of_times`
- `a_failed_own_store_subscribe_send_does_not_mark_it_unavailable`
- `a_failed_own_store_subscribe_send_retries_a_bounded_number_of_times`
- `a_mailbox_subscribe_failure_retries_a_bounded_number_of_times`
- `a_stale_mailbox_failure_does_not_undo_a_different_stores_mapping`

**Every new guard mutation-checked red**: each was temporarily reverted to the original bug (or had its ownership guard dropped) and confirmed the new test fails, then restored.

`cargo test --workspace`: all green (harvest-ui 625/625, up from 620). `cargo fmt --check` and `cargo clippy --all-targets` clean. `cargo check --target wasm32-unknown-unknown` clean (the actual bug surface).

No contract or delegate WASM touched — pure `ui/src/state.rs` plus a `docs/untested-invariants.md` update. Contract hashes reverified unchanged: `index_contract 0df754b5`, `store_contract b5eddce7`.

## Scope

The other three instances from #107 (`bitcoin.subscribed`'s watch-contract flavor, conversation key requests, buyer conversation recalls / `stores_remembered`) are left for a follow-on PR, per triage — these three block a purchase, those three degrade the app.

Closes part of #107 (three of six; tracking issue stays open for the remainder).

https://claude.ai/code/session_01KsdWgxxp8FNBaQZDdpguYK


sanity:
## Consolidated review record (posted after merge — process gap, see below)

This PR went through four review passes across four commits. None of it was posted here as it happened; that is itself a gap, addressed at the end of this comment.

### Round 1 — commit `71326c9` (the three original fixes)
Two Claude subagents (`freenet:skeptical-reviewer`, `freenet:code-first-reviewer`) plus `codex review --base origin/main`, run in parallel.
- **Codex found a real P1**: the bounded retries for the mailbox and own-store subscribes depended on an external `StoreList`/`ListStores` re-answer to re-drive them after releasing the marker. That event fires only once per ghostkey connect for the seller's own store, and a **buyer's** mailbox subscribe (from `deliver_to_seller`, or the once-per-session conversation recall) has no such event at all. A buyer sending one message and then waiting would get exactly one attempt, regardless of marker state — the exact purchase-blocking symptom the PR exists to fix.
- Both Claude lenses independently flagged: the failure counters were never reset on success (a few transient failures spread across a long session would permanently exhaust the retry budget), and `on_mailbox_subscribe_failed` bumped its counter before checking ownership (a stale/foreign callback could nag).

### Fix → commit `54faf05`
Made `subscribe_to_mailbox`/`subscribe_to_own_store` retry **themselves** on a timer rather than depending on the external event. Added reset-on-genuine-arrival for all three counters. Reordered the mailbox ownership check to run first. Five new tests, all mutation-checked red at the time.

### Round 2 — verification on `54faf05`
One more `skeptical-reviewer` pass, specifically checking whether the round-1 fix was correct.
- **Found a new, real, high-severity bug — empirically reproduced, not just traced.** The internal timer retry calls `subscribe_to_mailbox`/`subscribe_to_own_store` directly, never going back through `register_store_mailbox`/`note_store_subscribed`. But the failure handlers still released their marker on *every* failure below the cap (the shape meant for an external re-registration to reclaim it). So: attempt 1 fails → marker released → attempt 2 fires → its ownership check sees no mapping → treats it as "stale/foreign" → silently gives up with the counter stuck at 1 and **no notification ever shown**. Worse than the pre-#107 bug, and on exactly the buyer path the P1 was about.

### Fix → commit `7d11d11`
Both markers now stay claimed for every failure below the cap; released only when the cap is finally reached. Rewrote both bounded-retry tests to match the real recursive-retry call pattern (no re-registration between attempts — the previous test version's re-registration is what hid round 2's bug). Mutation-checked red.

### Round 3 — verification on `7d11d11`
One more `skeptical-reviewer` pass, trying hard to find a third bug in the same interaction.
- **No third blocking bug found**, after tracing the full attempt sequence, checking for cross-trigger races with `index_flow.rs`'s separate marker-sharing mechanism, and re-verifying the success path.
- **Did surface one real, non-blocking gap**: the failure counters are lifetime-scoped, reset only on genuine arrival — never on a fresh external reclaim. So a store/mailbox that ever crossed the cap once early in a session got exactly one silent attempt on every later re-trigger for the rest of the session, contradicting the doc comments' "start over from scratch" claim.

### Fix → commit `bc88758` — **merged without its own independent review**
`note_store_subscribed` and `register_store_mailbox` now reset their failure counter on every fresh claim. Strengthened both tests to assert the reclaim grants a genuinely fresh budget. Mutation-checked red by the author. **This specific increment was self-certified and merged without a fourth independent lens** — a process gap, not a quality one as it turned out.

### Round 4 (after merge) — independent lens on `bc88758` alone
Run after the gap above was noticed. Verdict: **no blocking issue, no republish required.** Verified the reset placement fires only on a genuine fresh claim (not a re-affirming call), verified key types (mailbox-keyed vs store-keyed) are correct, reproduced both claimed mutations plus a wrong-key variant that wasn't claimed, confirmed the strengthened tests aren't papering over anything, and checked the interaction with `index_flow.rs`'s separate `subscribed_stores`-sharing mechanism (safe, by claim exclusivity). `cargo test -p harvest-ui`: 628/628. wasm32 check clean.

Three **consider**-level findings from this pass, not fixed yet:
1. The notification bar (`notifications: Vec<String>`, never cleared/deduped/dismissed) can now accumulate one line per cap-crossing, where before the reset it could fire at most once per store/mailbox per session. Direct consequence of granting a fresh budget rather than a new bug; cheap mitigation is a dedupe or a dismissible bar.
2. Neither test covers a second cap-crossing, so this repeat-notify behavior has zero coverage in either direction.
3. `on_own_store_subscribe_send_failed` has no ownership guard analogous to the mailbox side's (currently unreachable given claim exclusivity, but that safety property lives in a different file, `index_flow.rs`, and isn't pinned anywhere).

### On the process gap
Three rounds (2-3 Claude lenses plus Codex) is beyond what Light tier asks of a UI-state-handling diff — the tier this PR should have been judged, explicitly, at the start (it wasn't; that omission is exactly how the last increment fell below Light's floor without anyone noticing). The substance was strong: round 1 caught a genuine P1, round 2 caught a real high-severity bug *in the round-1 fix itself*, reproduced rather than merely argued. Every guard across all four commits was mutation-checked red. The two real defects: the final merged commit (`bc88758`) had no independent read before merging, and none of rounds 1-3 were written down anywhere until this comment. Both are now closed — this comment is the record, and round 4 is the missing read.

[AI-assisted - Claude]

PR #113, opened 2026-09-22
fix(purchase): the three remaining #107 optimistic markers

### LINKED ISSUE #107: Six optimistic markers claimed before their operation succeeds, never released on failure (three block a purchase)

sanity:
## Problem

harvest#107 found six instances of one defect class: a set-membership or `Option` "already did this" marker claimed **before** the operation it guards succeeds, with the failure path never releasing it. #111 fixed the three purchase-blocking instances. This PR fixes the three remaining ("degrades the app" tier):

1. **`bitcoin.subscribed`'s watch-contract flavor** (`register_watch_contract`, `watch_purchase_addresses`) — a manual watch's row stayed "watching..." forever on a send failure; a purchase's payment address was never watched again this session, so a buyer could never see whether they'd already paid.
2. **`pending_conversation_key_requests`** (`ask_for_conversation_keys`) — a stranded entry made a peer's messages permanently unreadable for the session.
3. **`buyer_conversations_recalled` / `stores_remembered`** (`recall_buyer_conversations`, `remember_store`, and the queued `RememberStore` path in `sync_remembered_stores`) — a buyer's earlier conversations with a store were never asked about again; a remembered store silently never joined the user's durable list.

## Approach

All three follow #111's matured pattern (see `docs/untested-invariants.md`'s optimistic-marker section): split the state mutation out of the spawned send so it's testable and mutation-checkable off-target, a dedicated send function that retries itself on a timer (the natural external re-trigger — a store's state re-arriving — is not periodic, unlike the tip contract's genuine 10-minute bridge refresh), a bounded per-key failure counter left CLAIMED across failures below the cap and released only at the cap (so a concurrent external call can't race the internal retry chain — the exact regression #111's round 3 caught for the mailbox case), and the counter reset both on a fresh claim and on genuine arrival.

**One deliberate asymmetry, called out explicitly rather than left implicit:** `pending_conversation_key_requests` (item 2) is fixed differently — an unconditional release + notify, with no bounded cap. That marker is an ephemeral per-attempt map, not a persistent "asked once ever" latch, and this file already treats a delegate-returned `Err` answer the same way (`on_conversation_keys`, doc comment: "this is the transient case... and it records nothing, so the next mailbox update tries the whole set again" — deliberate, unbounded, by design). Mirroring that for a send failure is consistent; bolting a bounded-cap shape onto only the send-failure half of one marker, while its sibling delegate-error half stays uncapped, would be the inconsistency.

Also fixed the queued `RememberStore` path in `sync_remembered_stores` (sent once a delegate registers) — it shares `stores_remembered` with the direct `remember_store` path and has the identical defect, reachable at delegate-registration time. Not spelled out separately in #107's text, but the same instance of the same marker.

## Testing

11 new tests (628 → 639). 9 of the new guards mutation-checked red (all four `on_*_failed` release-on-cap methods, the buyer-conversation-recall ownership guard, three genuine-arrival resets, one fresh-claim reset, the `sync_remembered_stores` queued-path routing) — every mutation caught by its test, all reverted. Did not separately mutation-check `watch_purchase_addresses`'s fresh-claim reset or `remember_loaded_store`'s / `buyer_conversations_to_recall`'s fresh-claim resets: identical one-line `.remove()` pattern already verified for a sibling caller.

`cargo test -p harvest-ui`: 639/639. `cargo fmt --check` clean. `cargo clippy --all-targets` clean. `cargo check --target wasm32-unknown-unknown` clean. No contract or delegate WASM touched — pure `ui/src/state.rs`. Contract hashes reverified unchanged: `index_contract 0df754b5`, `store_contract b5eddce7`.

## Review tier

**Light, with a second lens** (per the standing review rule): no contract/delegate WASM, no wire format change, self-contained diff to one file — but real logic (concurrent self-retry timers, bounded caps) warrants a second, distinct adversarial read alongside the first. Requesting a skeptical lens (retry/race conditions) and a code-first lens (intent vs. implementation, especially the item-2 asymmetry above).

Closes #107.

[AI-assisted - Claude]


sanity:
## Consolidated review — Light tier + a second lens (as declared in the PR description)

Three lenses ran: **skeptical** (adversarial, retry/race focus), **code-first** (read the code before the description, checked for intent-vs-impl mismatches), and **Codex** (external, non-Claude — ran successfully, verified as a substantive pass and not a truncated/quota failure).

### Codex
Zero findings. Verified full diff read (904 lines), rustfmt clean, ran the test binary (639/639) independently.

### Findings addressed (commit 5111059)

1. **[should-fix, code-first + skeptical]** `docs/untested-invariants.md`'s optimistic-marker section still described the three remaining #107 instances as an open follow-on, and didn't name the four new `on_*_failed` functions. Fixed: the section now says all six #107 instances are fixed, lists the new functions, and documents the `pending_conversation_key_requests` exception.
2. **[should-fix, skeptical; consider, code-first — same underlying point from both lenses]** `sync_remembered_stores`'s "no delegate registered" arm is unreachable in production today (its only call site fires immediately after `harvest_delegate_key` is set, which nothing outside tests ever resets to `None`), and unlike its two siblings' identical defensive arms, had no comment saying so — and its regression test exercises a manual re-invocation loop that doesn't reflect how production actually calls it. Fixed: added the same "unreachable today, guarded anyway" documentation as its siblings, and annotated the test to make clear the repeated call is a stand-in for a hypothetical future re-drive, not a claim about current behavior.
3. **[consider, skeptical]** Doc comment said `bitcoin.subscribed` has two claimants; it actually has three (the pre-existing tip contract is the third, via its own separate counter and separate — deliberately different — release logic). Fixed: clarified inline.

### Findings dismissed, with justification

- **[skeptical]** `on_tip_subscribe_failed`'s release-below-cap / stay-claimed-at-cap shape is inverted relative to this PR's new functions. Confirmed: this is pre-existing code from #111 (out of this diff), and the inversion is deliberate — tip's re-trigger is the bridge's genuinely periodic 10-minute refresh (so it's safe, and correct, to release eagerly below cap and only stop at the cap), while this PR's three markers have no periodic external re-trigger and instead use a self-timer retry chain, which is why *they* need the opposite shape (stay-claimed-until-cap, so a concurrent external re-claim can't race the internal chain — the exact regression #111's round 3 caught). #111's own test (`a_tip_subscribe_failure_retries_a_bounded_number_of_times`, "past the cap it stops retrying, leaving the marker set") confirms this is intentional and tested, not a latent bug. Not a regression, not touched by this diff, out of scope for this PR.
- **[consider, skeptical]** Unbounded notification growth for conversation-key send failures during a sustained outage. This mirrors `on_conversation_keys`'s existing `Err`-answer arm, which already has the identical characteristic and predates this PR — not a new risk introduced here. (Also previously flagged, deferred, and explicitly not-yet-actioned in the predecessor PR's post-merge review, section 3 of the handoff doc.)
- **[consider, skeptical]** The delegate key is captured once at the start of each retry chain and reused on retries rather than re-read. Not exploitable: `harvest_delegate_key` never changes after registration (verified independently by both lenses).

### Verification after the fix commit

`cargo test -p harvest-ui`: 639/639. `cargo fmt --check`: clean. `cargo clippy --all-targets`: clean. `cargo check --target wasm32-unknown-unknown`: clean. Contract hashes reverified unchanged after both commits: `index_contract 0df754b5`, `store_contract b5eddce7`.

[AI-assisted - Claude]

PR #115, opened 2026-09-22
fix(purchase): let a buyer publish Paid, and keep the chain tip fresh

### LINKED ISSUE #74: The tip contract is read once too, and a stale one blocks every settlement
Found while reviewing #73, which fixes the same shape for an order's **address** contract. Filing rather than fixing there, for a reason given below.

## The shape

`register_tip_contract_with_id` GET-and-subscribes the chain-tip contract once, behind the insert-only `bitcoin.subscribed` set, and nothing ever asks again. That is exactly the pattern #67 was about.

It gates settlement completely:

- `settled_orders` returns early, with no log, when there is no `signed_tip` for the order's network.
- `assemble_on_chain_proof` measures confirmation depth against that tip, so a stale one under-counts depth and refuses a payment that is deep enough.
- `worth_watching_for_payment` answers `true` with no tip, so the address re-reads added in #73 have no bound at all until one arrives.

So a node serving a stale or absent tip reproduces #67's exact symptom -- a confirmed payment reading as awaiting payment -- with the address contract perfectly fresh.

## Why #73 deliberately did not fix it

The two contracts are not equally exposed. **A tip contract gets an update per block, so its own update stream heals a missed fan-out within about ten minutes. An address contract may receive exactly one update in its entire life**, which is what makes it the pathological case and why it was fixed first.

That argument is worth keeping, and it is also not a guarantee: it assumes the update stream is working, which is the assumption that failed in the first place.

## What would close it

The same shape as #73: re-read the tip contract on a timer while any order is unsettled. `crate::address_reread` is already the decision half and is not address-specific; the bound would be "some order is awaiting payment" rather than a per-address one.

Worth doing when someone is next in this code, not urgent on its own.

[AI-assisted - Claude]

### LINKED ISSUE #75: A buyer cannot publish the Paid transition, so an order only settles when the seller opens the app

sanity:
Phase 0 of #53 — the two prerequisites its buyer half rests on. Both are UI-only and neither re-keys anything.

## Problem

**#75.** Every buyer-side feature in #53 keys off the *published* record reaching `Paid`. Today only the seller can publish it, so in the exact case #53 exists for — the seller takes the money and vanishes — the record never reaches `Paid`, no reader-side window ever starts, and the complaint is never offered. Building the rest of #53 on today's behaviour would ship a feature that is inert precisely when it matters.

It is also silent: the buyer's own card reads the live address view and says "Payment seen on chain" either way, so the gap is visible only to a third party reading the store contract.

**#74.** `register_tip_contract_with_id` GET-and-subscribes a network's tip contract once, behind the same insert-only marker #67 was about, and nothing ever asks again. A stale tip is not a cosmetic wrong number: `settled_orders` returns nothing at all without a `signed_tip`, and `assemble_on_chain_proof` measures confirmation depth against it — so a stale tip **refuses a payment that is deep enough**, reproducing #67's exact symptom with the address contract perfectly fresh. It also under-counts every window the wider #53 design measures against it, so nothing ever goes overdue.

## The premise, verified before anything was written

The design flagged "no origin check in the store contract" as an *inference* read from `apply_delta` rather than from `update_state` end to end. It was checked directly, all four entry points of `contracts/store-contract/src/lib.rs`:

- `update_state` reads only `parameters`, `state` and the update payloads. No requestor, sender or origin is available to it at all.
- `validate_state` is documented as a pure function of state + parameters, and deliberately consults no related contracts.
- `OrdersV1::apply_delta` (`common/src/store.rs:846`) verifies every record against `owner_key(parent_state)` — the owner the **current state** names — and nothing else.
- `AuthorizedOrder::fields_used` (`common/src/payment.rs:1493`) is the authority table: `Paid | PaymentReversed => (payment_proof used, status_signature UNUSED)`. `Paid` carries Bitcoin evidence and is signed by nobody, and `verify_unused_fields_absent` pins the signature fields to `None`, so a third party cannot even attach a spurious one.

The `Requestor::WebApp` scoping that appears in signatures is a scope on the *seller's* terms signature, not a check on who publishes. So the gate was purely client-side, as the issue says.

### #75

`store_ops::owned_store_key` resolved the contract key out of `my_stores` and failed with "this store is not one of yours". Evidence-authorized transitions stop going through it:

- **`reconstruct_store_key`** split out of `store_contract_key`. It needs no registration — a `ContractKey` is an instance id plus a code hash, the instance id *is* the store contract id, and the code hash is the bundled contract's. (`store_contract_key`'s own doc comment already said this.)
- **`settlement_store_key`** + **`submit_settled_order_by_id`** use it, and still prefer a *recorded* key for a store we do own, since the reconstruction is right only for a store published under the contract this build bundles.
- **`settlement_owner_key`** falls through to the owner the store's loaded state names — the key `apply_delta` will actually check the record against.
- **`publish_settled_orders`** drops its ownership early return.

Everything a seller **signs** — listings, store details, invoices — still goes through `owned_store_key`. Resolving a key without a registration there would only build an update nobody could sign, so the new function is named for the transition rather than for the key it returns.

`AppState::owns_store` is removed along with its only caller: leaving a "do we own this store?" helper beside the settlement path invites the next author to re-add the gate.

### The guard #75 would otherwise have bypassed

`docs/untested-invariants.md` recorded this against itself: *"It holds only because only the seller's tab publishes Paid today, so #75 would bypass it."*

`settlement_hold` withholds a provable payment when another uncancelled order on the same address has an overlapping window, and it was fed by `refresh_same_address_orders`, which scanned **own stores only**. Correct while only a seller could publish; blind in exactly the wrong direction once a buyer can, because a buyer's two purchases from one address-reusing seller both sit in *that seller's* store. One payment would have auto-published `Paid` on both.

`our_order_addresses` is refactored into `our_orders` — one definition of "ours" — and the scan widened to our orders in every store on screen. That keeps the property the narrower rule was protecting, which its comment stated: an order this node is not party to still cannot stall anything, whoever wrote it.

Worth saying plainly, because a future reader will not otherwise see the argument: **that file predicted this against itself.** It did not merely describe the hold — it recorded the precondition the hold rested on, and named #75 as the thing that would remove it. The hazard was therefore found by reading the docs, not by a reviewer catching it after the fact, and it is a correctness bug that did not exist before this diff and would have existed after it. A change that *introduces* a hazard owns it, so this is part of #75 rather than a follow-up. Keeping `untested-invariants.md` current is what made that possible.

### #74

`address_reread::AddressRereads` is keyed by contract id and knows nothing about addresses, so it is reused unchanged in its own tracker (`bitcoin.tip_rereads`; a shared one would have each set pruning the other's entries). The bound is "some order is unsettled" rather than per-address: the networks of `due_address_rereads`' own `wanted` list, so there is **one** rule for when an order stops being worth asking about rather than two that can disagree. Both share the existing tick, which is where the expensive walk (it decrypts conversations) already happens.

## Testing

645 tests, up from 639. `cargo fmt` clean, `cargo clippy --workspace --all-targets` clean, `wasm32-unknown-unknown` check clean.

Every new guard mutation-checked red, each mutation reverted cleanly:

| Mutation | Test | |
|---|---|---|
| ownership gate restored in `publish_settled_orders` | `a_buyers_tab_publishes_the_settlement` | RED |
| `settlement_owner_key` loses its loaded-state fallback | `a_settlement_names_the_owner_the_stores_own_state_names` | RED |
| `refresh_same_address_orders` scans own stores only | `one_payment_does_not_settle_two_of_a_buyers_orders_on_one_address` | RED |
| `tip_contracts_to_reread` loses the unsettled-network bound | both tip tests | RED |
| `send_tip_rereads` stops recording the ask | `an_unsettled_order_asks_for_the_chain_tip_again` | RED |
| `register_tip_contract_with_id` keeps a replaced id in the map | `only_a_live_tip_...` | RED |

**One mutation SURVIVED, and it is reported here because acting on it changed the code.** A `retired_contracts` filter in `tip_contracts_to_reread` could never fire: `register_tip_contract_with_id` already removes a replaced generation from `tip_contract_network` before inserting it into `retired_contracts`, so the two sets are disjoint by construction. The dead filter is deleted, the doc comment says why, and the test now asserts the map deletion — which *is* red when removed. A guard that cannot fire is one nobody can verify.

`docs/untested-invariants.md` is updated rather than left describing the old behaviour: the #75 row's recorded gap is marked closed with its new pins, the twin-hold row's stated limit is rewritten, and a new row covers the tip re-read. The residual is stated there too — `settlement_store_key` and `submit_settled_order_by_id` are wasm-gated, so the key a buyer's send actually carries is not exercised off-target; what is tested is that the reconstruction agrees with `store_contract_key`'s reconstructed answer in instance **and** code hash.

Also narrowed, as #75 asks: `settled_orders`' "any reader holding the claims can publish" paragraph had described something the app did not do since it was written. It now says which half was a statement about the contract and which about the app, and that both are true now.

## Re-key check

`scripts/build-contract-wasm.sh` run at the base commit and again at this head. All five artifacts byte-identical:

```
store_contract       b5eddce776f5f38e47682e94ede5bbd7f4d12fb0acd23bc826dcbff81708737b
index_contract       0df754b5c0066bf4ed02b800a293b8eb219ff3c4178af1a0239a8cf7eec7faec
reputation_contract  57af8e42fb73260f5b38e9fc27116eaa9e4183b464c8b921e971adfb4dff72db
mailbox_contract     fad8339c66fe289d07a7b651fef92f55aa36444862c59151523d0e68dac93cf2
harvest_delegate     09507e3f4f1432dd746cf4644aaa81678e25314798dc6069ce38409d1499fddd
```

No `legacy/*.toml` row is needed, and the merge-law sweep still describes live code.

## Review tier

**Light plus additional lenses**, declared before reviewing. The diff is UI-only, touches no contract or delegate WASM and no wire format — but it changes *who may publish a payment transition*, so a single read is not enough. Three independent blind Claude lenses (authorization, skeptical, testing) plus an external non-Claude pass; the consolidated result is posted on this PR.

Closes #75
Closes #74

https://claude.ai/code/session_01KsdWgxxp8FNBaQZDdpguYK

[AI-assisted - Claude]


sanity:
## Consolidated review

**Tier: Light plus additional lenses**, declared before reviewing and escalated past the initial read once the diff grew past a minimal key swap — this changes *who may publish a payment transition*.

**Lenses that ran, all blind to each other:**

| Lens | Model | Result |
|---|---|---|
| Authorization (code-first) | Claude Opus | **1 blocking**, 2 consider, 1 perf should-fix — then caught my incomplete fix for its own blocking finding, twice |
| Skeptical (adversarial) | Claude Opus | ran; remaining scope reported below |
| Testing (coverage/vacuity) | Claude Sonnet | 2 should-fix, confirmed the new tests non-vacuous by independent trace |
| **External, non-Claude** | `codex review` | **3 findings** — 2 fixed, 1 filed |

The external pass ran and did not need substituting.

---

### Blocking — found by the authorization lens, fixed

`settled_orders` walked the store's **whole order book**, on the argument that `bitcoin.addresses` holds only our own addresses so a stranger's order could never assemble. That argument answers *"whose address"* when the question is *"whose order"*, and address reuse is precisely where those part company — which is the case `settlement_hold` exists for.

Meanwhile this PR had narrowed the guard's input to our orders. Before #75 the two sets coincided, because the only party who could publish was the store owner, whose own orders **are** every order in their store. Once a buyer publishes, the publisher's set is the seller's entire book while the guard's set is one or two records.

No attacker required: a seller reuses one address for buyer B's and buyer C's orders, equal price, overlapping windows. B pays once. `assemble_on_chain_proof` ties a payment to an order only by script, amount, window and depth — nothing distinguishes C's. So B's tab publishes `Paid` for an order B is not party to and nobody paid. **Unrecallable**, since `Paid` is monotone under `merge_order` and only `PaymentReversed` (needing a bridge-signed retraction) outranks it; and the seller's own tab then skips it as no longer `AwaitingPayment`, so the seller loses the one chance their hold had to catch it and ships against money that never arrived.

### The fix took three rounds, and the middle one is the lesson

1. **First fix:** one rule, `orders_we_may_settle`, used by both sides — publish set and twin set made *equal*.
2. **The lens rejected that as incomplete**, and was right. Set equality is not the property. Confusability is defined by `(network, script, window)` — by the **chain**, not by party — so the twin set must be a **superset** of the publish set. Narrowing the publish set closed the direction where we publish someone else's order; the mirror stayed open, where a **stranger's** order at the same address is invisible as a twin of **ours**, so their payment (folded into the address view our own order made us subscribe to, since the contract is keyed by script) makes us publish `Paid` for our order off money we never sent. Same net harm — the seller ships twice for one payment. Worse, the test I had just added asserted that residual *as intended behaviour*.
3. **Second fix:** a store's whole book is taken wherever this node holds an order it may settle. Not poisonable by a third party — every order in a store is verified against that store's owner, so only the seller we are already transacting with can add a decoy, and a seller griefing their own buyer merely delays their own sale. The anti-stall property that "own stores only" was protecting survives.
4. **Third round:** the lens noted the property then held by *case analysis living in our heads*, across two copies of a filter. `confusable_with_a_payment` is now the single definition, ANDed into `settled_orders`, so P ⊆ T is true **by construction** and a future fourth conjunct narrows both sides together.

> **Worth recording for whoever reads this next**, in the lens's own framing: the incomplete first fix was a reasonable fix, not a careless one — it closed the direction everyone sees first. The mirror stayed open because *"the two sets are the same set"* is a very convincing-sounding invariant, and it is the wrong one.

### External model (Codex) — 3 findings

- **P2, fixed:** the tip arm of `on_contract_state` never called `publish_settled_orders`, so a fresher tip did not settle the order it unblocked. `assemble_on_chain_proof` measures depth against the tip, so claims can sit in hand for an hour with only a newer tip missing — this made #74 very nearly inert.
- **P2, fixed:** `apply_tip_state` replaced the `TipView` outright, so a delayed GET answer landing after a newer subscribed update could move the tip **backwards** and refuse a payment that is deep enough — #74's own symptom, reintroduced by #74's own mechanism. Now monotonic by height. Residual stated in the doc: a genuine reorg lowers the real tip, and `PaymentReversed` (which has no producer anywhere yet) is what handles a reorg, so this trades an unhandled case for a live one rather than the reverse.
- **P1, filed as #116:** a buyer's twin check sees only stores opened this session. Not fixed, because the obvious gate makes the feature inert — the buyer's equivalent of `unloaded_stores`' enumerable `my_stores` is `remembered_stores`, whose state nothing fetches, so gating on it would withhold *every* buyer settlement forever.

### Testing lens

- Confirmed `one_payment_does_not_settle_two_of_a_buyers_orders_on_one_address` is **not vacuous** by tracing the mechanism rather than trusting the mutation result: the twin's sealed `OrderAccepted` matches `buyer_holding_with`'s shape argument-for-argument, so `buyer_purchases` genuinely picks it up. Also confirmed both new tip tests genuine.
- Confirmed removing `a_buyers_tab_does_not_try_to_publish` left nothing unpinned.
- **Should-fix, done:** no buyer-side coverage of `settlement_publish_failed`. The mechanism does not branch on ownership, but a buyer's rebuilt key makes a permanently-failing send reachable in a way it was not before, and an apology per retry would rebuild the wall harvest#73 removed.

### Mutation results — 12 mutations, and the two that *survived* were the most useful

Every new guard was reverted and the test confirmed red, then restored. Two survivals changed the code:

- A `retired_contracts` filter in `tip_contracts_to_reread` **could never fire** — `register_tip_contract_with_id` already removes a replaced generation from `tip_contract_network` before retiring it, so the two sets are disjoint by construction. Dead filter deleted; the test now asserts the map deletion, which *is* red when removed.
- The owned-store branch of `orders_we_may_settle` was **unpinned**, because the ordinary seller fixtures are buyer *and* seller of one order, so `our_orders` covered them through the purchase path and the branch did nothing observable. Added `a_seller_settles_their_own_order_whose_fingerprint_has_since_moved`, now red.

Both are the ornamental-guard shape: a test passing because its subject cannot fail.

### Dismissed with reasons

- **Hand-built cross-owner delta** (consider, pre-existing): `StoreStateV1`'s doc argues the cross-owner transient is bounded because such a delta carries everything the winner holds — true of `delta()`, not of a single-part delta from `orders_delta_bytes`. Two sentences added to that doc on the lens's advice that this beats an issue nobody picks up. No code change; contract build re-run afterwards to confirm all five hashes still byte-identical.
- **Per-store decryption sweep** (should-fix): real, and the early-out does **not** bound it for a live store — I corrected a comment of mine that claimed otherwise, since a comment asserting a bound it does not deliver is worse than none. Tracked in #116.
- All three residuals folded into **one** issue on the lens's argument that filing them separately invites three partial fixes: they share one answer, keying the twin check off the address contract rather than the store.

### Verification

651 tests (from 639). `fmt`, `clippy --workspace --all-targets`, `wasm32` all clean. Contract hashes verified **five times** — at base, and after each of the four commits including the one touching `common/` — all five artifacts byte-identical throughout, and CI's own `no contract or delegate address moved` job agrees.

[AI-assisted - Claude]

sanity:
### Review status: closed, with one lens substituted-by-self-check (recorded, not hidden)

The **skeptical** lens ran but did not report on its remaining scope after two nudges and roughly an hour and three quarters. Rather than merge on silence or claim a review that did not happen, I checked its three remaining questions directly against the code and wrote the answers into the source:

1. **`withheld_settlements` on a buyer's tab** — cannot grow without bound: `publish_settled_orders` rebuilds its store's entries on every call, and migration prunes them. Not confirmable by the buyer, since "Confirm paid" is on the seller's order card, so an ambiguous buyer settlement waits for the seller — which is what happened before #75 for *every* buyer settlement rather than just the ambiguous ones, so it narrows the old dead end rather than adding a new one.
2. **The `unloaded_stores()` false hold** for someone who both buys and sells — real, and self-clearing: `republish_withheld` is called on a store-list answer and on store-state arrival, so it is a delay rather than a block.
3. **The `our_order_addresses` → `our_orders` refactor** — behaviour-equivalent. Same iteration order, same dedup; the `seen` set moving from a borrowed `&order.id` to an owned clone is identical in semantics, and the short-circuit ordering around it is preserved.

All three are now documented on `publish_settled_orders` so a reader meets them at the code.

Much of that lens's intended ground was covered anyway: the **authorization** lens is also adversarial, ran on Opus, and is the one that found the blocking bug and then rejected my first two fixes for it. The **perf** question the skeptical lens was asked about was answered independently by the authorization lens and is tracked in #116 — including correcting a comment of mine that claimed a bound the early-out does not actually deliver.

Recorded per the standing rule that an unavailable or non-reporting reviewer is substituted and **said so**, never silently downgraded.

Head is now `c2a7f59`. The only change since the consolidated review above is doc-only.

[AI-assisted - Claude]

Issue #70, opened 2026-09-17
a listing has no quantity, cannot be marked sold, and cannot be taken down


sanity:
Ian asked whether a listing carries a count of how many are available, and whether the seller has any UI for it. Neither exists, and the third part is the one that bites.

**No quantity on the listing.** `Listing` (`common/src/listing.rs:139`) is id, title, description, kind, price, created_at. Nothing records stock, nothing decrements, and there is no sold-out state.

**Quantity is a buyer-side field only.** A purchase request carries `quantity: u32` (`ui/src/messaging.rs:154`), and its own comment explains why nothing multiplies by it: a listing's price is free-text `PriceInfo` in an arbitrary currency, so no honest conversion to satoshis exists in that crate. The seller names the amount when they accept.

**A listing cannot be removed.** `ListingsV1` is grow-only with no removal path (`common/src/store.rs:215-227`), and deliberately so: `ui/src/migrate.rs`'s fold-all policy resurrects anything deleted by mere absence, and its soundness argument is precisely that absence is never a deletion here.

So a seller with one of something can neither show a count, nor mark it sold, nor take the listing down, nor (harvest#69) see their own listings at all. The only signal reaching a buyer is whatever the seller writes in the description.

**What makes this more than a missing field:** because listings are grow-only and the fold resurrects by absence, "sold out" cannot be a deletion. It has to be a published state change that survives folding an older generation, which means either

- a superseding signed record under the same listing id, with a rule that picks the newer one (today `apply_delta` only pushes, and first-writer-wins on id would keep the ORIGINAL, so this needs a real ordering rule, e.g. highest version wins like `StoreInfoV1`); or
- an explicit tombstone carried alongside, the way the Bitcoin work handles retractions.

Either is a store-contract change and a re-key, so it wants deciding rather than drifting. A quantity field on the listing is the easy half; what "remaining" means when two peers merge is the part worth designing.

Found while reviewing screenshots of the end-to-end test.

[AI-assisted - Claude]

PR #122, opened 2026-09-22
fix(listing): never publish a listing without its Ghost Key certificate (#118)

### LINKED ISSUE #118: A listing added without a cached Ghost Key certificate is published with none, so buyers see it as "not this seller's" and cannot buy it
## What happens

A listing a seller adds in any session where Harvest has not already fetched their Ghost Key certificate is published with **no certificate**. Every buyer then sees it flagged "This listing's ghostkey certificate is not this seller's." and gets **no Buy button** for it.

That is the ordinary case: open Harvest, go to My Store, click Add Listing. The only sessions that escape are ones where the seller also created the store or submitted Edit details first, because those are the only paths that ask the vault for the certificate.

## Evidence (live, published app `6FzSeAUKcqJrveKyU8RJgGKc5jRB1Z2juvxXtwTA4Em9`, main `5b837b2`)

Two throwaway peers, throwaway Ghost Key, store code `7SvqFBjw5vGE5JYG`:

1. Seller opens Harvest, connects the Ghost Key (store recovered from it), saves a payment key, adds listing "AUDIT TEST". The buyer's view on the second peer shows it flagged and with no Buy control.
2. Hypothesis test. In a new session the seller first submits Edit details (which requests the certificate), then adds "AUDIT TEST 2". The buyer's view shows AUDIT TEST 2 clean with "Buy this". AUDIT TEST is still flagged.

Same seller, same Ghost Key, same store, same app build. The only difference is whether the certificate was in memory when the listing was signed.

## Cause

`on_signature`, `ui/src/state.rs:7389-7397`:

```rust
let certificate_pem = self
    .certificates
    .get(&pending.fingerprint)
    .cloned()
    .unwrap_or(certificate_pem);
```

Since #98 the store key signs listings, and `certificate_pem` is empty for a store-key signature (the doc comment on `on_signature` says so). `self.certificates` is an in-memory map filled only when a `Certificate`/`GhostKeyDetail` response arrives (`state.rs:7776-7812`). `request_certificate` is called from just one place, the store-edit path (`state.rs:6273`), plus store creation (`my_store.rs:904`). A listing therefore falls through to `""`. `unverified_listings` (`state.rs:976`) returns `CertificateStatus::Absent` for it, and `ListingCard` shows the mismatch warning and withholds `BuyControl` (`store_view.rs:276-290`, `:446-450`).

The store edit path already has the right shape. `start_store_edit_if_ready` refuses to sign until a non-empty certificate is cached, and asks for one if it is missing. The listing path needs the same gate: park the listing until the certificate arrives, and never publish one with an empty certificate.

Two smaller things in the same area:

- `Absent` is shown with the copy for a certificate that belongs to someone else. "Not this seller's" is the wrong accusation when nothing was attached. It reads as fraud to a buyer.
- Listings already published this way stay unbuyable. Listing management (#70) can't take them down yet, so the fix needs a way to re-sign them, or at least a note in the seller's view.

## Why this is a launch blocker

This is the core seller action on every visit after the first. The seller gets no signal that anything is wrong, because My Store does not show listings (#69), and the buyer sees what looks like a fraud warning.

Found by the launch audit's live walkthrough.

[AI-assisted - Claude]

sanity:
Fixes #118.

## What was wrong

Since #98 the store key signs listings, and its answer carries no certificate. `on_signature` filled the listing's certificate from `AppState::certificates`, an in-memory cache that only store creation and the store-edit form ever populated, and fell back to an empty string otherwise. A listing added in the ordinary session (the seller did not create the store or submit Edit details first) therefore went out with no certificate. Every buyer saw "This listing's ghostkey certificate is not this seller's." and got no Buy button. The launch audit reproduced it live (see #118).

## The fix

- `PendingListing` now carries `certificate_pem`, attached before the listing is sent for signing, so the signature path no longer depends on the cache.
- `queue_listing_signature` (`ui/src/backing_flow.rs`): with the certificate cached, sign at once. Without it, park the listing in `listings_awaiting_certificate` and ask the vault (`request_certificate`, one request per Ghost Key). The `Certificate` / `GhostKeyDetail` answer releases it.
- Every path where the certificate does not arrive drops the parked listing with a notification naming it ("... was not published ... Add it again to retry"): the vault refusal arms (`drop_vault_signatures`), a request that never left the tab (`on_certificate_request_failed`), a named refusal taken as a watch request's (`watch_signature_failed`), and an empty certificate in the answer.
- A parked listing counts in `user_signature_under_way`, the same way a parked store edit does, so a vault refusal is attributed to it rather than to a background watch request.
- `on_signature` refuses to publish a signed listing whose certificate is empty. This is a backstop.

## Sibling check

I checked every signed record the UI publishes for the same shape, and only listings had it:
- store details: `start_store_edit_if_ready` already waits for a non-empty certificate
- store creation and its backing statement: `start_store_creation_if_ready` already waits on it
- orders and invoices: carry no certificate
- watch requests: signed by the vault, whose `SignResult` brings its own
- reputation `owner_certificate_pem`: set at creation, from the certificate creation waited for

## Tests (all mutation-checked red)

- `a_listing_added_without_a_cached_certificate_waits_for_it_and_carries_it`: the #118 regression. Fresh session, add listing, certificate arrives, the published listing carries it. Red with the pre-fix behaviour restored, and red with the parked listing no longer counted in `user_signature_under_way`.
- `a_listing_with_a_cached_certificate_is_signed_at_once`
- `every_listing_waiting_on_a_certificate_is_released_by_it_alone`
- `a_listing_whose_certificate_cannot_be_had_is_dropped_visibly`: red with the drop removed from `drop_vault_signatures`, then from `on_certificate_request_failed`.
- `a_named_watch_refusal_drops_a_listing_waiting_on_that_keys_certificate`: red with the drop removed from `watch_signature_failed`.
- `an_empty_certificate_drops_the_listing_rather_than_publishing_it`: red with an empty certificate accepted on release.
- `a_signed_listing_with_no_certificate_is_not_published`: red with the backstop removed.

Locally at ce28804: `cargo test -p harvest-ui` passes (658), and clippy is clean both natively and for wasm32. `docs/untested-invariants.md` has a new #118 section.

## Not covered here

- Listings already published without a certificate stay flagged. The store's listings are grow-only. Adding the listing again gives a buyable copy under a new id, and the flagged original stays until #70 can mark it unavailable. Harvest is pre-launch, so only test listings are affected.
- #118 also points out that a buyer shown an Absent certificate gets the copy meant for someone else's certificate. That is a buyer-side wording change in `store_view.rs`, where the seller lane is working. With this fix, no new listing can be Absent.
- A listing waiting on a vault that never answers `GetCertificate` has no timeout. Parked store edits have the same limitation. A reload clears it.

UI-only. No contract, delegate or common code changes, so contract hashes should not move. CI's code-hash check verifies that.

[AI-assisted - Claude]


sanity:
**Review tier: Full.** This touches signing and certificates. Plan: four blind Claude lenses (code-first, testing, skeptical, and certificate/vault-attribution failure mode) plus `codex review --base origin/main`. I'll post each round's consolidated result here.

[AI-assisted - Claude]

sanity:
## Review round 1 (Full tier) on ce28804: consolidated

**Lenses that ran:** code-first (Opus), testing (Sonnet), skeptical (Opus), and end-to-end buyability as the failure-mode lens for this change (Opus). The external pass was `codex review --base origin/main`, which completed. **None of the five found anything blocking.** The buyability lens traced the path from Add Listing to the buyer's `verify_record_certificate`. It confirmed the certificate is the connected backer's, which is in the store's backers, and that nothing else in the buyer path withholds Buy.

Fixed in c3e27e7:

| Finding | From | Resolution |
|---|---|---|
| Deferred custody is not restarted when the certificate wait ends | codex P2, code-first, skeptical, buyability | `start_custody_where_needed()` now runs after the `Certificate`/`GhostKeyDetail` arms, after `on_certificate_request_failed`, and on expiry. Test `a_certificate_arriving_starts_custody_deferred_behind_a_listing`. |
| A parked listing has no timeout. This is now the common path, so a lost `GetCertificate` would hold the listing and the vault silently until a reload | skeptical (should-fix) | `LISTING_CERTIFICATE_TIMEOUT_MS` (2 min), with a re-checking timer like custody's, gives the listing up with a notification. Test `a_listing_whose_certificate_never_comes_is_given_up_in_time`. |
| Take the certificate from the store's own backing instead of depending on the vault | skeptical (should-fix) | `certificate_for` tries the cached vault certificate first, then the certificate in the connecting Ghost Key's own backing of the loaded store (looked up by that key's verifying key). The vault is asked only if both are missing. Test `a_listing_takes_its_certificate_from_the_ghost_keys_own_backing` also checks that another connected key does not get the backer's certificate. |
| The one-request-per-key dedupe blocks a retry and cannot be tested | buyability, skeptical, testing | Dedupe removed. Every parked listing asks the vault. Duplicates are harmless, since any answer releases them all. |
| Drop notifications blame the certificate for any vault refusal | code-first, testing | Each path now says what actually happened: the vault refused a request while waiting, the send failed, the vault's own refusal text, no certificate returned, or a timeout. |
| Test gaps: the `GhostKeyDetail` release, the release-time error branch, and the no-store-key return were untested | testing | Added `a_ghost_key_detail_releases_a_waiting_listing`, `a_listing_whose_store_went_away_while_it_waited_is_reported`, and `a_listing_for_a_store_without_a_key_is_refused_without_waiting`. |
| Stale doc comments: `publish_store_details`' doc had attached to `on_certificate_request_failed`, and the `user_signature_under_way` wording was off | code-first, buyability | Both fixed. |
| The log said "Sent listing for signing" when the listing was only parked | code-first | Now reads "Queued listing for signing". |

Every new guard was mutation-checked red: the backing lookup removed, and any backing matched; the `GhostKeyDetail` release removed; expiry disabled, and off by one; the custody restart removed. The round-0 mutations were re-run against the new code and are all still red. Locally at c3e27e7, 664 ui tests pass, clippy is clean natively and for wasm32, and fmt is clean. `docs/untested-invariants.md` is updated.

Not changed, with reasons:

- **A named refusal taken as a watch request's strands a parked store edit or creation for the same key** (code-first, should-fix). This predates the PR. The PR does not add a new strand of this kind: `watch_signature_failed` drops parked listings. Fixing edits and creations changes the refusal-attribution design reviewed in #99/#101, so it is filed as #127.
- **A parked listing makes a late refusal of a timed-out watch prompt count as the seller's** (skeptical). That is the documented rule for any seller vault work (`named_refusal_is_a_watch_request`: "handled as the seller's"). A parked store edit already behaves this way. With the backing-certificate path and the timeout, the window is now rare and short.
- **`GetCertificate` ignores the one-prompt-at-a-time gate** (skeptical, consider). `GetCertificate` needs no prompt once the key is connected: `RequestAnyAccess` grants `ReadPublic`. `publish_store_details` behaves the same way. The backing-certificate path removes the request in the ordinary case.
- **Carried listings in a legacy move: "Add it again" wording, and the path is untested** (code-first, testing). Creation waits for the certificate and caches it before this loop runs, so these listings sign at once and never reach a drop path. The path is wasm-only, and the invariants doc records it as not driven.
- **The whitespace-trim difference from `start_store_edit_if_ready`** (nit). Stricter here on purpose. Changing the edit path is out of scope.
- **Absent-certificate copy on the buyer side, and listings already published without a certificate**: see the PR body.

CI at ce28804 was green (CI and contract drift). CI for c3e27e7 is running. Round 2 is a lighter re-check of the delta.

[AI-assisted - Claude]

sanity:
## Review rounds 2 and 3: consolidated

**Round 2, on c3e27e7:** a skeptical lens (Opus) re-checked the round-1 delta and `codex review --base origin/main` ran again. Nothing blocking.

| Finding | From | Resolution |
|---|---|---|
| `spawn_backing_statement_signature`'s doc comment had ended up on the new `listing_not_published` helper | skeptical (should-fix) | Moved back onto its function (335c7e6). |
| Restart custody after a named refusal is taken as a watch request's | codex P2 | Added in 335c7e6, then **reverted in 19937e1** because of round 3's finding below. Justification is under round 3. |
| "Duplicate requests are harmless" is only true for successful answers. A duplicate refusal can be attributed to whatever was asked next | skeptical (consider) | Comment narrowed to say so. With the backing-certificate path the vault is rarely asked at all, and the misattribution is bounded to one. Kept the removal of dedupe, because it makes re-adding a listing a real retry. |
| "Add it again" is wrong when the store has no store key | skeptical (consider) | That path's notification now omits it. |
| One failed send drops every parked listing on that key | skeptical (consider) | Left as is. The seller is told about each one, and a failed send means the vault is unreachable, so the other requests are unlikely to have made it either. |
| Check the certificate against the store's backers before signing | skeptical (consider) | Not added. The lens itself traced both sources and found no case where the attached certificate would be rejected: the contract pins each backing to its backer's own slot, and the backing is signed over its certificate. The check needs Freenet's master key verification, and tests cannot mint that. |

The lens confirmed several things are safe: the backing lookup matches the buyer's `backers` (retired backers included), the timer arithmetic and per-listing expiry are correct, it is borrow-safe under `APP_STATE`, and the custody restart after a certificate *arrives* cannot re-raise a prompt the seller refused.

**Round 3, on 335c7e6:** a skeptical lens (Sonnet) looked only at the new custody restart in `watch_signature_failed`. It found that a named refusal can mean the key's `Sign` grant is gone, and restarting custody there would send a custody `SignMessage` for the **same key** in the same turn. That can put a second vault prompt in front of the seller right after they declined one. I agree, so 19937e1 reverts that one call and adds a comment. **Codex's round-2 P2 is therefore declined:** after a refusal, the next store update restarts custody, which is how it worked before this PR. Custody restarts on success, on send failure and on timeout are all kept.

CI was green (all 5 checks, including "no contract or delegate address moved") at 335c7e6. 19937e1 only removes that one line from 335c7e6 and adds a comment. It is being re-verified by CI before merge. Locally at 19937e1: 664 tests pass, clippy and fmt are clean.

[AI-assisted - Claude]

PR #120, opened 2026-09-22
feat(purchase): seller cancel and reader-side order stages (#53 Phase A)

sanity:
## Problem

harvest#53: a purchase stopped at Paid. Nothing produced `Cancelled` outside tests, and once an order was paid neither party's screen said anything about what happens next. A buyer's own purchase card for a paid order read "This order is no longer awaiting payment (Paid), so there is nothing to pay. ... Look again in a moment."

This is Phase A of the design settled for #53: cancel plus reader-side lapse display, with no re-key and no delegate change.

## Approach

**Seller cancel.** The seller's invoice panel gets a two-step "Cancel invoice" for an unpaid invoice. The store key signs `(order id, Cancelled)` through the existing `SignStoreUpdate` (the delegate already classifies that message as `StoreKeyMessage::OrderStatus`, so the delegate is unchanged). The record is verified against the store key before it is published, because the contract refuses a bad record in silence. `Paid` still outranks `Cancelled`, so a buyer who pays anyway settles the order and the seller owes the goods. After payment there is no cancel.

**A payment on a cancelled invoice still settles it**, in the app as well as the contract: settling, address re-reads and bridge watch requests treat a cancelled order as open until its payment window closes, a cancelled order counts as a twin, and the cancel is refused while a payment is visible at the address.

**Buyer cancel is NOT in this PR, and moves to Phase B.** A buyer has no key the store contract recognises until the order carries `buyer_receipt_key`, which is Phase B's store re-key. Adding a buyer arm to the contract's `Cancelled` verification is itself a store re-key. Until then a buyer who changes their mind simply does not pay, and the order lapses by reading, which writes nothing.

**Reader-side stages** (`ui/src/fulfilment.rs`, new). Every window is judged by the reader in blocks against their own tip; no contract reads a clock and nothing is written when a window passes:
- unpaid: lapsed once its payment window (`anchor + PAYMENT_WINDOW_BLOCKS`) closes;
- paid: despatch due within `DESPATCH_WINDOW_BLOCKS` (1008) of the block at which in-window confirmed value first covered the order, read from the Paid record's own evidence through the verifier's fold, so every reader agrees;
- then overdue, with a complaint window of a further `COMPLAINT_WINDOW_BLOCKS` (2016);
- then complete, by silence.

The constants live in the UI crate so tuning them can never move a contract address.

[...]

PR #125, opened 2026-09-22
feat: seller-first My store with listing availability (#69, #70, #93 phase 2)

### LINKED ISSUE #69: a seller cannot see their own listings: My Store shows the store but never lists what is in it
Ian spotted this from a screenshot: the step captioned "a listing" showed no listing, because there is nowhere in My Store that a listing appears.

`ui/src/components/my_store.rs` renders the seller's identities, the share link, the store name with Edit details, the payments panel and an Add Listing button. It never renders the store's listings. The only feedback that a listing was created at all is that "Issue an invoice" becomes available, and the only way a seller can see what they have listed is to open their own store link in Browse, which shows the buyer's view.

Consequences beyond the surprise:

- A seller cannot tell what is currently listed, or that a listing published at all, without leaving the page they manage the store from.
- There is no way to edit or remove a listing (the store contract's listings are grow-only, so removal is its own question, but editing a price or a description is not).
- A duplicate listing is easy to create by accident, and `ListingsV1::apply_delta` is first-writer-wins, so the seller cannot tell which one the network kept.

Suggested shape: list the store's listings under My Store the way the buyer's view does, with the invoice action per listing rather than a single "Issue an invoice" that opens a picker. Worth deciding at the same time whether a listing can be edited, since the answer changes what the row needs.

Found while taking screenshots of the end-to-end test.

[AI-assisted - Claude]

### LINKED ISSUE #70: a listing has no quantity, cannot be marked sold, and cannot be taken down

sanity:
## Problem

A seller could not manage their store:

- **#69**: My Store showed the store but never listed what is in it.
- **#70**: a listing had no quantity, could not be marked sold, and could not be taken down. Listing ids are hashes of their terms and `ListingsV1` is grow-only, so nothing about a published listing could change at all.
- **#93 phase 2**: navigation was still Browse / My Store / Reputation / Payments, with My Store organised around the identity rather than the store.

## Approach

**Contract (329ee7d, re-key).** What changes about a listing now lives beside it. A new store part, `listing_statuses`, holds one `ListingStatus { listing, revision, availability }` per listing id, where availability is `Available { quantity }`, `SoldOut` or `Withdrawn`. The store key signs it, and the merge keeps the highest revision and then the smaller encoding. That is `SignedRecord::rank`, which every other signed record leaves at zero, so their merge is unchanged. A listing's terms and id are untouched. Taking a listing down is a higher-revision status, not a deletion, so the `FoldAll` migration policy stays sound. Serde default plus skip on state, summary and delta means a state with no statuses encodes exactly as before, and every V19 record verifies against the new contract. `classify_store_key_message` recognises the new record, so the delegate signs it.

Re-key, measured: store `b5eddce7 -> 874d4e0c`, reputation `57af8e42 -> 870d459a`, index `0df754b5 -> 2889c25a`, delegate `09507e3f -> d5162880`. Mailbox is unchanged. The superseded hashes are recorded in `legacy/`. This is the index's first row, so it leaves `MAY_BE_EMPTY`. They are pinned in the git-history test.

**UI (e0dcadf).** The navigation is now Stores / My purchases / My store. My store shows one store at a time, titled by its name, with a switcher, and has four pages:
- **Overview**: needs-you items, a set-up checklist, the code and link to share, and the record.
- **Listings**: add, edit, one sold, mark sold out, back on sale, take down, put back.
- **Orders**: the inbox and the invoices.
- **Settings**: details, payout wallet, backing Ghost Key, another store.

The Reputation tab is gone, and each store's record now sits on its own page. The Payments diagnostics are reachable from the footer. A buyer's store page hides taken-down listings, marks sold-out ones and offers no Buy on them. A seller viewing their own store gets a banner pointing to My store.

No listing change is shown before the network has it. While the store key signs, the row says "Saving". After that it shows the store's state, so there is no local marker to strand (the #107 class).

## Not in this PR

- Structured prices or fiat display (#79).
- Per-store payout wallet (#93 phase 1e).
- #53's order/fulfilment work. That lane re-keys the store contract too, so this PR will be rebased onto it, with corpora regenerated and a re-review.

## Testing

- Common: 9 new tests (supersede in either order, equal-revision tie-break, seeded merge laws over clashing revisions, refusal of a non-store-key status with the state left unchanged, wrong slot, signed-content, encoding unchanged with no statuses, classification). `rank -> 0` and `verify -> Ok` were each mutated and went red.
- UI: `listing_status_flow` (revisions above held and pending, count-only edit versus term edit, publishing a signed status, state ingest), labels, one sold, the quantity parser, and My purchases rows. The pending-queue term, always-withdraw and the ingest were each mutated and went red.
- Merge-law generator: new `status` family (`store-status`, `store-status-bad`), added to `run.sh`. The sweep results will be posted here.

Review tier: **Full** (contract/state-shape change, re-key, delegate).

Closes #69
Closes #70

[AI-assisted - Claude]


sanity:
**Review tier: Full.** This changes the store contract's state shape and re-keys the store, reputation and index contracts and the delegate. The plan is 4-5 blind Claude lenses (code-first, testing, skeptical, big-picture, merge-law/migration) plus `codex review --base origin/main`. Each round's consolidated result will be posted here. Held for Ian's screenshot review before merge.

[AI-assisted - Claude]

sanity:
## Review round 1 (Full tier), consolidated

**Reviewed:** `41869ec`, the pre-rebase head with three commits.
**Lenses** (all blind, each in its own read-only worktree):
- code-first (Opus)
- skeptical (Opus)
- merge-law/migration, this change's failure mode (Opus)
- testing (Sonnet)
- big-picture (Sonnet)
- external: `codex review --base origin/main`

**Blocking findings: none.** The merge-law lens checked the contract claims and found them sound: `keeps_held` is a total order, rank 0 leaves every other record's merge byte-identical, there is no stale-record ping-pong, signature domain separation holds, the serde default/skip keeps V19 states canonical, and the registry hashes were recomputed with b3sum from `origin/main`, including the delegate key.

**Fixed in `9e6b9f8`** (hashes after the rebase onto #122):

| Finding | Raised by | Fix |
|---|---|---|
| Editing a sold-out listing put it back on sale | all 5 lenses and codex | An edit keeps sold out unless the seller gives a count (`availability_after_edit`); the form explains this. The replacement listing carries the old one's availability. |
| A term edit could withdraw the old listing even if the new one never published | code-first, skeptical, codex | The old listing is withdrawn only from the new listing's publish-success path (`on_listing_published`). A failure leaves the original up. |
| "One sold" could lose a decrement, and a revision could be reused while the publish was in flight | code-first, skeptical, codex | A new `SentStatus` holds the row at "Saving" until the state echoes it, the publish fails, or 60 s pass. The revision floor includes what was sent. |
| A `u64::MAX` revision froze the listing | code-first, merge-law | Refused, not tied. |
| A sold-out or taken-down listing could still be invoiced through Accept; the badge counted such requests | skeptical, code-first | `issue_invoice` refuses them. They are excluded from the count, and the inbox explains why instead of offering Accept. |
| The details form carried over when switching stores, so one store's details could be published onto another | skeptical | The store body is now a keyed one-item list, so it remounts. A key on a lone node is not diffed. |
| No way to switch Ghost Key before the first store | codex | "Use a different Ghost Key" link |
| Settings lost an in-flight creation or second-store question when the tab was left | skeptical | Kept open |
| "Nothing needs you" never came back once any conversation existed | code-first, skeptical | Conversations no longer count as a need |
| Mailbox decrypted twice per store per render | code-first, skeptical | Once. It still happens on every render, like the inbox itself. |
| A double click published an empty listing; the count-only check was sensitive to formatting | skeptical, code-first | Guard added. `same_terms` compares every term the way the form builds them. |
| Untested filters: visible listings, Buy gate, invoice picker, `seller_stores`, same-terms, per-conversation keys | testing | Extracted where needed and tested |
| Stale docs: index-lineage comment, statuses growth claim, `store-status-adv` name, misplaced doc comments | merge-law, code-first | Fixed. Corpus gains a statuses+backing+retirement state. |

Every new guard was mutation-checked red. One mutation first survived: dropping the `kind` comparison in `same_terms`. A case was added and it now goes red.

**Addressed without a code change:**
- **"Saving" can stick if the delegate never answers a signature.** True of every store-key request today, not just this one. Documented in `docs/untested-invariants.md`, and it will be tracked with that class instead of fixed only here.
- **Restoring the count on "Back on sale".** A sold-out status has no count to restore. The seller sets one through Edit.
- **Capping a buyer's requested quantity at the count.** The seller names the amount when invoicing, so the count stays advisory.
- **`describe_lost_store` should mention statuses.** Consider-level. A lost status only affects availability, which re-merges from the network copy.
- **"Change Ghost Key" in Settings.** Not built anywhere yet; it belongs to #93's remaining work and is not claimed here.
- **Delegate re-key rehearsal.** Done on a live throwaway peer: the shared test Ghost Key's V19 store migrated to the new store address with its listings, and custody recovered its store key through the new delegate ("Recovered your store's key from your Ghost Key").

**Rebase onto `origin/main`** (#122 cert fix, `7a22772`):
- Two conflicts.
- `docs/untested-invariants.md`: both new sections kept.
- `my_store.rs`: `sign_and_submit_listing`. #122 changed only its log line; this PR deletes the function. New listings go through `queue_listing_signature`, which is #122's certificate path.
- Fixtures now hold a certificate (`08530e0`).
- Contract hashes are unchanged by the rebase: `874d4e0c` / `870d459a` / `2889c25a` / `d5162880` against main's `b5eddce7` / `57af8e42` / `0df754b5` / `09507e3f`.

A re-review of the fix commit and the rebase follows.

[AI-assisted - Claude]

sanity:
## Review rounds 2 to 5, rebases, and the screenshot run: consolidated

Round 1 is in the comment above. Everything since has been re-reviewed as it landed. Each round got blind lenses in a read-only worktree at the stated SHA.

| Round | SHA | Lenses | Blocking | Result |
|---|---|---|---|---|
| 2 | `08530e0` (fixes + rebase on #122) | skeptical (Opus), code-first + rebase resolution (Opus) | **1**: the inbox accept controls lost their list key when wrapped in if/else (dioxus reads a key only from a loop body's first node), so accepting one request showed the next as accepted | Fixed in `d1da31d`. Also fixed there: the invoice guard was blocking legitimate flows (answering a request for an edited or sold-out listing, reissuing an expired invoice); "0 left" edits; an edit in flight allowed a second edit to publish a second replacement; #118 drop paths leaked the pending withdrawal; "One sold" double click; a failed publish unlocking a newer send; unstable store order; the listings loop key. |
| 3 | `bdeb989` | skeptical (Opus) | 0 | Two drop paths still leaked; the Orders count included requests the seller could never clear; the store shown could switch under the seller. Fixed in `2630cc2`. |
| 4 | `2e443c7` | skeptical (Sonnet) | 0 | A dropped edit lost its reason (e.g. "no store key on this device"). Fixed in `06c3689`. |
| Rebase | `e50c2e0` (onto #120 #53 Phase A, #126, #128) | code-first on the resolution (Opus) | 0 | Nothing of main's lost; `range-diff` clean. One medium: "Needs you" nagged about a lapsed invoice until it was cancelled publicly. Fixed in `1d6dba4`. |
| 5 | `9bf7abe` and `4293781` (found by screenshots) | skeptical (Opus, Sonnet) | 0 | My purchases' background loading: the indicator could stick; the load ran before the remembered stores were known; a background arrival claimed the open store; a failed send could remove another flow's data. Fixed in `d46f0b7` and `7f89e66`. |

**Invoice policy, as it now stands:**
- A fresh invoice from the picker is refused only for a taken-down listing.
- Sold-out listings stay in the picker, marked, so an expired invoice can be issued again.
- A buyer's request can always be answered, with a note if its listing is no longer on sale.
- The Orders count and badge include only requests for a listing still on sale in the store.

**Found by the screenshot run and fixed** (the UI running on throwaway peers):
- After a reload, My purchases was empty for a buyer with an open order: a store is listed only once its state has loaded, and nothing loaded a remembered store until the buyer opened it. Fixed in `9bf7abe`.
- The effect that loads remembered stores re-ran for ever, because it wrote to the app state it reads, which froze the page. Fixed in `4293781`.

**CI:**
- The committed `store_contract.wasm` was stale after the round-1 doc edits in `common/src/store.rs` (panic locations move the bytes). Rebuilt: store is now **`b8a1bd88`**. `874d4e0c` was branch-only, so it gets no registry row. The other artifacts are unchanged.
- The PR body is updated, and the PR is labelled `contract-rekey-acknowledged`.
- Green at `1d6dba4` (CI, contract drift). A run on the current head is in progress.

**External model:** `codex review` failed at 16:38 and 17:02 CT: usage limit ("try again at 4:50 PM"), then the same error again after 4:50. Per the rule, this was retried once and then substituted with the Claude lenses above. Codex's round-1 findings on `41869ec` were all addressed.

**Merge-law sweep:** full corpus regeneration and `tests/merge-laws/run.sh` against the final WASM (store `b8a1bd88`), results to follow. An earlier partial sweep on the pre-doc-edit bytes (`874d4e0c`, 22 of 38 corpora) found only the known standing `store-adv` 6 violations (#81), and the `status` family's native laws were clean (12 states, 1728 triples).

[AI-assisted - Claude]

sanity:
## Merge-law sweep on the final WASM

Ran `tests/merge-laws/run.sh` on the full corpus list, regenerated against the final WASM:

| Contract | BLAKE3 hash |
|---|---|
| store | `b8a1bd88...` |
| reputation | `870d459a...` |
| mailbox | `fad8339c...` |
| index | `2889c25a...` |

These are the bytes committed at the current head `7f89e66` (no contract byte changed after `e50c2e0`).

**Result:** 38 corpora swept, none missing.

- **Violations: only the known standing ones.** `store-adv` has 6 and `reputation-adv` has 8, on `state_commutativity` / `reconciliation_cycle`. Both are tracked in #81, with the same counts as the README's baseline.
- **New `store-status`** (statuses racing across revisions, equal-revision clashes, a status arriving before its listing, statuses beside a backing and a retirement, 17 delta steps): 0 violations, 0 inconclusive. State pass 907 cases, bundle pass 1393.
- **`store-status-bad`** (a stranger's status, a status in the wrong slot): 0 violations. It has inconclusive cases, as every invalid-state corpus does: those are the states the contract refuses.
- The generator's native law check on the `status` family was clean: 12 states, 1728 triples.
- Single "no delta path to compare against" inconclusives in `store-empty`, `store-rr` and the reputation and mailbox corpora are structural: a pair with nothing to send.

[AI-assisted - Claude]

sanity:
**Rebased onto main `087c57c` as `27fceec`.** The conflict resolution (22 files: store state shape, Phase C reputation UI, generator, legacy rows) is unreviewed, so the prior rounds do not cover it.

Review tier for this round: **Full** (contract state shape, re-key of store/reputation/index/delegate, delegate signing, complaint UI placement). Lenses: code-first, testing, skeptical, big-picture, and a rebase-resolution / migration-and-threat-model lens, plus `codex review --base origin/main`. Merge-law sweep over freshly generated corpora and a network-mode rehearsal are running; results will be posted here.

One resolution bug was already caught locally before review: the "keep both" resolution of `store.rs` had dropped the `#[serde(default, skip_serializing_if)]` attribute from `listing_statuses` on the summary and delta (the attribute lines belonged to the hunk's context above main's `fulfilment`). `no_statuses_encode_as_before_they_existed` went red on it; fixed before the push, and the WASM rebuilt.

[AI-assisted - Claude]

sanity:
## Review round 1 of the rebase (Full tier): consolidated

Lenses (blind, parallel, read-only): rebase-resolution + migration + threat model (Opus), skeptical (Opus), testing, code-first, big-picture (Sonnet).
**External model:** `codex review --base origin/main` failed: "You've hit your usage limit ... try again at Sep 26th, 2026 5:35 PM". It will be retried once in round 2; the two Opus lenses above are the substitute.

Main moved during the round (#154, UI and docs only, no contract change); rebased cleanly onto `a8c448d`, no conflicts. Fixes are in `5909e8c`.

| # | Lens | Sev | Finding | Outcome |
|---|---|---|---|---|
| 1 | rebase | P1 | `StoreRecord` said "No complaints." for an unread record, undoing #143 P1-5 | **Fixed**: shows the badge text until `RecordLoad::Loaded`; `an_unread_record_is_not_called_clean` (red when the load state is ignored) |
| 2 | rebase | P1 | My purchases showed a kept order twice (store card + kept list), two complaint controls | **Fixed**: `kept_purchases_to_list` leaves out orders a card on the page shows; `my_purchases_lists_a_kept_purchase_once` (red with `shown` ignored) |
| 3 | rebase | P1 | Threat-model / `KeptPurchases` doc wording stale after the move | **Fixed** (3.6, the row in section 2's table, the component doc). R3 itself holds: the address still comes only from `PurchaseCard`, gated on `PurchaseNotKept` |
| 4 | rebase | P2 | V16 derived with today's `reputation_params` in both impl and test helper | **Fixed**: test helper uses a frozen `{store_key}` struct |
| 5 | rebase, code-first | P2 | Mis-attached doc comment in `state.rs`; diagnostics first-run test still reads kept purchases; heading "Payments"; "Payments tab" strings in tests | **Fixed** |
| 6 | skeptical | P2 | A status sent to a store still on a pre-status generation is dropped by serde | **Residual, documented** in `docs/untested-invariants.md`: same exposure as every part added since V1 (Phase B despatch included); migration repoints at adopt (#148/#151) |
| 7 | skeptical | P2 | "Saving" expiry only re-evaluated on an unrelated render | **Fixed**: Listings re-renders every 5 s (wasm) |
| 8 | skeptical | P2 | An edit's carried status can fail while the replacement publishes: sold-out reads on sale | **Mitigated + documented**: both failure paths now say so and tell the seller to set it again (`LISTING_STATUS_NOT_SAVED`); residual row added |
| 9 | skeptical | P2 | A lost "One sold" is overwritten by the next click | **Not a defect**: the row returns to the count the store holds once the window ends, so the seller sees the missed sale before clicking again; a decrement is always from the held count, never from a local guess |
| 10 | skeptical | P2 | Invoice-panel comment says sold-out listings are excluded | **Fixed** |
| 11 | testing | P1 | No test of the overview's record badge wiring or expired-invoice count | **Fixed**: `the_overview_counts_only_orders_that_need_reissuing` (red with `needs_reissue` dropped, and with the badge forced to Loaded) |
| 12 | testing | P2 | `any_kept` empty-state gate untested | Covered indirectly by #2's helper; the gate itself is render-only |
| 13 | big-picture | P1 | Review stale: main moved (#154) | Rebased (clean); round 2 runs on the new head |
| 14 | code-first | P2 | Listing certificate warning copy changed without mention | Intentional (branch, pre-rebase); noted here |

Local at `5909e8c`: `cargo fmt --check`, clippy host + wasm32 UI + address guard, `cargo test --workspace`, rehearsal `cargo check`, `check-code-hashes.sh` all green. Contract hashes unchanged by the fixes (UI only). Merge-law sweep over regenerated corpora is running; network-mode rehearsal next.

[AI-assisted - Claude]

sanity:
## Review round 2 (on `5909e8c`, the round-1 fixes plus the clean rebase onto `a8c448d`)

Lenses: skeptical (Opus), code-first + testing (Sonnet), blind, read-only.
**External model:** `codex review --base origin/main` retried once: same failure ("You've hit your usage limit ... try again at Sep 26th, 2026 5:35 PM"). Substituted by the Opus skeptical lens, per the review rule; not waiting for it.

| # | Lens | Sev | Finding | Outcome |
|---|---|---|---|---|
| 1 | skeptical | P1 | The kept-list dedupe matched on order id alone. A store card built from an acceptance can name an order kept under ANOTHER store's key, and that card offers no complaint, so the only complaint control for a paid order could be hidden by any store in the buyer's conversations | **Fixed** in `1de02a6`: `AppState::kept_purchases_shown_at` yields `(store owner key, order id)`; the kept list matches the pair; a store with no known owner names nothing. `my_purchases_lists_a_kept_purchase_once` now covers the other-store and ownerless cases (red when matched on id alone). A redundant `kept_copy` filter was removed: with the pair match it could never change the result |
| 2 | skeptical | P2 | The overview's "needs reissue" count may flag an order a buyer can still pay late (#154 watches through the whole payment window) | **Not changed**: the count is main's existing per-invoice `needs_reissue` marker (invoice list), summed; it only counts orders whose stage is still `AwaitingPayment` with this reader's payment sight, so an order with a seen payment is not counted |
| 3 | skeptical | - | `StoreRecord` empty text, the 5 s re-render loop, the `LISTING_STATUS_NOT_SAVED` early return, the frozen V16 struct, and #154 interaction | Checked, no finding |
| 4 | code-first | P3 | Section 10's R5-B row still said "Payments-tab list" | **Fixed** |
| 5 | code-first | P3 | Round 1 item 12 overstated coverage: the `any_kept` empty-state gate is not exercised by the dedupe test | Agreed, wording corrected here: the gate is render-only and untested; it reads the unfiltered kept list, so "Nothing yet" never shows while any kept purchase exists |

Local at `1de02a6`: fmt, clippy host + wasm32 UI, `cargo test -p harvest-ui` (894) green. UI-only, contract hashes unchanged.

[AI-assisted - Claude]

sanity:
## Round 3 + rehearsal

**Round 3** (skeptical, Opus, on `1de02a6`): no P0/P1. "Never hidden without a working control" holds: a card hides a kept purchase only when its store's owner key equals the kept key, and that card judges the purchase from the same kept copy (`paid_copy` prefers a kept `Paid` copy; `complaint_checks` takes the kept branch). P2s:
- The test used an unpaid purchase, so nothing pinned that the hiding card offers the complaint. **Fixed** in `f44685b`: a paid kept purchase is hidden only while the card reports it paid with the same `complaint_refusal` answer as the kept row (red when `paid_copy` stops preferring the kept `Paid` copy).
- Shown twice when two loaded stores share one owner key, or while a store's owner is unknown. **Accepted**: both copies carry working controls and share the sent marker; the ownerless case is the deliberate "never hide" side. Predates this commit.
- `complaint_sent` keyed by order id alone: predates #125, noted.

**Rehearsal, isolated network-mode gateway** (`freenet network --is-gateway --skip-load-from-network`, freenet 0.2.136, 0 peer connections in the log):
- `tests/rehearsal/delegate-rehearsal.sh` against the build under test (delegate `9463b53b`): every generation V5-V22 answers both predecessor calls; scenario **skipped** (V21 holds the data, V22 never ran): `V21: imported 8`, every secret value for value, **PASS**; scenario **newest** (V22 holds the data): `V22: imported 6`, every secret, **PASS**. The harness's seed had to learn that V22 (Phase C) no longer mints an RSA key (`a870d88`); the RSA import itself is exercised by the V21 scenario.
- Store: new scenario 5 (`REHEARSAL_ONLY=newest`): a store planted at V21 (`4aa47d75`, owned by its key, code-addressed) is found by the walk, sealed, and its state accepted by this build's contract with store name and listing intact, listing signature re-verified. **PASS**.
- Reputation (`REHEARSAL_ONLY=reputation`): 4a, 4b, 4c pass; 4d's fold and PUT-forward pass (146 kept, honest 3/3), then its extra `UpdateData::State` merge of the 150-complaint state is refused by the node. The reputation contract and `common/src/reputation.rs` are identical to main, so this is not from #125: filed as #157. Scenario 1 is the known #142.

The webapp used for the delegate rehearsal was built at `5909e8c`; everything after it is UI dedupe and test code, nothing on the migration path.

[AI-assisted - Claude]

sanity:
## Round 4 (light: the test-only commits `f44685b`, `a870d88`)

Lens: skeptical (Sonnet), read-only.

| Sev | Finding | Outcome |
|---|---|---|
| P1 | The delegate rehearsal's seed dropped its "RSA key is present" assertion for every generation, so a generation that SHOULD mint one could seed none and the check would report two absences as a pass | **Fixed** in `9b09a61`: the seed records whether the generation minted (`rsa_minted`) and asserts a key exactly then. Re-run: delegate rehearsal PASSED (V21 scenario 270-byte key carried; V22 scenario none minted, none held), scenario 5 PASSED |
| P2 | The kept-purchase pin compared two refusals that might both be `None` | **Fixed**: also asserts the card's complaint is open (`None`) and `card.paid` is set, so the equality is about an offered control |
| P2 | Scenario 5 printed the status count without asserting it | **Fixed**: asserts no listing status appears |

CI at `9b09a61877a61ce9172b52333a590bc59d489fe7`: `CI` success (run 35929311204), `contract drift` success (run 35929311339). Merge-law sweep still running (42 corpora generated, generator exited 0; so far only `store-adv`'s 6 violations, identical to main's baseline sweep, the known #81).

[AI-assisted - Claude]

sanity:
## Merge-law sweep: done

Corpora regenerated from this branch (`tests/merge-laws/gen`, generator exited 0, **42 corpora**, including the new `store-status` and `store-status-bad`), then the full `run.sh` list swept against the WASM snapshot `store 307332ee`, `reputation 492c6e74`, `mailbox 4b02a3f4`, `index e23c43c8` (the hashes in the table above). 42 of 42 corpora swept.

- `store-status`: 907 cases, 907 hold, 0 violations; bundle run 1393/1393.
- `store-status-bad`: 0 violations (85 cases, 55 inconclusive, as a refusal corpus should be).
- Every other corpus: 0 violations, except `store-adv`: 6, identical to main's baseline sweep at Phase C (the known #81).

With CI green at `9b09a61` and every finding from rounds 1-4 fixed or answered above, this is ready to merge.

[AI-assisted - Claude]

sanity:
Merged as `f9558cd`. Published from `main` at that commit under the lanes' publish lock: version `1790209518`, the node serves `harvest-ui-dxh2a4ad618a1dbab51.js`, the asset the build produced (checked on the publishing node and again through a fresh network peer). Counter PR #158 merged (`f744832`).

Smoke test on a throwaway network peer (10 connections): the published UI loads, the navigation reads Stores | My purchases | My store, and My purchases opens.

Follow-up, deliberately not in this PR: the "My store" tab placement / navigation rework from the design pass, as a UI-only change. Rehearsal-harness rot found on the way: #142 (scenario 1) and #157 (scenario 4d).

[AI-assisted - Claude]

PR #126, opened 2026-09-22
fix(messaging): wait for the node to hold a contract before writing to it (#119)

### LINKED ISSUE #119: A buyer's first message to a seller is lost when their node lacks the mailbox contract: the update races the priming GET and the node's retry request is ignored
## What happens

A buyer's first purchase request to a store was lost on the first try in the launch audit's live run. The buyer saw "Your request ... has been handed to your Freenet node". The seller's inbox still said "No one has written to this store yet" several minutes later, and nothing ever told the buyer.

It was a fresh peer's first write to that seller's mailbox. That is the normal case for every new buyer, so on a fresh node a first "Buy this" or "Contact Seller" can simply vanish.

## Evidence (live, published app, main `5b837b2`, freenet 0.2.136)

Buyer peer log, one Buy-this request (contract `98X4cGfh…` is the seller's mailbox):

```
19:27:37.831  Created new operation ... resource=Update { key: ... 98X4cGfh7BsNpYBMxaFUMLsiVV1coKFRiCuRpWCwvt2e ... }   (req-57000007)
19:27:37.849  WARN  Contract parameters not found in state_store contract=98X4cGfh...
19:27:37.849  ERROR Failed to update contract value ... reason: missing contract parameters
19:27:37.849  WARN  update: originator has no contract code/params for this contract; triggering auto-fetch from target and asking client to retry
19:27:37.850  Delivering result for transaction 01M359D1V76DTE89X0X6JVVWG4 to 1 clients
19:27:39.584  get (driver targeted sub-op): UPDATE auto-fetch succeeded ... bytes=11      (the empty mailbox)
19:27:39.614  subscribe: subscribed ... 98X4cGfh...
```

There is no second `Update` for that contract in the log. The node asked the client to retry, and Harvest never did.

## Cause

This is the residual that `gateway/mailbox_ops.rs::send_message` (`ui/src/gateway/mailbox_ops.rs:115-161`) and `docs/untested-invariants.md:286-287` already describe: the priming `get_contract` and the `update_contract` both resolve on WebSocket send, so the update can reach the node before the node has the contract. Both documents said it needed a node to characterise. This run characterises it: the race was lost on the first attempt, and the node's answer is an explicit "retry", not a silent drop.

## Possible fixes

- Await the priming GET's actual response (the subscribe path already has a deadline pattern) before sending the update.
- Or, on the node's retry/error reply, re-send the same sealed bytes. The doc comment's objection to retrying is that a re-seal would create a duplicate. Re-sending the identical `EncryptedMessage` dedups by its nonce, so that path is idempotent.
- Either way, the "not yet visible" state (`unconfirmed_sent`) should eventually tell the buyer that the seller has not received the request, rather than leaving it indefinitely.

A freenet-core change could also queue the update behind its own auto-fetch instead of bouncing it. That would help every app, but Harvest cannot rely on it being deployed.

## Why this is a launch blocker

This is step one of every purchase. The buyer is told it worked, the seller never sees it, and the buyer's only recourse is to guess and send again.

[AI-assisted - Claude]

sanity:
## Problem

freenet/harvest#119, a launch blocker. A buyer's first message to a seller's mailbox from a fresh node was lost. The priming `get_contract` and the `update_contract` both resolved on WebSocket send, so the UPDATE reached the node before the node held the mailbox. freenet-core then starts an auto-fetch and answers the client with an error asking it to retry (`operations/update/op_ctx_task.rs`, `phase = "auto_fetch_originator"`). That error has no correlation id and names the contract only in English prose, and Harvest never retried. The buyer was told the request was handed over; the seller never saw it.

## Approach

**1. Every contract write waits for the node to hold the contract (`ui/src/gateway/prime.rs`).** `gateway::update_contract` now registers a waiter, sends a GET (no subscribe), and sends the UPDATE only when the node has answered for that contract (`GetResponse` or `NotFound`, routed from `response_handler`) or a 30s deadline passes. This is correct by construction, not by timing: on the client-GET path freenet-core awaits `cache_contract_locally` before `build_host_response` (checked in v0.2.136 `get/op_ctx_task.rs`), so a `GetResponse` proves the node holds the contract. On timeout or NotFound the update goes out anyway, which is exactly the old behaviour, so this can only help.

**2. The buyer's message is verified and re-sent (`mailbox_ops::deliver`).** 20s after the write, the mailbox is re-read (waiting for the node's answer). If the message is missing, the IDENTICAL sealed bytes are sent once more; the mailbox keys entries by `entry_digest`, so this cannot deliver twice. Never a re-seal: a fresh seal is a fresh nonce and a second message. If it is still missing, `SentMessage::not_arrived` is set and the thread shows "The seller has not received this yet" with **Send again** (same bytes again).

Why both: priming alone leaves the case where the GET never answers (dead end, slow network) and the write is bounced; the check alone would bounce every first write and rely on the retry. Why not act on the node's retry error: it is uncorrelated, and keying behaviour on node log prose couples Harvest to a message string. Observing the mailbox is the delivery signal Harvest already uses (`unconfirmed_sent`).

## Sibling sites ("finish the fix")

All contract writes in the UI go through `gateway::update_contract` / `put_contract` (the raw update is now `delegate_api::send_update`, `pub(super)`, called only from `prime`). So every update site is fixed by construction:

- mailbox: buyer send, seller reply, seller acceptance
- store contract: buyer's Paid settlement (writes to a SELLER's store, exposed on a fresh buyer node), listing, key copy, store details, invoice
- bridge inbox watch request (`state.rs` `InboxEntry`; exposed on a fresh node, previously self-healed only by its own grace-period re-request)

Not exposed: PUTs (store/reputation/mailbox creation, Ghost Key index, migration) carry the contract code, so the node never lacks it. Tip and address contracts are read-only from Harvest. Reputation/feedback writes do not exist yet (#53); when they land they get priming for free by going through `update_contract`. Delegates make no contract writes.

## Testing

- `gateway::prime` tests drive the real ordering function against a fake node whose GET answers when the test says: update withheld until the answer, another contract's answer releases nothing, deadline, NotFound, unsendable GET, waiter pruning.
- `response_handler`: a `NotFound` and a `GetResponse` each release a waiting write.
- `mailbox_ops::deliver`: never lands (identical bytes resent once, then reported), lands on the resend, lands first time (sent once), node unreachable (reported with reason).
- `state`: a resend returns the identical sealed message and clears the warning; a late-landing message stops being reported.
- Mutation-checked, all red: old back-to-back ordering (5 tests), no automatic resend, no report, either handler release removed, resend keeps warning, `mark_not_arrived` no-op, resend with a changed nonce.
- `docs/untested-invariants.md` updated: the old "race is tolerable / untested" rows are replaced.
- Live proof on two throwaway peers: posted as a comment below.

Review tier: **Full** (async ordering, messaging, every contract write path).

Closes #119

[AI-assisted - Claude]


sanity:
**Review tier: Full.** This touches async ordering on every contract write path and the messaging delivery path. Planned lenses: code-first, testing, skeptical (async/ordering), big-picture, plus a lens on this change's failure mode (a write held back or duplicated: waiter lifecycle, response-handler routing, idempotence of the resend), and `codex review --base origin/main`. Each round's consolidated result will be posted here.

Reviewed head: `4390657cb67a4c58ba09f4ac69e81d6376b3c4f2`

[AI-assisted - Claude]

sanity:
## Review round 1 (head `4390657`), consolidated

Lenses: code-first (Sonnet), testing (Sonnet), big-picture (Sonnet), skeptical async/wasm-executor (Opus), this change's failure mode: held or duplicated writes and a wrong "not received" (Opus), plus `codex review --base origin/main` (ran, gpt-5.6-sol). No blocking findings. Verified independently by two lenses: a `GetResponse` does prove the node holds the contract (v0.2.136, `cache_contract_locally` awaited before `build_host_response`); the woken writer runs after the response handler's synchronous pass; no new RefCell double-borrow; no bypass of `update_contract`; no write loops from the pre-write state the priming GET delivers.

Findings and disposition (fixes land in the next commit):

1. **should-fix (skeptical + failure-mode):** the landing check reads `browsing_stores[store]`, but `mailbox_to_store` maps a mailbox (derived from the Ghost Key, not the store) to the FIRST store that registered it. Two stores under one Ghost Key, or a mapping dropped after failed subscribes, means a message that landed never reads as landed, and the buyer is told "not received" with a Send again that cannot help. **Fix:** check landing by mailbox id through `mailbox_to_store`; an unmapped mailbox is "cannot tell", which never produces the warning. `resend` re-registers the mailbox as the first send does.
2. **should-fix (codex P2):** a re-read that timed out or could not be sent was discarded and the stale cached state judged as absence. **Fix:** the re-read outcome is propagated; only a check backed by a fresh node answer can conclude "missing". An unverifiable check may still re-send (identical bytes, harmless) but never tells the buyer "not received".
3. **should-fix (failure-mode, skeptical #4):** when the node could not be reached, the card said "Harvest sent it more than once". **Fix:** the mark records why (never reached the node vs. not in the mailbox) and the card words each.
4. **consider → fixing (failure-mode #3):** a resend addressed the store's CURRENT seller key. **Fix:** the buyer's record keeps the key the message was sealed to and addressed with; resend uses it. Only a buyer's own messages are resendable.
5. **consider → fixing (skeptical #3):** double-click on Send again started two loops. **Fix:** resend is a no-op unless the message is marked.
6. **should-fix (testing):** `take_for_resend` was tested with one store. **Fix:** two-store test.
7. **consider → fixing (code-first):** a GET that could not be sent left its waiter until the next write to that id. **Fix:** `register` prunes abandoned waiters across all contracts.
8. **consider → documenting (code-first, skeptical #5, failure-mode #5):** worst-case time to the warning is roughly 30s prime + 2 × (20s wait + up to 30s re-read), and a write to a contract the node holds without interest may now wait on a network GET. Documented at the constants. The soft "not yet visible" state covers the interval; before this PR these writes had no check at all.
9. **consider → rewording (failure-mode #4):** buy-form copy "checks that it reaches the seller's mailbox" → "checks that it shows up in the seller's mailbox", the standard the check actually applies.
10. **not changed, with reason (codex P2, skeptical #2):** an answer to an OLDER in-flight GET (or a stale NotFound from the subscribe GET) can resolve a waiter. GET responses carry no request id, so per-request correlation is not available from the protocol. The effect is bounded to the safe direction: an early release is caught by the delivery check, a resend of identical bytes cannot duplicate, and a "not received" card clears by itself as soon as any later read shows the message (`unconfirmed_sent` filters on what is in the mailbox). A false warning now needs two consecutive verified checks to both see a stale state. Recorded in `docs/untested-invariants.md`.
11. **not changed (testing):** the thread-local `WAITERS` is touched by one test with fixed ids it fully resolves; a comment warns against reusing them.

Live two-peer proof: in progress on a build signed with a throwaway web-container key (own contract id, not the production one), peers with private webapp caches.

[AI-assisted - Claude]

sanity:
## Review round 2 (head `9d4954b`), consolidated; fixes in `b74f7b4`

Lenses: skeptical re-review of the round-1 fixes (Opus). `codex review` was attempted on a clean checkout of `9d4954b` and failed mid-run with "You've hit your usage limit ... try again at 4:50 PM"; retried once on `b74f7b4`, same error. Substituted: an independent Opus lens standing in for Codex on `b74f7b4` (result in the round-3 comment). Not waiting for the reset.

Verified by the lens: `record_sent_to_seller` attaches to the right record; `mailbox_holds` cannot be fooled by a planted entry (`entry_digest` covers nonce, conversation id, sender key, ciphertext and timestamp); freshness comes only from an answer for that mailbox; the double-click guard works; the `register` prune is correct; the new tests each fail under the mutation they target.

Findings and disposition:

1. **should-fix, fixed:** a socket drop during the wait made the RESEND's write fail, which reported "it never reached your node" although the first copy had. Now only a first-write failure says that; a later one marks the message *Unconfirmed*. Test `a_resend_that_cannot_reach_the_node_claims_nothing_about_the_first_copy` (red with the old rule).
2. **should-fix, partly fixed and recorded:** with two stores under one Ghost Key, the mailbox state is filed under the first store, so the second store's thread and "not yet visible" list never see a landed message. This PR's check no longer produces a false "not received" there (it reads by mailbox), but the older display gap remains. Filed as #130 rather than widened into this PR; the round-1 claim that a false card always clears itself is corrected in the code docs and `docs/untested-invariants.md`.
3. **consider, fixed:** "could not check" (re-reads unanswered, mailbox not routed) ended in a permanent "not yet visible" with no button, and that is #119's own condition. It now marks the message *Unconfirmed*: "Harvest could not check whether this reached the seller's mailbox ... It may have arrived", with the same Send again. Tests updated; red when this case stays silent.
4. **consider, fixed (docs):** one stale answer on the second check is enough for a false card, not two; wording corrected.
5. **consider, fixed:** `resend` re-marks the message if the mailbox key cannot be derived, so the mark is never cleared with nothing sent.
6. **consider, fixed:** `resend` re-registers the mailbox only when the store records none or records this same one, so a re-backed store is not repointed at an old mailbox.

Mutations for this round (all red): later write failure reported as never sent; could-not-check stays silent.

[AI-assisted - Claude]

sanity:
## Live proof on two throwaway peers (freenet 0.2.136), branch build `9068a01`

**Setup.** Two fresh `freenet network` peers on nova, each with its own config/data/log dirs and a private `FREENET_WEBAPP_CACHE_DIR`. Neither is 7509/7510.
- **Seller (peer1)** ran the published production app (served asset `harvest-ui-dxh3fd5e82be7d8eb8f.js`, the #122 publish) with the throwaway test Ghost Key. It recovered the existing test store `7SvqFBjw5vGE5JYG`.
- **Buyer (peer2)** ran this branch built at `9068a01` with the production contract id. Contract hashes were identical to production (store `b5eddce7`, index `0df754b5`, mailbox `fad8339c`, delegate `09507e3f`). The build was placed only in peer2's private webapp cache, so nothing was published to the network. peer2 served `harvest-ui-dxha19aca9e1bc31f20.js`, which matches the build.
- Before the send, peer2's log had no line for the seller's mailbox `98X4cGfh…`. That is the #119 precondition: the node does not hold the contract.

**Result: the first "Buy this" from the fresh buyer node arrived.** Buyer peer log, one request:

```
2026-09-22T21:08:58.098302Z  INFO contract_handling:upsert_contract_state_deferrable{key=98X4cGfh7BsNpYBMxaFUMLsiVV1coKFRiCuRpWCwvt2e}: freenet::contract::executor::runtime::executor_impl: Contract initial state installed contract
2026-09-22T21:08:58.109171Z  INFO freenet::client_events::session_actor: Contract GET response ready for delivery: contract=98X4cGfh… bytes=6738
2026-09-22T21:08:58.116158Z  INFO client_event_handling:process_client_request: freenet::node::request_router: Created new operation - starting network request transaction=01M35F6KM4R10CHD6AX6MK7Y84 resource=Update { key: Contract
2026-09-22T21:08:58.544285Z  INFO freenet::client_events::session_actor: Contract GET response ready for delivery: contract=98X4cGfh… bytes=8972
2026-09-22T21:08:58.795674Z  INFO freenet::operations::subscribe::op_ctx_task: subscribe: subscribed tx=01M35F6M1G42Q93294H7CXWBG3 attempt_tx=01M35F6M1GJJ7FQ391B5KR2H83 contract=98X4cGfh… target=174.65.98.50:44182 retries=0 atte
2026-09-22T21:12:29.104537Z  INFO freenet::operations::subscribe::op_ctx_task: subscribe: subscribed tx=01M35FCZGEZPPFDQR5DK48VV03 attempt_tx=01M35FCZGFGX8BA70RYV63WBR3 contract=98X4cGfh… target=153.66.97.171:37962 retries=0 att
```

- The contract was installed at .098, the GET answer went back to the client at .109, and only then, at .116, was the UPDATE created.
- There is no `missing contract parameters` or `auto_fetch_originator` line anywhere in the buyer log (count 0), and the next read of the mailbox grew from 6738 to 8972 bytes.
- In the audit's run on `5b837b2`, this same step logged the bounce, and the message was lost.
- The seller's Inbox on peer1 showed the request (`MBX119-160819`) on its first poll after the send.
- The buyer's own thread, reopened later, shows it under "Addressed to the seller" as a message read back from the mailbox. Neither the "not received" nor the "not confirmed" card was shown.

**#118 / #122 check (asked by the overseer).** On peer1, with the published app, the seller added a NEW listing ("FRESH-SESSION 118 CHECK") in a fresh session without opening "Edit details". The listing was accepted. On the buyer's peer it shows **Buy this** with no "certificate is not this seller's" warning, and the buyer bought it (that is the request above). The only listing carrying the warning is the pre-fix "AUDIT TEST" listing from the audit run.

Not covered live: the automatic resend and the "Send again" card. Both need a node that drops or bounces the write; they are covered by the host tests listed in the description.

[AI-assisted - Claude]

sanity:
## Review round 3 (head `b74f7b4`, then rebased), consolidated; fixes in `9068a01`

Lens: an independent Opus reviewer standing in for Codex, which failed twice with "You've hit your usage limit ... try again at 4:50 PM" (on `9d4954b` and on `b74f7b4`). Not waiting for the reset.

The branch was then rebased onto `origin/main` `ecd56b1` (#122, #129). The rebase was clean with no conflicts; 683 tests passed before the round-3 fixes and 684 after.

Findings and disposition:

1. **should-fix, fixed:** a manual "Send again" starts a fresh `deliver`, so a failed write there was reported as "never reached your node, nothing was sent", although an earlier copy had. `take_for_resend` now reports whether a copy was already handed over, and `deliver` carries it. Test `a_manual_resend_that_cannot_reach_the_node_claims_nothing` fails with the flag ignored; the two-store resend test fails when `take_for_resend` always claims a hand-over.
2. **should-fix, fixed:** the Unconfirmed card said "your Freenet node did not answer", which is false for a NotFound or an unrouted mailbox. It now says only "Harvest could not confirm that this reached the seller's mailbox. It may have arrived."
3. **should-fix, fixed:** `docs/untested-invariants.md` (the `Thread` row, the resend row) and the `message_view.rs` module doc still described pre-round-2 behaviour. Both updated.
4. **should-fix, done:** the live two-peer proof is now posted (comment above). The doc row cites #126.
5. **consider, fixed:** the resend-failure notification now says "could not be sent again", so it no longer contradicts "It may have arrived".
6. **consider, fixed (docs):** the `prime.rs` module doc now says what a priming `GetResponse` triggers through `on_contract_state`, and why none of it can loop.
7. **consider, fixed:** the card label match is exhaustive.
8. **consider, done:** rebased; CI runs on the rebased head.

Verified by the lens: no caller of `update_contract` has a timeout that races the 30s prime (all are spawned fire-and-forget; the watch request's grace is 2 min), and the resend guards and round-2 plumbing are correct.

Lenses across all rounds: code-first, testing, big-picture, skeptical async (twice), failure-mode, Codex (round 1 only), plus an Opus substitute for Codex (round 3). No round found a blocking issue; every should-fix is fixed or has a recorded reason (round 1 #10, #11; #130 for the pre-existing two-stores display gap).

[AI-assisted - Claude]

sanity:
## Re-check of the round-3 fix commit `9068a01` (Sonnet, skeptical)

No blocking or should-fix findings. The lens traced every path of the `handed_over` flag: both production callers (`deliver_to_seller` with false, `resend` with the value from `take_for_resend`) are updated. `NotInMailbox` and `Unconfirmed` can only be produced after a successful write, so a copy that got through is never reported as "never reached your node", and the reverse cannot happen either.

Two consider items, not changed:
- `mailbox_ops.rs` notification wording uses `_ =>` for the "sent again" case. Only `Unconfirmed` can reach that arm with an error: `NotInMailbox` is always reported without one. This is notification text, not a decision, and a new variant would have to be added to the exhaustive card matches anyway.
- In `resend`, if deriving the mailbox key fails, the message is re-marked `Unconfirmed` even when it was `NeverReachedNode`. The derivation already succeeded for the same key on the first send, so this is unreachable, and the code comment says so.

CI green at head `9068a01d0311a3eb0ab9255baa1afd977d547698` (CI and contract drift, both completed/success). Contract hashes unchanged (store `b5eddce7`, index `0df754b5`, mailbox `fad8339c`, delegate `09507e3f`). Merging.

[AI-assisted - Claude]

